PR #22, opened 2015-06-15
Changed current to 1.0.0.Beta1 for Packetbeat

tsg:
The docs in the master start to diverge from the ones in Beta1 (which is our current released version). So I'd like to change `current` to the `1.0.0.Beta1` branch in packetbeat. Let me know if this is not the right way to do it.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -59,6 +59,7 @@ repos:
         url:        https://github.com/elastic/packetbeat.git
         current:    1.0.0.Beta1
         branches:
+            - master
             - 1.0.0.Beta1
 
     marvel:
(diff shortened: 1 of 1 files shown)


clintongormley:
Hi @tsg 

This looks fine.  You can also support multiple versions if you like, eg the beta PLUS the master

tsg:
I added master as well, maybe it will be useful. Just for my knowledge, what happens if building one of the branches breaks? Only that portion of the guide is no longer updated or the whole thing?

clintongormley:
All docs for everybody break :)

tsg:
Ok :-). Fine to merge this myself?

clintongormley:
sure, go for it

PR #36, opened 2015-09-06
Fixed typo (Used the ... to Use the ...) in Kibana4 docs

jakommo:


clintongormley:
@jakommo the docs are generated from the source repos.  you should open this in kibana instead

PR #65, opened 2015-12-18
Removed version 1.0.0 for Beats

tsg:
It's practically identical with 1.0.1, so we don't need to keep it around.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -70,7 +70,6 @@ repos:
         branches:
             - master
             - 1.0.1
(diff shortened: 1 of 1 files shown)


clintongormley:
@tsg we usually just have a single version of the docs per branch (eg 1.0, master), and use the coming[], added[] and deprecated[] tags to annotate changes

tsg:
Thanks @clintongormley doing this for 1.1.

PR #68, opened 2015-12-29
Document analyzer options in the order they're executed.

tiegz:
This updates `analysis-custom-analyzer.html` to be more clear. It places `char_filter` option before `tokenizer` and `filter`, since character filtering is done at the beginning of analysis.

Character filtering happens before tokenizing and token filtering, as mentioned [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-analyzers.html):

> Analyzers are composed of a single Tokenizer and zero or more TokenFilters. The tokenizer may be preceded by one or more CharFilters.

I think documenting these options in the order that they're executed could make ES docs clearer for newcomers.

(this probably applies elsewhere in the docs too -- I'd be happy to followup with another PR for those)


clintongormley:
Hi @tiegz 

Please open a PR against the source repo, in this case: http://github.com/elastic/elasticsearch/

PR #87, opened 2016-05-05
Build Alpha2 Logstash docs from new v5.0.0-alpha2-docs

dedemorton:
I created the v5.0.0-alpha2-docs branch so we can continue to update the 5.0 branch without affecting the published alpha2 docs. We'll work in the 5.0 branch for Beta.

@clintongormley Does this look OK?  Oddly enough, this is the first time I've created a branch in a main repo. :-)

@suyograo FYI

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -66,7 +66,7 @@ repos:
         current:    2.3
         branches:
             - master
-            - 5.0: 5.0.0-alpha2
+            - v5.0.0-alpha2-docs
             - 2.3
             - 2.2
             - 2.1


clintongormley:
@dedemorton we usually keep working on the same branch.  when you add new things, use the coming/added/deprecated annotations https://github.com/elastic/docs/#additions-and-deprecations which we can then update or remove when we do the release.  (using the `docs/update_versions.pl` script)

If you create a new branch, then you're going to delete it again, which means google will suddenly return 404s.

clintongormley:
Also, the benefit of working on the published branch is that it is built every half an hour, so we spot problems early instead of minutes before the release

dedemorton:
@suyograo Do you have any concerns with using the "coming" annotation instead of a separate branch? Are there situations where we wouldn't want users to know what's coming? (I don't _think_ there are, but figured it's worth asking.) I'm honestly not too worried about build errors cropping up at the last minute (we'll also be building in master, so we'll see any errors there). And it's pretty easy to get google to re-crawl the site. I'd actually prefer to use the "coming" annotation, but it won't be useful unless we can get everyone to use it consistently. WDYT?

@debadair @palecur See Clint's suggestion for flagging upcoming changes. I'd forgotten that we can do this. I haven't seen these annotations in anything that I've worked on, so I wonder if anyone on the team is actually using these. Deb: I'm wondering if this would work for you with the X-Pack docs. It would be nice if we used the same approach across the release.

dedemorton:
@tsg See Clint's suggestion on how to flag coming changes. We should consider this approach for the Beats doc, too. I'd really like for us to be consistent in how we do this so users know what they're looking at.

suyograo:
> Do you have any concerns with using the "coming" annotation instead of a separate branch? Are there situations where we wouldn't want users to know what's coming?

None from my side

dedemorton:
OK, then I'll close this request and use the "coming" annotation instead.

tsg:
@dedemorton  +1 for using this approach in the Beats docs as well.

PR #90, opened 2016-05-19
Fix example for spark.

withccm:
[The Writing to dynamic/multi-resources example](https://www.elastic.co/guide/en/elasticsearch/hadoop/current/spark.html#spark-write-dyn) is not working.

I was confused because it was used with 'media-type' and 'media_type'.
I checked the source files, 'media_type' has been used more than 'media-type'.

as-is
'media-type' 
to-be
'media_type'


clintongormley:
Hi @withccm 

This issue should be opened in https://github.com/elastic/elasticsearch-hadoop instead

withccm:
Thank you.

I open [this issue](https://github.com/elastic/elasticsearch-hadoop/issues/767).

PR #104, opened 2016-07-29
Docs said something different then the command needs

cskiwi:
It had to be `file:/` instead of `file://`


clintongormley:
Hi @cskiwi 

This is not the right place to submit changes for docs.  They should go to the repository which holds the docs sources.  Also, protocols always take two slashes eg `http://`, `file://`.  Then the `file://` protocol takes an absolute path, giving three slashes.

PR #105, opened 2016-07-29
Uncomment java REST client docs

javanna:
We disabled the java REST client docs temporarily till we get the next release out, otherwise it would be documentation about code that is not released yet.

We have to get this PR in once the next version is out.


javanna:
not needed anymore after https://github.com/elastic/docs/commit/9a87e6eacf61dfce45456449268705461c3b9da6 .

PR #110, opened 2016-08-27
Add COPY AS CURL link to CONSOLE

nik9000:
Adds a link that copies the `// CONSOLE` snippets as valid `bash`
snippets using `curl`.

```js
PUT my_index/my_type/1
{
  "title": "This is a document"
}

PUT my_index/my_type/2?refresh=true
{
  "title": "This is another document",
  "body": "This document has a body"
}

GET my_index/_search
{
  "query": {
    "terms": {
      "_field_names": [ "title" ]
    }
  }
}
```

becomes

```bash
curl -XPUT 'localhost:9200/my_index/my_type/1?pretty' -d'
{
  "title": "This is a document"
}'
curl -XPUT 'localhost:9200/my_index/my_type/2?refresh=true&pretty' -d'
{
  "title": "This is another document",
  "body": "This document has a body"
}'
curl -XGET 'localhost:9200/my_index/_search?pretty' -d'
{
  "query": {
    "terms": {
      "_field_names": [ "title" ]
    }
  }
}'
```


nik9000:
I'm not experienced at all with Javascript or browsers in general but the hard bit is really just `document.execCommand('copy');` which looks to be supported in modern Chrome (43), Firefox (41), and IE (11, though I've seen reports of 10). Unfortunately modern IE is rarer than modern Chrome or Firefox so it likely isn't going to work for IE users. I'm not sure what to do for them.

nik9000:
I'm not sure who should review things here, @clin

resources/web/docs.js
```diff
@@ -214,6 +216,51 @@ jQuery(function() {
...
+        var comment = match[1];
+        var method = match[2];
+        var path = match[3];
+        var body = match[4];
+        if (comment) {
+          curlText += comment + '\n';
+        } else {
+          path = path.replace(/^\//, '');
+          path += path.includes('?') ? '&pretty' : '?pretty';
+          body = body.replace(/\'/g, '\'"\'"\'');
```
clintongormley: I wonder if this escaping works across shells.  Perhaps replace`'` with `\u0027` instead?

clintongormley:
thanks @nik9000 - this is awesome.  left one question

resources/web/docs.js
nik9000: I can try it! I don't use other shells but I can certainly use this as an excuse to spend a few minutes installing and trying a few.

nik9000: What I have now works in bash, zsh, and ksh. It doesn't work in tcsh but that seems to require `\` before every new line which isn't cool. I can still do `\u0027` if you'd like though.

nik9000: Just tried `\u0027`. Works for the same shells.

clintongormley:
thanks @nik9000 - merged

nik9000:
Hurray!! Thanks for reviewing!

PR #111, opened 2016-08-27
Don't add console widget to snippets marked // NOTCONSOLE

nik9000:
Those are just for explicitly tracking snippets that shouldn't be
converted to `// CONSOLE`. The console widget doesn't make sense
for them.


nik9000:
See https://github.com/elastic/elasticsearch/pull/19934

clintongormley:
thanks @nik9000

PR #122, opened 2016-10-25
Add stack upgrade docs

dedemorton:
@debadair @pickypg Please review. 

Here's what the landing page will look like. I'm just throwing something out there for discussion.

landingpage

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -160,7 +160,7 @@ contents:
         sections:
           -
             title:      Installation and Upgrade Guide
-            prefix:     en/elasticstack
+            prefix:     en/elastic-stack
             repo:       stack
             index:      docs/index.asciidoc
             tags:       Elastic Stack/Installation and Upgrade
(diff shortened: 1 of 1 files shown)


conf.yaml
```diff
@@ -150,6 +156,15 @@ repos:
 contents_title:     Elastic Stack and Product Documentation
 
 contents:
+    -   title:      "Elastic Stack"
+        sections:
+          -
+            title:      Installation and Upgrade Guide
+            prefix:     en/elasticstack
```
pickypg: Should we do `elastic-stack` or `elasticstack`? I misread it at first as `elasticsearch` without the dash.

conf.yaml
```diff
@@ -4,6 +4,12 @@ paths:
     branch_tracker: html/branches.yaml
 
 repos:
+    stack:
+        url:        https://github.com/elastic/stack-docs.git
+        current:    master
```
nrichers: This upgrade guide is going to apply to a specific version for every release. Do we maybe want to build `current` from a 5.0 release for now?

pickypg: (We talked on Slack: I think we should wait to start versioning once we're sure that it is stable, then fix it).

Issue #123, opened 2016-10-26
Typo in Enabling Anonymous Access documentation


cbismuth:
Replace `role` by `role2` for role assigned to anonymous user.

clintongormley:
Hi @cbismuth 

I'm afraid I have no idea which page you're referring to here :)  Any chance of a link?

clintongormley:
Ah I see you sent a PR https://github.com/elastic/docs/pull/124

PR #124, opened 2016-10-26
#123 Fix typo in Enabling Anonymous Access documentation

cbismuth:
Replace `role` by `role2` for role assigned to anonymous user.


clintongormley:
Fixed in upstream repo.  thanks for reporting

PR #144, opened 2017-01-20
Add Content-Type header to COPY AS CURL

### LINKED ISSUE #143: copy-as-curl should include content-type
With https://github.com/elastic/elasticsearch/pull/22691, the content-type header will become mandatory for HTTP requests with bodies.  copy-as-curl should be updated to include that header.

/cc @nik9000 in case you pick it up before i do

nik9000:
Adds the `Content-Type: application/json` header to the curl
request generated by the `COPY AS CURL` links if the request
has a body.

Closes #143


nik9000:
Thanks for reviewing @dadoonet!

PR #151, opened 2017-03-01
Add include-tagged macro

nik9000:
This macro allows you to include a portion of a file like this:
```
["source","java",subs="attributes,callouts,macros"]
include-tagged::{doc-tests}/DeleteDocumentationIT.java[delete-request]
<1> Index name
<2> Type
<3> Document id
```

Which can be used to extract a section of a file like this:
```
        // tag::delete-request
        DeleteRequest request = new DeleteRequest(
            "index",    // <1>
            "type",     // <2>
            "id");      // <3>
        // end::delete-request
```

### DIFF
--- a/resources/es-asciidoc.conf
+++ b/resources/es-asciidoc.conf
@@ -14,6 +14,7 @@
 (?us)[\\]?(?P<name>(experimental))\[]=[]
 (?us)[\\]?(?P<name>(beta))\[\s*(?P<passtext>[^\]]+?)\s*]=[]
 (?us)[\\]?(?P<name>(beta))\[]=[]
+(?us)^\s*[\\]?(?P<name>include-tagged)::(?P<file>[^\[]+)\[(?P<tag>[^\]]+)\]\s*$=[]
 
 [added-inlinemacro]
 <phrase revisionflag="added" revision="{1}">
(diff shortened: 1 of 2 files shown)


dadoonet:
I gave a quick look at it (well I'm not a Python guy :) ). It looks good.

I guess that automatically find what is the indentation of the tag with `indentation = m.group(1)`.

nik9000:
> I guess that automatically find what is the indentation of the tag with indentation = m.group(1).

Right. We copy the indentation of the tag and strip it from all the other fields. At least, that is what I tried to do.

resources/es-asciidoc.conf
```diff
@@ -40,6 +41,10 @@
 {passtext}
 </phrase>
 
+[include-tagged-inlinemacro]
+# it is a nightmare to get this to work with eval3 so we're forking a new process every time....
```
nik9000: I think `eval3` should be `sys3` actually.

nik9000:
@clintongormley, do you want to have a look at this?

resources/extract-tagged.py
```diff
@@ -0,0 +1,32 @@
...
+            if foundTag:
+                result = result + line.replace(indentation, '', 1)
+            else:
+                m = start.search(line)
+                if m:
+                    foundTag = True
+                    indentation = m.group(1)
+
+
+if __name__ == '__main__':
```
clintongormley: I'd tidy up the error output a bit, to look more like this:

```
if __name__ == '__main__':
    from sys import argv, stdout, stderr
    try:
      result = extract_tagged(argv[1], argv[2])
      if result:
          stdout.write(result)
      else:
          raise Exception("Couldn't find tag")
    except Exception as e:
      stderr.write("asciidoc: ERROR: Extracting tag `" + argv[2] + "` from file `" + argv[1]+ "`: " + e.__str__() + "\n")
      exit(1)
```

nik9000:
Thanks @clintongormley!

PR #159, opened 2017-04-03
[DOCS] Re-add docs/en for X-Pack

lcawl:
This PR attempts to re-add the "docs/en" in the conf.yaml for the X-Pack Reference doc builds

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -5,7 +5,7 @@ paths:
 
 docs_repo:
     url:            https://github.com/elastic/html-docs.git
-    
+
 repos:
     beats:                https://github.com/elastic/beats.git
     cloud:                https://github.com/elastic/cloud.git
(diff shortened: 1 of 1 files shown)


conf.yaml
```diff
@@ -291,20 +291,20 @@ contents:
...
             private:    1
             branches:   [ 5.3, 5.2, 5.1, 5.0 ]
             sources:
               -
                 repo:   x-pack
-                path:   docs/public
+                path:   docs/en
               -
                 repo:   x-pack-kibana
-                path:   docs/public-kb
```
clintongormley: this shouldn't change as you've only renamed public -> en in `x-pack`,  not in `x-pack-kibana`

conf.yaml
```diff
@@ -291,20 +291,20 @@ contents:
...
+                path:   docs/en
                 exclude_branches:   [ 5.3, 5.2, 5.1, 5.0, 2.4, 2.3, 2.2, 2.1, 2.0]
               -
                 repo:   x-pack-elasticsearch
-                path:   docs/public
```
clintongormley: this shouldn't change as you've only renamed public -> en in `x-pack`,  not in `x-pack-elasticsearch`

conf.yaml
```diff
@@ -291,20 +291,20 @@ contents:
...
+                path:   docs/en
                 exclude_branches: [ 5.3, 5.2, 5.1, 5.0, 2.4, 2.3, 2.2, 2.1, 2.0]
```
clintongormley: `exclude_branches` should be a subset of `branches`, so you can delete all the 2.x branches

conf.yaml
```diff
@@ -291,20 +291,20 @@ contents:
...
             sources:
               -
                 repo:   x-pack
-                path:   docs/public
+                path:   docs/en
               -
                 repo:   x-pack-kibana
-                path:   docs/public-kb
+                path:   docs/en
                 exclude_branches:   [ 5.3, 5.2, 5.1, 5.0, 2.4, 2.3, 2.2, 2.1, 2.0]
```
clintongormley: `exclude_branches` should be a subset of `branches`, so you can delete all the 2.x branches

conf.yaml
lcawl: It's not used by 5.3 or earlier.  But in 5.x and master, it *is* used, and it *is* renamed.

conf.yaml
lcawl: Ditto comment above.  It's only used in 5.x and master, where it *is* renamed

conf.yaml
lcawl: Thanks, I've committed that change.

PR #160, opened 2017-04-05
Add ECE 1.0.0-beta2 branch

nrichers:
We're releasing a 1.0.0-beta2 branch next week, which requires an accompanying doc branch.


uric:
LGTM

PR #165, opened 2017-04-15
Change Elastic Cloud docs build to use --chunk=1

nrichers:
The Elastic Cloud Enterprise docs already used chunk=1. I did some reshuffling of Elastic Cloud content and I would prefer that content to also use --chunk=1. By and large, we have the right [float] tags in place to make this work.


dedemorton:
LGTM

PR #167, opened 2017-04-19
Corrected mistake on enabling a user

cheeweee:
Corrected a mistake on disabling a user, when it should be enabling.

Please review the changes.


dedemorton:
@cheeweee The file you've modified is generated by the doc build process. To make the change, you need to update the source asciidoc in the repo for the component you are updating. Can you make the change in the following repo, instead, and submit a new request? Thanks!

https://github.com/elastic/x-pack-elasticsearch/blob/master/docs/en/rest-api/security/users.asciidoc

Closing this request because we do not make changes to the files generated by the doc build.

PR #168, opened 2017-04-19
Add Beats devguide

dedemorton:
Adds the Beats Developer Guide to the docs.

This PR needs to be merged at the same time as https://github.com/elastic/beats/pull/4030

Note that `generate/metricbeat/metricset` was removed from the Metricbeat section because the file at the path is no longer used in the docs.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -500,7 +500,28 @@ contents:
                 path:   libbeat/docs
               -
                 repo:   beats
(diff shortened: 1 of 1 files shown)


nrichers:
LGTM. As per @dedemorton, we're listing only `master` as a branch, because we are not back porting the change.

PR #171, opened 2017-04-26
Remove `//` from included callouts

nik9000:
Java files that are references by `include-tagged` have to
comment out the callouts so the java files compile:
```java
        // tag::bool
        boolQuery()
                .must(termQuery("content", "test1"))    // <1>
                .must(termQuery("content", "test4"))    // <1>
                .mustNot(termQuery("content", "test2")) // <2>
                .should(termQuery("content", "test3"))  // <3>
                .filter(termQuery("content", "test5")); // <4>
        // end::bool
```

This removes the `// ` part from the docs because it is distracting.

### DIFF
--- a/resources/extract-tagged.py
+++ b/resources/extract-tagged.py
@@ -6,6 +6,7 @@ def extract_tagged(target, tag):
     import re
     start = re.compile('^(\s+).+tag::' + tag)
     end = re.compile('end::' + tag)
+    callout = re.compile(r'// (<[^>]+>)')
     foundTag = False
 
     result = ''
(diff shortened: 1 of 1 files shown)


resources/extract-tagged.py
```diff
@@ -6,6 +6,7 @@ def extract_tagged(target, tag):
     import re
     start = re.compile('^(\s+).+tag::' + tag)
     end = re.compile('end::' + tag)
+    callout = re.compile(r'// (<[^>]+>)')
```
clintongormley: to make this safer, what about:

    callout = re.compile(r'// (<[^>]+>)\s*?\n')

resources/extract-tagged.py
```diff
@@ -14,7 +15,9 @@ def extract_tagged(target, tag):
             if end.search(line):
                 return result
             if foundTag:
-                result = result + line.replace(indentation, '', 1)
+                line = line.replace(indentation, '', 1)
```
clintongormley: with the above change you'll need to add a `\n` here too

nik9000:
@clintongormley I've pushed the extra safety you asked for.

nik9000:
Thanks for reviewing @clintongormley!

PR #172, opened 2017-04-26
Fix include-tagged when tags are similar

nik9000:
If one tag is a prefix of another tag then `include-tagged` the
suffix could be included instead of the prefix if it came first
in the file. This fixes that by expecting the tag to be followed
by only spaces and a newline.


nik9000:
@clintongormley `\b` works for me! I've pushed that.

nik9000:
@clintongormley can you have a look? I know you are busy but I figured it'd be nice to get this and https://github.com/elastic/elasticsearch/pull/24354 in.

PR #174, opened 2017-04-29
[DOCS] Add role=screenshot formatting for images

lcawl:
This PR implements shadow formatting for images that have "role"="screenshot".
It works in my local tests against the X-Pack Reference, in particular the Machine Learning Getting Started Tutorial.

It converts asciidoc content like this:
[role="screenshot"]
image::images/ml-kibana.jpg[Job Management]

To output HTML content like this:

----

<div class="screenshot"><div class="mediaobject">Job Management</div></div>

----

I'm not a XSLT guru, however, so if you have suggestions for a more elegant solution, please let me know.

### DIFF
--- a/resources/fo.xsl
+++ b/resources/fo.xsl
@@ -82,5 +82,15 @@
   <!-- Don't display in ToC -->
   <xsl:template match="phrase" mode="no.anchor.mode" />
 
-</xsl:stylesheet>
+  <!--Images -->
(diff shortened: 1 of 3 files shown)


clintongormley:
thanks @lcawl 

This didn't work for images with titles, but I've pushed a change that does.  Also, we can use any role name now, not just `screenshot`, so we can add more styles later.

Issue #176, opened 2017-05-08
Create custom admonition block to highlight X-Pack-specific content 


lcawl:
As X-Pack-specific information is added to non-X-Pack-specific books, it becomes important to be able to draw attention to its specific context. 

Per discussion with @clintongormley, it should be possible to create a customized admonition block (for example, with role=xpack) that has a unique icon.

PR #177, opened 2017-05-08
[DOCS] Add X-pack admonition block

lcawl:
This PR is related to https://github.com/elastic/docs/issues/176, 
which aims to create an X-Pack-specific admonition block.

When I tested these changes with the following example in an Elasticsearch page that contained:
<pre>
[role="xpack"]
[NOTE]
=====================================================================
<<machine-learning-node,Machine learning node>>::
A node that has `xpack.ml.enabled` and `node.ml` set to `true`, which is the
default behavior when X-Pack is installed. If you want to use X-Pack machine
learning features, there must be at least one machine learning node in your
cluster. For more information about X-Pack machine learning features, see
{xpack}/xpack-ml.html[Machine Learning in the Elastic Stack].
=====================================================================
</pre>

And added the xpackicon.png to elasticsearch/docs/reference, it generated the following:

admonition_test

I would ideally prefer to see this icon in the margin and the text aligned with the rest, but perhaps this can be tweaked.  What do you think?


clintongormley:
i like the idea for admonition blocks.  Not sure the same thing works in eg a definition list like you're doing it here, where (i agree) the text should have the same alignment as the other definitions.

What about having a title style (which can be used for definition lists as well) where we have a small X icon at the beginning of the title, eg

    Some defn
        Text text .....
    [X] X-pack defn
        Text text .....

lcawl:
@clintongormley I've updated the styles.css to add an pack icon at the beginning of titles like you suggested, and it works when I update the output html to contain <h3 class="xpack">... for example. I haven't gotten the XSL to successfully generate those classes yet, though. 

Shouldn't it just be similar to what we did for adding classes to images (in website_common.xsl), but in this case matching on part|chapter|section? Maybe this is a case we can use in one of the future doc build skills transfers?

clintongormley:
> Shouldn't it just be similar to what we did for adding classes to images (in website_common.xsl), but in this case matching on part|chapter|section? Maybe this is a case we can use in one of the future doc build skills transfers?

If only...

```
[role="foo"]
== Some title
``` 

... ends up applying the `foo` class to the whole chapter (or section), rather than to just the title.  Maybe that's OK?

lcawl:
Yes, I think having the role applied to the whole section would be okay (as long as I removed the admonition block code).  I did try that earlier this week, but it wasn't picking up the role when I tagged it that way. I'll keep playing.

clintongormley:
@lcawl to make that work just replace `figure|informalfigure` [here](https://github.com/elastic/docs/blob/master/resources/website_common.xsl#L352) with `*`.

lcawl:
Thanks for the tip, @clintongormley! I am heading into a long week-end, so I'll implement that on Monday.

lcawl:
I updated the website_common.xsl to use "*" but for some reason it still doesn't apply the role to the titles.  For example, if I have:

<pre>
[role="xpack"]
<<machine-learning-node,Machine learning node>>:: 
A node that has `xpack.ml.enabled` and `node.ml` set to `true`, which is ... 
</pre>

The resulting <dt> and <dd> do not have class=xpack. An embedded paragraph does, however.

Nothing happens for sections like this:

<pre>
[float]
[[machine-learning-node]]
[role="xpack"]
=== Machine learning node
</pre>

... even when I try `[float, role="pack"]` and '[float][role="xpack"]`.

clintongormley:
> The resulting `<dt>` and `<dd>` do not have class=xpack. An embedded paragraph does, however.

correct.  the role is not passed on by asciidoc to docbook.  i couldn't find any way to configure asciidoc to do so.  You can only apply a role to the whole defn list.

lcawl:
I think it's okay to not have a fancy way of adding the graphic to definition lists. We can either add it manually or else leave it only for chapters/sections. 

With respect to sections, I finally got my example to work on that too, but only after I removed the [float]. Per http://www.methods.co.nz/asciidoc/userguide.html, "The first positional attribute (role attribute) is translated by AsciiDoc to an HTML span element class attribute or a DocBook phrase element role attribute".  But it also says "Setting the title’s first positional attribute or style attribute to float generates a free-floating title".  So I think they're both fighting for the "first positional attribute" in this case.  I tried setting both explicitly with `[role="xpack",style="float"]` but the role wasn't carried forward.  Do you think it's an acceptable limitation that we can't use `[float]` for the sections that use this role?

clintongormley:
These seem to work:

This sets the `xpack` role for the whole chapter/section/etc:

```
[role="xpack"]
== Some title
```

To set the class for a floated title, do:

```
[style="float",role="xpack"]
=== Some title
```

And for any inline text, you can wrap it with `<span class="xpack>...</span>` as follows:

```
[xpack]#Some term#::

Some definition including a bit about [xpack]#X-Pack only#.
```

The `#` imply "quoted text" (like `+`, `*`, `_`, etc) but don't have any other effect like causing underlines.  They just allow you to apply the role.

lcawl:
Thanks @clintongormley.  Those all seem to work so I've updated the style.css accordingly.  There is one slight difficulty, however.  When I do this in node.asciidoc:

<pre>
[sytle="float",role="xpack"]
[[machine-learning-node]]
=== Machine learning node
....
</pre>

All subsequent sections in that page are flagged with the role (and therefore image) too.  It doesn't matter if I try unsetting the role.  For example:

<pre>
[style="float",role=""]
== Node data path settings
...
</pre>

When I look at the output html, it looks like the machine learning section creates a `<div class="xpack section">` that must include all the others, even though they're not really subsections.  I think this can work around in the short term by ensuring that any x-pack-specific sections are at the end of a page, but do you know of any other ways of closing that section?

lcawl:
It looks like 
<pre>
[float]
[[machine-learning-node]]
=== [xpack]#Machine learning node#
</pre>
might work... I'll keep playing

lcawl:
I think it finally works!

As you mentioned, for definition lists, we can use: 
<pre>
[xpack]#<<machine-learning-node,Machine learning node>>#::
</pre>

For non-floating sections, we can use:
<pre>
[role="xpack"]
[[machine-learning-node]]
=== Machine learning node
</pre>

For floating sections, we can use:
<pre>
[float]
[[machine-learning-node]]
=== [xpack]#Machine learning node#
</pre>

I updated the .css (hopefully for the last time) to make all those work and tested it with node.asciidoc.

clintongormley:
> To set the class for a floated title, do:

It turns out it was passing the role to docbook, but docbook wasn't then using the `role` in bridgehead.  My bad.

Your suggested solutions make sense.

clintongormley:
The xpack image isn't attached to this PR so I can't see what it looks like.  Btw, instead of attaching it as a file, you can embed it as a base64 encoded image in the CSS itself, like I've done with other images here: https://github.com/elastic/docs/blob/master/resources/web/styles.css#L132

You can use this site to encode the image: https://www.base64-image.de/

lcawl:
Thanks, @clintongormley, that's much better than adding the image in each book.  
I also fixed the span formatting in the "On this page" section (for the case where the title contained a span element due to the role=xpack).

lcawl:
@clintongormley and @debadair can we go ahead with this as-is and optionally improve the graphic when additional feedback is received?  I would like to start using this functionality in the "graph" and "reporting" content that I'm integrating in the Kibana guide.

clintongormley:
Sure - I'll merge

PR #178, opened 2017-05-09
Use python2.7 rather than system default python version

stu-elastic:
My OS defaults to python 3
```
/usr/bin/env python --version
Python 3.6.0
```
so doc generation fails.  

I propose changing the hash bang lines to be explicit about using python2.7 the changes should work cross platform but I've only tested locally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ bin
 !html/sitemap.xml
 metastore_db
 html_docs/
+.idea
--- a/resources/asciidoc-8.6.8/a2x
+++ b/resources/asciidoc-8.6.8/a2x
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2.7
 '''
 a2x - A toolchain manager for AsciiDoc (converts Asciidoc text files to other
       file formats)
--- a/resources/asciidoc-8.6.8/asciidoc
+++ b/resources/asciidoc-8.6.8/asciidoc
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2.7
 """
 asciidoc - converts an AsciiDoc text file to HTML or DocBook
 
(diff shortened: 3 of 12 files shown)


clintongormley:
Looks like asciidoc upstream doesn't have a solution for this.  I'm not crazy about making changes to the `asciidoc-8.6.8` directory as those will be lost when we upgrade.  Also, not sure about hardcoding to 2.7, what if people have a different version of python installed?

Is there something else we can do to automatically in the build script?

nik9000:
> Is there something else we can do to automatically in the build script?

So include-tagged shouldn't need any special handling to use the right python but we may as well change it because it is ours. Asciidoc uses the following to launch it:
```
{sys3:{python} {asciidoc-dir}/../extract-tagged.py {file} {tag}}
```

`{python}` means "use the python that you launched asciidoc with" so we're fine there.

I think we could modify `Util.pm` to run `python2.7` instead of `a2x` directly. We should probably make `python2.7` a variable just so we can add special handling just in case folks don't have python 2.7 at that name, though I think that name is pretty common. I know that is the name my mac and my ubuntu machine have for it.

clintongormley:
> I think we could modify Util.pm to run python2.7 instead of a2x directly.

But `a2x` needs to call `asciidoc` as well, and it does so be shelling out.

stu-elastic:
>  Also, not sure about hardcoding to 2.7, what if people have a different version of python installed?

I've used this method across OS X and Linux systems (not Windows tho).

This option works as long as they have python 2.7 installed rather than 2.5 or 2.6 (OS X since 10.6, most linuxes forever and RHEL since 7.0).  Python 2.7 was released in 2010.

[PEP 394](https://www.python.org/dev/peps/pep-0394/) recommends using `python2` and `python3` if you need version specific behavior, which would also work but I don't have experience with it.

Looking through the a2x code, it doesn't use any python2.7 specific functionality, so using `python2` is more general.

stu-elastic:
> I think we could modify Util.pm to run python2.7 instead of a2x directly.

I'm happy to take a pass at that if you'd prefer it over the `python2` option.

clintongormley:
Take a look at https://github.com/elastic/docs/pull/179 - i think this should work without us having to edit asciidoc to hardcode versions

PR #179, opened 2017-05-11
Add a proxy for the python executable which can look up python2 binaries

clintongormley:
This PR adds the `docs/` dir into the `PATH` and provides a `python` executable (written in Perl) which tries to find `python2.7` down to `python2.3`.

Closes #178

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -452,7 +452,9 @@ sub init_env {
         file('resources/docbook-xml-4.5/catalog.xml')->absolute;
 
     $ENV{PATH}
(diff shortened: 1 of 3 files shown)


clintongormley:
There's already https://github.com/asciidoc/asciidoc/issues/83 which dates from 2015...

nik9000:
Upstream pushed a change to asciidoc so it has `python2` as the shebang.

clintongormley:
@nik9000 thanks! i've opened https://github.com/elastic/docs/issues/180 to upgrade when the next version is released (although the last version came out 4 years ago)

PR #189, opened 2017-06-01
Remove pre-1.0.0 ECE docs except beta2

nrichers:
ECE has GA'ed and we no longer offer the install scripts for versions before GA. Time to remove the old docs?

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -447,7 +447,7 @@ contents:
             prefix:     en/cloud-enterprise
             tags:       CloudEnterprise/Reference
             current:    1.0.0
-            branches:   [ 1.0.0 ]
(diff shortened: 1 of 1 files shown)


uric:
I think we should remove the alphas and beta1, but keep beta2 up for a few more weeks in case people are still testing it. 
@nrichers can we add a disclaimer at the top of the beta2 docs that says something like "this is a beta version of ECE, we recommend upgrading to the official 1.0 GA release"?

nrichers:
@uric I re-added the beta2 branch to `conf.yaml`. If LGTM, approve this PR?

Separately, I pushed a beta2 doc PR to tweak the header as requested. Please approve https://github.com/elastic/cloud/pull/6012?

uric:
LGTM

PR #197, opened 2017-06-15
[DOCS] Remove redundant X-Pack icons

lcawl:
Due to the way class attributes are inherited, setting "role=xpack" at the highest level of a page causes  the X-Pack icon to appear on the page title and every section title within the page.  This is overwhelming.  For example:

toomanyxpackicons

I have therefore updated the stylesheet so that the icon is only added when the role is explicitly specified.   

For example:
1. Set at the highest level:
fewerxpackicons1

2. Set at a lower section level:
fewerxpackicons2


lcawl:
I've compressed the new code a bit and I also commented out the code that adds an X-pack icon to the "On this page" section of the page.  This is because that solution was incomplete and did not cover all role=xpack scenarios, per https://github.com/elastic/x-pack-elasticsearch/issues/1731.  It can be re-added when a full solution is found.

PR #202, opened 2017-06-26
[DOCS] Created shared attribute file

lcawl:
This pull request creates a list of attributes that can be used by all the documentation builds, so that attribute names are synchronized across the repositories.  This is particularly important since many of the builds now integrate content from multiple repositories.


debadair:
LGTM

PR #215, opened 2017-07-25
Update conf.yaml to include root dir of each beat

dedemorton:
Add directory paths to the root folder for each beat to pick up the reference.yml file example.


debadair:
LGTM

PR #216, opened 2017-07-27
Add attribute ref for Java Rest Client

tlrx:


shared/attributes.asciidoc
```diff
@@ -4,6 +4,7 @@
 :kibana-ref:      https://www.elastic.co/guide/en/kibana/{branch}
 :stack-ref:       http://www.elastic.co/guide/en/elastic-stack/{branch}
 :javaclient:      https://www.elastic.co/guide/en/elasticsearch/client/java-api/{branch}
+:java-rest:       https://www.elastic.co/guide/en/elasticsearch/client/java-rest/{branch}
```
javanna: do we also need to add the high level client ?

tlrx: I don't think so, both are part of the same "Java REST Client" book

javanna: oh good point, thanks

PR #219, opened 2017-08-02
[DOCS] Add xpack icon for sidebars

lcawl:
This pull request adds the x-pack icon for sidebars that are tagged as follows:

````
[role="xpack"]
.Block heading
*******************************************
Block text
*******************************************
````


lcawl:
The resulting output is like this:
sidebar1

PR #221, opened 2017-08-03
Update Elastic Cloud and ECE doc builds

nrichers:
Two changes here:
- Remove the 1.0.1 and 1.0.0 doc builds now that we are building from 1.0 as our preferred 1.0.x doc branch (depends on redirects to 1.0 being live via https://github.com/elastic/website-www.elastic.co/issues/1782)
- Add builds for master branches for both EC and ECE


dedemorton:
LGTM

PR #230, opened 2017-08-25
transport.profiles.client.xpack.security.ssl.client_authentication: n…

andel7:
…o is not valid value

transport.profiles.client.xpack.security.ssl.client_authentication: no should actually be transport.profiles.client.xpack.security.ssl.client_authentication: none otherwise elasticsearch doesn't start.


clintongormley:
thanks for the PR @andel7 but this is the wrong repository.  i've made the change in the x-pack source repo to fix this

PR #236, opened 2017-09-15
Add apm related asciidoc attributes

watson:
I'm not sure if this is the normal procedure for adding new references, so please let me know if I should do this in another way.

We're going to be making a lot of cross reference links in the docs between the different APM components. So having these references will help keep everything nice and clean.


watson:
I'm not sure if I should have `{branch}` on each of the agents as they are not part of the normal release train and will not have for instance a `6.0` branch.

dedemorton:
LGTM

dedemorton:
@watson You can use `current` instead of `{branch}` if you always want to send users to the most current published version of the doc.

watson:
@dedemorton Updated it to use `current` for the agents

watson:
@beniwohli fyi

PR #241, opened 2017-09-25
Correct statement on bulk failure with create

portante:
The current statement seems wrong, "create will fail if a document with the same *index* and type exists already, ...".  It seems like it should read, "create will fail if a document with the same `_id` and type exists already, ...".


portante:
This was noticed in the 2.4 docs, but seems to apply to all versions if this change is actually correct.

clintongormley:
Hi @portante 

Thanks for the PR, but it is the wrong repo.  Docs need to be corrected in the source repo (in this case, in elasticsearch)

PR #246, opened 2017-10-27
Add 6.0 version for apm-server documentation.

simitt:
This should only be merged shortly before rc2 is released.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -724,8 +724,8 @@ contents:
             title:      Getting Started with Elastic APM
             prefix:     en/apm/get-started
             index:      docs/guide/index.asciidoc
-            current:    master
-            branches:   [ master ]
+            current:    6.0
+            branches:   [ master,  { 6.0: 6.0.0-rc2  } ]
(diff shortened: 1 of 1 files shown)


roncohen:
I think we can wait until 6.1 is out with versioning them. WDYT @simitt ?

simitt:
In case we need to update something in the documentation we would have the same issues as we had them after releasing alpha1 - either the documentation for the released version or the documentation for the github master branch would be outdated.

roncohen:
I was thinking that it would save us a menu item in the dropdown, but you're right, this is better 👍

dedemorton:
Just a heads-up that these changes require a build --all of the docs. Someone (usually Clint or Deb) does this on the day of the release. You just want to make sure your changes get into the conf.yaml before the full doc build happens.

simitt:
@roncohen should we get this merged in tomorrow then?

debadair:
Yep--I'll check before we kick off the build & merge it if it hasn't been already.

PR #247, opened 2017-10-27
Elastic Cloud docs: Switch to saas-release and retire external branch

nrichers:
Elastic Cloud has switched to the `saas-release` branch for service releases. In an effort to keep the code and the docs together, I'd like to switch to the same branch for the current Elastic Cloud docs and to retire the arbitrarily chosen docs-only `external` branch.

Merging this PR should coincide with the release of the Cloud ID on Tuesday Oct 31 (see https://github.com/elastic/cloud/issues/7412), as `saas-external` already contains those docs.


debadair:
LGTM

PR #256, opened 2017-11-14
Change current ECE branch to 1.1

nrichers:
This PR sets the current branch for ECE to `1.1`, which is the correct branch when we GA tomorrow morning.


debadair:
LGTM.

Issue #249, opened 2017-11-01
Make doc builds more modular


debadair:
We've outgrown the monolithic approach to building, verifying, and publishing the docs. Along the way, the doc repo has grown to the point where it's problematic for people to clone.

The most obvious signs of this are our release day delays. People don't always build the docs before merging and *very* rarely build everything, which is the only way to check cross doc links. Last minute and unrelated changes frequently require us to restart the doc build, and because we have to build everything, that takes a non-trivial amount of time.

This also affects our day to day workflow, which features frequent diversions to chase down unrelated build failures so we can finish the tasks at hand.

Ideally, we'd be able to:

- Run pre-build checks on the asciidoc files to find common problems that are slipping through the cracks--undefined attributes, broken external links, maybe even spellcheck?
- Check cross-doc links and perform other post-build verification on specific books using the already-built versions of everything else. 
- Run the Kibana link checks independently/get a report that shows the topics that are being linked to in a specific book. 
- Publish selected books/branches. Updating the current version for Curator, for example, shouldn't require rebuilding and publishing everything else that's changed. Doc build failures on master shouldn't block a version release. Cloud releases shouldn't be blocked by doc build failures in the individual product docs.

In addition to making the doc build itself more modular, we need to figure out how to make the docs repo easier to deal with--perhaps by splitting the build infrastructure and generated docs apart. It can literally take people hours to clone the repo, so people avoid putting other doc tools there and opt out of contributing to the docs.

We've started to chip at these problems in the following PRs:
- [x]  Add ability to omit/mute link checking: https://github.com/elastic/docs/pull/261
- [x]  Add ability to check links faster (i.e. without rebuilding all docs): https://github.com/elastic/docs/pull/262

dedemorton:
Would be great to have intelligent spell checking (that's based on a dictionary we define), so we can catch misspellings for product names and such and avoid false positives (code, API names, etc).

nrichers:
Re: **Run pre-build checks on the Asciidoc files**: There are additional things we should check for that we do not check today:

* Accessibility issues with tables that do not have proper headers, with spatial and color references, etc.
* Forbidden words and phrases, such as "It's easy to ...", "simply," etc,
* (Future: Style checks, maybe? We have a wide range of skills at Elastic and there is often low-hanging fruit when I edit content that an automated check could flag just as well I as a human, at a fraction of the cost/time.)

If not part of the initial scope of this issue, then at least as something we'd want to check for in the future.

[...]

PR #261, opened 2017-11-21
[DOCS] Added --nolinkcheck parameter

lcawl:
Related to https://github.com/elastic/docs/issues/249

This PR adds the --nolinkcheck parameter, which skips the step that checks links at the end of a --all build.  For example, you can use it as follows:

> ./build_docs.pl —all —nolinkcheck
...
Writing main TOC
Writing extra HTML redirects
Skipped Checking links

This is required for when a broken link is preventing us from successfully publishing the documentation in a time-critical situation.  

The expectation is that any broken links would be fixed as soon as possible and link checking would be resumed immediately thereafter.

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -51,7 +51,7 @@ GetOptions(
     $Opts,    #
     'all', 'push', 'update!',    #
     'single',  'pdf',     'doc=s',           'out=s',  'toc', 'chunk=i',
-    'open',    'nolinkcheck', 'staging', 'procs=i',         'user=s', 'lang=s',
+    'open',    'skiplinkcheck', 'staging', 'procs=i',         'user=s', 'lang=s',
     'lenient', 'verbose', 'reload_template', 'resource=s@'
 ) || exit usage();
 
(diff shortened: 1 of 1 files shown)


debadair:
I'd pull the push out of the nolinkcheck if/else, other than that LGTM.

lcawl:
Good suggestion, @debadair, done!

build_docs.pl
```diff
@@ -191,12 +191,15 @@ sub build_all {
...
-        if $Opts->{push};
+    if ( $Opts->{nolinkcheck} ) {
+      say "Skipped Checking links";
+      }
+     else {
+       say "Checking links";
+       check_links($build_dir);
+         }
+     push_changes($build_dir)
+     if $Opts->{push};
```
clintongormley: niggle - indent the `if` line

build_docs.pl
```diff
@@ -51,7 +51,7 @@ BEGIN
     $Opts,    #
     'all', 'push', 'update!',    #
     'single',  'pdf',     'doc=s',           'out=s',  'toc', 'chunk=i',
-    'open',    'staging', 'procs=i',         'user=s', 'lang=s',
+    'open',    'nolinkcheck', 'staging', 'procs=i',         'user=s', 'lang=s',
```
clintongormley: let's call it `skiplinkcheck` instead of `nolinkcheck`

PR #269, opened 2017-12-06
[DOCS] Add kibana-extra folder to conf.yaml

lcawl:
Adds the kibana-extra folder to the paths in the conf.yaml

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -349,6 +349,7 @@ contents:
                 path:   docs/en
               -
                 repo:   x-pack-kibana
+                prefix: kibana-extra/x-pack-kibana
                 path:   docs/en
                 exclude_branches:   [ 5.3, 5.2, 5.1, 5.0]
               -
@@ -484,7 +485,7 @@ contents:
             prefix:     en/kibana
             current:    6.1
             branches:   [ master, 6.x, 6.1, 6.0, 5.6, 5.5, 5.4, 5.3, 5.2, 5.1, 5.0, 4.6, 4.5, 4.4, 4.3, 4.2, 4.1, 4.0, 3.0 ]
-            index:      ../x-pack-kibana/docs/en/index.asciidoc
+            index:      ../kibana-extra/x-pack-kibana/docs/en/index.asciidoc
             chunk:      1
             tags:       Kibana/Reference
             sources:
@@ -494,6 +495,7 @@ contents:
                 path:   docs/
               -
                 repo:   x-pack-kibana
+                prefix: kibana-extra/x-pack-kibana
                 path:   docs/en
               -
                 repo:   x-pack-elasticsearch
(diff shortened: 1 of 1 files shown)


kjbekkelund:
@lcawl We will only be backporting this to 6.x, so maybe we need to use `exclude_branches` to exclude all earlier versions?

lcawl:
I think that adding exclude_versions would work for most cases in the conf.yaml, but I'm not sure about what to put for the Kibana Reference's index value:

>   ...
> title:      Kibana: Explore, Visualize, and Share
>         sections:
>           -
>             title:      Kibana Reference
>             prefix:     en/kibana
>             current:    6.0
>             branches:   [ master, 6.x, 6.1, 6.0, 5.6, 5.5, 5.4, 5.3, 5.2, 5.1, 5.0, 4.6, 4.5, 4.4, 4.3, 4.2, 4.1, 4.0, 3.0 ]
>             index:      ../kibana-extra/x-pack-kibana/docs/en/index.asciidoc
> ...

 @clintongormley If the path to the index.asciidoc is going to differ between 6.1 and later versions (due to addition of "kibana-extra" folder), do I need to split that content into a new book in the conf.yaml?

lcawl:
It looks like in x-pack-elasticsearch the elasticsearch-extra folder was added all the way back to 0.90: https://github.com/elastic/x-pack-elasticsearch/tree/0.90

clintongormley:
> @clintongormley If the path to the index.asciidoc is going to differ between 6.1 and later versions (due to addition of "kibana-extra" folder), do I need to split that content into a new book in the conf.yaml?

Yes, you can't have different paths to index.asciidoc.  As suggested above, the alternative is to make the same change back to 0.90

kjbekkelund:
@epixa: @lcawl asked me on Slack "Will it be problematic if we add the kibana-extra folder all the way back to 3.0 for docs purposes?" I don't really have an idea about these docs builds, so I'm not sure what effect that has. Do you know what would be okey or not for Kibana?

epixa:
I'm not an expert at our docs system or anything, but I don't think it's going to be a problem to use `kibana-extra` for `x-pack-kibana` all the way back to the 3.0 branch.  6.x introduced the *requirement* to have x-pack-kibana inside a kibana-extra directory, but it was *possible* to use that same setup prior to 6.x as well.  And even if it weren't possible, the docs system doesn't run Kibana or anything like that, it just needs to know where to find the relevant asciidoc files, so it shouldn't be a problem to have the repo cloned to a different place.

lcawl:
Great, thanks, @epixa.

kjbekkelund:
Hm, I wonder if we don't need this change if we revert the change in https://github.com/elastic/x-pack-kibana/pull/4066/files#diff-de10d134c6b3aa58eba72abd79ef45ba (so we don't make any changes to the docs paths at all).

kjbekkelund:
I merged that change, and it seemed to build docs properly. I can create a separate commit to x-pack-kibana that changes the docs path, but as long as docs are working as expected now, it's not something we need to do immediately, I think?

PR #274, opened 2017-12-15
set current branch of apm-agent-python to "1.x"

beniwohli:
The master branch will see some incompatibilities landed between
now and 6.2 (when APM goes GA). This should avoid that those
incompatibilities land on
https://www.elastic.co/guide/en/apm/agent/python/current/index.html
before GA.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -773,8 +773,8 @@ contents:
               -
                 title:      APM Python Agent
                 prefix:     python
-                current:    master
-                branches:   [ master ]
+                current:    1.x
+                branches:   [ master, 2.x, 1.x ]
                 index:      docs/index.asciidoc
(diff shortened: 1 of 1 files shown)


beniwohli:
Just saw an email from @drewr regarding slashes in branch names. Could this be a problem here?

beniwohli:
I renamed the branch from `releases/1.x` to `1.x` now.

debadair:
As a general rule, we've gone to publishing docs for upcoming releases in addition to the "current" release. I'd switch this to:

```
current:    1.x
branches:   [ master, 2.x, 1.x ]
```

Either way, as long as it builds locally with your changes, LGTM.

beniwohli:
Sounds good, thanks! I'll try and build it with all three branches locally, and merge if it works as expected

PR #280, opened 2018-01-05
Add configuration for the Ruby APM Agent docs

mikker:
Initial docs in https://github.com/elastic/apm-agent-ruby/pull/33

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -10,6 +10,7 @@ repos:
     apm-server:           https://github.com/elastic/apm-server.git
     apm-agent-nodejs:     https://github.com/elastic/apm-agent-nodejs.git
     apm-agent-python:     https://github.com/elastic/apm-agent-python.git
+    apm-agent-ruby:       https://github.com/elastic/apm-agent-ruby.git
     beats:                https://github.com/elastic/beats.git
     cloud:                https://github.com/elastic/cloud.git
     curator:              https://github.com/elastic/curator.git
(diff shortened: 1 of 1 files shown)


mikker:
Hi @dedemorton 👋 @beniwohli suggested I talk to you about getting this merged.

conf.yaml
```diff
@@ -782,6 +783,18 @@ contents:
                   -
                     repo:   apm-agent-python
                     path:   docs
+              -
+                title:      APM Ruby Agent
+                prefix:     python
+                current:    1.x
+                branches:   [ master ]
```
dedemorton: You've set `current` to 1.x, but you haven't included 1.x in `branches`. It should say:

```
branches: [ master, 1.x ]
```

This assumes that a 1.x branch exists. If you simply want to rename master in the drop-down list that appears in the published docs, there's a way to do that, too. I can't remember exactly how to do that, but I'll dig throguh my notes if that's what you need.

dedemorton: Aha, so 1.x does not exist. I'm guessing you want the label to say 1.x. I need to poke around a bit because I think our method for setting the label has changed since I last did that.

dedemorton: Looking at previous examples, I think you just need to use  `{ master: 1.x}` instead of `master` when you set `branches`. So you would have: 

```
current:    master
branches:   [ { master: 1.x } ]
```

conf.yaml
```diff
@@ -782,6 +783,18 @@ contents:
                   -
                     repo:   apm-agent-python
                     path:   docs
+              -
+                title:      APM Ruby Agent
+                prefix:     python
+                current:    1.x
```
dedemorton: Change to say `current: master`.

You can disregard my previous comments. I deleted them when I realized that you don't really need to override the label because you won't see a label when you have only one branch.

mikker:
Should be better now :)

mikker:
Thanks! 🎉

PR #287, opened 2018-01-15
apm-agent-nodejs: add 0.x and 1.x branch

watson:
In anticipation of the upcoming release of version 1.0.0 of the Node.js APM Agent we have made a `1.x` branch. In the upcoming Kibana 6.2 we have embedded links to the `1.x` version of the Node agent docs, so it's important that this is merged in before the 6.2 release.

Once version 1.0.0 is released I'll make a PR to switch `current` to point to `1.x` instead.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -764,8 +764,8 @@ contents:
               -
                 title:      APM Node.js Agent
                 prefix:     nodejs
-                current:    master
-                branches:   [ master ]
+                current:    0.x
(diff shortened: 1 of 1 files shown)


watson:
Just updated the PR with a `0.x` branch as well as we'll need to have documentation available to the users running the beta version of the agent. If not we would have lost those docs once we released 1.0.0

dedemorton:
LGTM.

PR #289, opened 2018-01-17
Check out client documentation test for java api docs

cbuescher:
This is required to include core snippets from core integration tests using the
transport client API in the documentation.

Opening this to get feedback about if this is the correct way of doing this (looks similar to what was previously done to include code snippets from client/rest-high-level/...). I would need this before merging https://github.com/elastic/elasticsearch/pull/28260


cbuescher:
@clintongormley Did something happen to the docs branch or was the PR so bad ;-) I guess I'll have to open it again in any case.

clintongormley:
sorry @cbuescher - we dropped all the history because it was 1.6GB.  I didn't think to check open PRs

PR #290, opened 2018-01-17
[DOCS] Added x-pack-kibana to jp and kr

lcawl:
The Japanese and Korean versions of the X-Pack Reference have dependencies on the x-pack-kibana repository, so this PR adds those dependencies to the conf.yaml.


debadair:
LGTM

PR #294, opened 2018-01-25
Add versioned plugin docs to the build

dedemorton:
Don't merge this yet. Waiting for final approval from dev team.

Adds the versioned plugin reference to the documentation build.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -534,6 +534,19 @@ contents:
                 repo:   logstash-docs
                 path:   docs/
 
+          -
+            title:      Logstash Versioned Plugin Reference
+            prefix:     en/logstash/plugins
+            current:    versioned_plugin_docs
+            branches:   [ versioned_plugin_docs ]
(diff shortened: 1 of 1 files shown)


andrewvc:
@dedemorton my one request is that we add

`<meta name="robots" content="noindex,nofollow>`

to all these pages. We want search to point at the main docs, not these.

Other than that, I really like the way this turned out. Once that's in, good to merge.

dedemorton:
@andrewvc I checked with the web team, and they say that I need to do this on my end.

I don't see any backend attributes (in asciidoc) or elements (in DocBook) that I can use to inject this meta content into the generated html pages.

@clintongormley might have a solution, but I'm stumped.

Do we need to fix this before we can publish?

clintongormley:
@dedemorton I've added a new `noindex` option which can be specified alongside `chunk` in the conf.yaml.  When true, it will add this meta header.

See https://github.com/elastic/docs/commit/f4d4ee190842150673dd7b2b60419745f3138332

PR #300, opened 2018-01-31
Check out client documentation test for java api docs

cbuescher:
This is required to include core snippets from core integration tests using the
transport client API in the documentation.

Opening this to get feedback about if this is the correct way of doing this (looks similar to what was previously done to include code snippets from client/rest-high-level/...). I would need this before merging elastic/elasticsearch#28260


cbuescher:
Closing since #316 includes this directory already now

PR #276, opened 2017-12-20
Escape raw versions of replacement characters in README

watson:
Old rendering:

image

New rendering:

image

PR #301, opened 2018-01-31
Escape raw versions of replacement characters in README

watson:
Reopen of #276

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -560,13 +560,24 @@ syntax highlighting.
 Inline text can be formatted as follows:
 
 [horizontal]
-`_emphasis_`::              _emphasis_
-`*bold*`::                  *bold*
-+\`mono`+::                 `mono`
-+\`\`double quoted''+::     ``double quoted''
-+\`single quoted'+::        `single quoted'
-`^superscript^`::           ^superscript^
-`~subscript~`::             ~subscript~
+ifdef::env-github[]
+`&#x5f;emphasis_`::         _emphasis_
+`&#x2a;bold*`::             *bold*
+`&#x60;mono'`::             `mono`
+`&#x60;`double quoted''`::  ``double quoted''
+`&#x60;single quoted'`::    `single quoted'
+`&#x5e;superscript^`::      ^superscript^
+`&#x7e;subscript~`::        ~subscript~
+endif::[]
+ifndef::env-github[]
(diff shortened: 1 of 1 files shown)


clintongormley:
@watson Actually this doesn't work as expected. Github uses Asciidoctor to render asciidoc, but we use Asciidoc, which ends up producing this:

image

I hoped that using `+`  instead of ` might work for both, but asciidoc ends up replacing the characters.

The following works for asciidoc and (hopefully) asciidoctor too:

```
[horizontal]
`--`::      -- (em dash)
+&#x2e;..+::    ...
+&#x2d;>+::     ->
+<&#x2d;+::     <-
+&#x3d;>+::     =>
+<&#x3d;+::     <=
+&#x28;C)+::    (C)
+&#x28;TM)+::   (TM)
+&#x28;R)+::    (R)
```

watson:
Is the README rendered and displayed in other locations than on GitHub?

clintongormley:
> Is the README rendered and displayed in other locations than on GitHub?

It is intended to be built locally with the build_docs tool - it just so happens that it also renders on github.  other than that, no

watson:
Updated.

I had to change `+` for `` ` `` though. Not sure what the `+` was for?

clintongormley:
Nope, in asciidoc this now renders as 

image

`+` and ```  are different forms of passthrough in asciidoc.  This is yet another place that asciidoctor has diverged from asciidoc.  Given that the primary purpose of this document is to be rendered with asciidoc (and that display on github is secondary), I think we shouldn't merge this PR.

watson:
I disagree. If there's no way to mark this up so it renders correctly on both GitHub and if you clone the repo and build it locally, IMHO I think we should optimize for reading the README on GitHub as this is the place most people will read it. I imagine only people who work with docs everyday will view it locally, and this documentation is targeted people who are new to working with asciidocs and hence will view it on GitHub.

clintongormley:
You know what we can do?  Just drop the monospacing.  So remove the `` ` ``, the `+`, and the html encoding, and just use the `\` escaping as per your original PR.  That SHOULD render properly for both.

watson:
Updated using conditional branching and validated both locally and on github. It looks ok on both using this technique. Is that ok?

clintongormley:
thanks @watson

PR #309, opened 2018-02-06
Adds formatting support for Kibana APIs

tylersmalley:
This was added in a similar format to the already existing code, though I did some minor cleaning up to better understand it. It still references Sense, which I am not sure if we can remove that. If we no longer build docs referencing Sense examples, then we should.

This code could really benefit from a more thorough cleanup or possible rewrite in the future.

screenshot 2018-02-06 07 48 13

screenshot 2018-02-06 07 48 29

Also fixed Copy as cURL commands to include the hostname configured through the settings.

### DIFF
--- a/resources/web/docs.js
+++ b/resources/web/docs.js
@@ -152,11 +152,10 @@
 
 // END jscookie
 
-
 jQuery(function() {
-
   var lang = jQuery('section#guide[lang]').attr('lang') || 'en';
   var Strings;
+
   if (lang === 'en') {
     Strings = {
       "Configure Console URL" : "Configure Console URL",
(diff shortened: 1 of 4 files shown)


tylersmalley:
This is a work in progress until I get the corresponding Kibana PR up to provide complete testing.

tylersmalley:
I need to revert the template literals change here.

tylersmalley:
Reverted template literals. Working on getting API PR up on Kibana.

resources/web/docs.js
```diff
@@ -303,8 +317,14 @@ jQuery(function() {
           curlText += comment + '\n';
         } else {
           path = path.replace(/^\//, '').replace(/\s+$/,'');
-          path += path.includes('?') ? '&pretty' : '?pretty';
-          curlText += 'curl -X' + method + " 'localhost:9200/" + path + "'";
```
tylersmalley: `localhost:9200` was hardcoded here. I updated it to include the host/port configured through the settings.

debadair:
LGTM

We do still publish docs that reference Sense, so for the time being that's still valid.

PR #314, opened 2018-02-13
Added resources for generating standalone docs.

debadair:
The `offline_docs.sh` script modifies the Elastic docs so that they can be viewed locally or hosted on an internal web server without access to the Internet. The required CSS, Javascript, image resources, and our legal disclaimers are loaded from static copies of those files.

The intent is to make a zip file of the standalone docs available to customers who use Elasticsearch in an air-gapped environment and submit a support request for offline versions of our documentation. Ultimately, we'll want to push the zip file to an S3 bucket and update it in place when we ship new releases. 

Publicly hosting the documentation is expressly forbidden by our existing conditions of use:

> The documents produced by this build process may be published only on http://www.elastic.co. They may not be published in any other form or on any other website without explicit permission.

**IMPORTANT**: This script currently modifies the generated docs in-place. (So as to not replicate gigabytes of data.) The resulting zip file is still about 2 GB. 

To run the script to and launch SimpleHTTPServer to view the docs, specify:

`./offline_docs.sh -o`

Specify the `-z` option to generate the zip file. By default, the script loads the resources from `docs/resources/offline` and processes the files in the `docs/html` directory. You can also explicitly specify the directory that contains the docs you want to process and the location of the static resources.


offline_docs.sh
```diff
@@ -0,0 +1,76 @@
...
+  echo "No HTML directory specified. Defaulting to 'html'."
+  HTML_DIR="html"
+fi
+
+if [[ $RESOURCE_DIR == "" ]] ; then
+  echo "No resource directory specified. Defaulting to 'resources/offline'."
+  RESOURCE_DIR="resources/offline"
+fi
+
+printf "Warning: This will modify the contents of $HTML_DIR in place. Continue? y/N: "
```
dedemorton: Might want to be more explicit here about what you mean by "in place." I suspect this means the script will modify whatever files are in the specified directory (or html, by default), but that's not totally clear. 
Maybe the script should be set to clone the docs repo (y/N) to get the latest version of the html files?

offline_docs.sh
```diff
@@ -0,0 +1,76 @@
+#!/bin/bash
+# Modify generated docs in a particular directory to work offline.
+
+date
+
+USAGE="Usage: offline_docs.sh [-d <dir-name>] [-r <dir-name>] [-z] [-o] [-h]."
```
dedemorton: I'm guessing you plan to add more help content here? 

Might be nice to mention that the script creates a zip file called elastic-docs.zip, or give the user the option of setting the name (nice-to-have).

lcawl:
The script succeeded for me, but I've found one link so far that's still trying to go the external site: 
http://0.0.0.0:8000/en/x-pack/current/installing-xpack.html
See the "Installing Logstash" bullet. 
Maybe that's outside the purview of this script, actually, but it will be hard for users to tell which links work (i.e. are local/peer) and which won't (i.e. are external).

clintongormley:
> Didn't notice any issues, except that the pages loaded more slowly than the website (probably because the website is optimized with caching and all that stuff).

That's probably because the SimpleHTTPServer only handles a single request at a time.  It will probably be faster to just open `index.html` in the web browser directly from the file system.  Not sure if that will interfere with the scripts that run.

> The required CSS, Javascript, image resources, and our legal disclaimers are loaded from static copies of those files.

Are you sure about this?  I know the template used to use javascript to handle a lot of layout, and download more JS, CSS, and images that weren't visible in the original HTML.  Are you sure that no other files are fetched remotely after the initial load?

offline_docs.sh
```diff
@@ -0,0 +1,85 @@
...
+  echo "Copying offline resources from $RESOURCE_DIR into $HTML_DIR"
+  cp -r $RESOURCE_DIR/. $HTML_DIR
+
+  cd $HTML_DIR
+
+  echo "Processing files."
+  echo "A LOT of files."
+  echo "This is going to take ten minutes or so."
+
+  find . -name '*.html' -exec sed -i '' -e 's/href="https:\/\/www\.elastic\.co\/guide/href="/g' -e  's/privacy-and-cookie-policy"/privacy-and-cookie-policy\.html"/g' -e 's/terms-of-use"/terms-of-use\.html"/g' -e 's/trademarks"/trademarks\.html"/g' -e 's/privacy-policy"/privacy-policy\.html"/g' -e 's/url="\/guide"/url="\/index.html"/g' -e 's/href="\/guide"/href="\/index.html"/g'  {} \;
```
Mogztter: `\;` will execute the `sed` command for each found result.
So if you found 5000 files, the `sed` command will be executed 5000 times (once per file).

```
sed [...] file1
sed [...] file2
...
```

To improve performance, you could replace `\;` with `\+`.

`\+` will concatenate the found results and will execute the `sed` command only once.

```
sed [...] file1 file2 ...
```

Here are the results on my laptop:

**find with -exec \\;**
```
time find html -name '*.html' -exec sed -i '' -e 's/href="https:\/\/www\.elastic\.co\/guide/href="/g' -e  's/privacy-and-cookie-policy"/privacy-and-cookie-policy\.html"/g' -e 's/terms-of-use"/terms-of-use\.html"/g' -e 's/trademarks"/trademarks\.html"/g' -e 's/privacy-policy"/privacy-policy\.html"/g' -e 's/url="\/guide"/url="\/index.html"/g' -e 's/href="\/guide"/href="\/index.html"/g'  {} \;
```
```
real	7m40.897s
user	5m42.922s
sys	1m14.886s
```

**find with -exec \\+**
```
time find html -name '*.html' -exec sed -i '' -e 's/href="https:\/\/www\.elastic\.co\/guide/href="/g' -e  's/privacy-and-cookie-policy"/privacy-and-cookie-policy\.html"/g' -e 's/terms-of-use"/terms-of-use\.html"/g' -e 's/trademarks"/trademarks\.html"/g' -e 's/privacy-policy"/privacy-policy\.html"/g' -e 's/url="\/guide"/url="\/index.html"/g' -e 's/href="\/guide"/href="\/index.html"/g'  {} \+
```
```
real	5m9.361s
user	4m47.892s
sys	0m19.930s
```

If you want to go one step further, you could use `xargs` to do `sed` in parallel mode.

In this example, I'm using 4 processes (on my 4 cores laptop):

**find with xargs parallel**
```
time find html -name '*.html' -type f -print0 | xargs -0 -P4 sed -i '' -e 's/href="https:\/\/www\.elastic\.co\/guide/href="/g' -e  's/privacy-and-cookie-policy"/privacy-and-cookie-policy\.html"/g' -e 's/terms-of-use"/terms-of-use\.html"/g' -e 's/trademarks"/trademarks\.html"/g' -e 's/privacy-policy"/privacy-policy\.html"/g' -e 's/url="\/guide"/url="\/index.html"/g' -e 's/href="\/guide"/href="\/index.html"/g'
```
```
real	1m32.393s
user	5m22.005s
sys	0m21.567s
```

Please note that most of your resources will be taken during this time (and you laptop may be noisy) 😉 

NOTE: if no `-n[int]` is specified, xargs uses the default of -n5000 (see `man xargs`).
This means that xargs uses up to 5000 parameters for the command and executes it once, instead of 5000 times.


Reference: https://www.everythingcli.org/find-exec-vs-find-xargs/

PR #317, opened 2018-03-06
Add support for """ strings in copy-as-curl

nik9000:
Adds support to copy-as-curl for `"""` delimited strings. `"""` delimited
strings are natively supported by Kibana's CONSOLE but don't work
properly in Elasticsearch unless you convert them into normal JSON. This
adds support to `copy-as-curl` to do that conversion.


nik9000:
Seeing no objections, I suppose it is time merge this.

PR #322, opened 2018-03-16
Clean up """ed strings in copy-as-curl

nik9000:
Clean up """ed strings output by copy-as-curl so they look fairly good.

* Trim leading newlines
* Trim leading whitespace from the first line
* Trim the same amount of whitespace from each subsequent line
* Trim trailing whitespace

### DIFF
--- a/resources/web/docs.js
+++ b/resources/web/docs.js
@@ -312,19 +312,36 @@ jQuery(function() {
             curlText += " -d'";
             var start = body.indexOf('"""');
             if (start < 0) {
-                curlText += body;
+              curlText += body;
             } else {
               var startOfNormal = 0;
               while (start >= 0) {
                 var end = body.indexOf('"""', start + 3);
                 if (end < 0) {
-                    end = body.length();
+                  end = body.length();
                 }
                 curlText += body.substring(startOfNormal, start);
                 curlText += '"';
-                curlText += body.substring(start + 3, end)
+                var quoteBody = body.substring(start + 3, end);
+                // Trim leading newline if there is one
+                quoteBody = quoteBody.replace(/^\n+/, '');
+                // Trim leading whitespace off of each line
+                // But not more whitespace than is on the first line
+                var leadingWhitespace = quoteBody.search(/\S/);
+                if (leadingWhitespace > 0) {
+                  var leadingString = '^'
+                  for (var i = 0; i < leadingWhitespace; i++) {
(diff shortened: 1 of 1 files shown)


archanid:
Can you give an example before and after output as a result of these changes? I'm curious why we do `Trim the same amount of whitespace from each subsequent line`

talevy:
@archanid I was wondering the same thing, so I tried it here:

https://jsfiddle.net/vgsbms7a/16/

nik9000:
Sure!

This is what we are transforming:
```
GET hockey/_search
{
  "query": {
    "function_score": {
      "script_score": {
        "script": {
          "lang": "painless",
          "source": """
            int total = 0;
            for (int i = 0; i < doc['goals'].length; ++i) {
              total += doc['goals'][i];
            }
            return total;
          """
        }
      }
    }
  }
}
```

The change in this PR is around the `"""`ed field.

Here is the output with this PR:
```
curl -XGET 'localhost:9200/hockey/_search?pretty' -H 'Content-Type: application/json' -d'
{
  "query": {
    "function_score": {
      "script_score": {
        "script": {
          "lang": "painless",
          "source": "int total = 0;\nfor (int i = 0; i < doc[\u0027goals\u0027].length; ++i) {\n  total += doc[\u0027goals\u0027][i];\n}\nreturn total;"
        }
      }
    }
  }
}
'
```

And here is what it looks like without this PR:

```
curl -XGET 'localhost:9200/hockey/_search?pretty' -H 'Content-Type: application/json' -d'
{
  "query": {
    "function_score": {
      "script_score": {
        "script": {
          "lang": "painless",
          "source": "\n            int total = 0;\n            for (int i = 0; i < doc[\u0027goals\u0027].length; ++i) {\n              total += doc[\u0027goals\u0027][i];\n            }\n            return total;\n          "
        }
      }
    }
  }
}
'
```

talevy:
should we be trimming trailing whitespace from each line?

nik9000:
> should we be trimming trailing whitespace from each line?

Trailing whitespace isn't a normal thing in the examples though. So I don't think it is worth it.

talevy:
good enough for me

nik9000:
Thanks for reviewing @archanid and @talevy! I'm going to get this merged so we can have nicer copy-as-curls for now. If you find anything else I'd be happy to change it!

Issue #161, opened 2017-04-06
Improve error message when there is an authentication failure


nik9000:
When you have multi factor auth turned out `build_docs.pl --all` doesn't need a *password*. It needs an authentication token.

dedemorton:
Closed by #328

PR #328, opened 2018-03-22
Improve error message for auth failure

dedemorton:
Resolves issue https://github.com/elastic/docs/issues/161

@nik9000 Is this sufficient to address your concern? I don't see an actual error...just keeps restarting. I don't want to mess too much with my keychain, tho, so maybe my testing is not thorough enough. Here's what the output looks like: first when I enter the wrong credentials, and then when I enter no password.

image

In the spirit of the existing text, I've tried to keep the messages short.

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -159,7 +159,7 @@ sub build_local_pdf {
 #===================================
 sub build_all {
 #===================================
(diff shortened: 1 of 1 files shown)


nik9000:
Thanks!

debadair:
LGTM

PR #334, opened 2018-04-19
Uses localhost:9200 for ES copy as curl

tylersmalley:
In #309 we began using the configured host for the "Copy as Curl" commands. However, the currently configured host is the Sense/Console host and not the ES host. Until we allow the ES host to be configurable in the way the Console/Sense host is, we will go back to using `localhost:9200`


debadair:
LGTM

I'll open a docs issue for allowing users to configure the ES host--that would indeed be helpful.

PR #344, opened 2018-04-25
Add info about shared attributes

dedemorton:
Devs and contributors are confused about how to do this now. I think we need to continue to mention that attributes can be defined in under the book title (until we get the whole library cleaned up), or people will be confused.


dedemorton:
@lcawl If this looks good when you review it, feel free to merge it directly.

PR #362, opened 2018-05-26
Fix Cloud branches

nrichers:
Looks like https://github.com/elastic/docs/pull/361 introduced an error into `conf.yaml`. This is the fix. I've hung my head in shame.


lcawl:
Yay!  LGTM

PR #364, opened 2018-05-29
conf: add apm-agent-java repo

felixbarny:


felixbarny:
relates to https://github.com/elastic/apm-agent-java/issues/19

dedemorton:
@felixbarny I ran a build and the source files returned errors. It’s not clear to me why you are using passthrough macros, but you can’t use them within back ticks, like this:  `pass:[true]`. You get errors:

```
asciidoc: WARNING: configuration.asciidoc: line 51: nested inline passthrough
asciidoc: WARNING: configuration.asciidoc: line 73: nested inline passthrough
asciidoc: WARNING: configuration.asciidoc: line 156: nested inline passthrough
asciidoc: WARNING: configuration.asciidoc: line 180: nested inline passthrough
asciidoc: WARNING: configuration.asciidoc: line 215: nested inline passthrough
asciidoc: WARNING: configuration.asciidoc: line 265: nested inline passthrough
asciidoc: WARNING: configuration.asciidoc: line 292: nested inline passthrough
asciidoc: WARNING: configuration.asciidoc: line 364: nested inline passthrough
asciidoc: WARNING: configuration.asciidoc: line 386: nested inline passthrough
asciidoc: WARNING: configuration.asciidoc: line 408: nested inline passthrough
asciidoc: WARNING: configuration.asciidoc: line 432: nested inline passthrough
asciidoc: WARNING: configuration.asciidoc: line 458: nested inline passthrough
asciidoc: WARNING: configuration.asciidoc: line 500: nested inline passthrough
asciidoc: WARNING: configuration.asciidoc: line 524: nested inline passthrough
```

If you need to use the passthrough, try using plus marks instead. For example:

```
 +pass:[true]+
```

Let me know when you've fixed the file and I'll take another look at this.

felixbarny:
Fixed in https://github.com/elastic/apm-agent-java/commit/06e47ea62e1a686b6316e5ef8127c6261aeea501

Could you give it another go?

dedemorton:
OK, I'll run the build again, but what I meant is that you should either use plus signs (instead of back ticks) so that the [passthrough macro](http://www.methods.co.nz/asciidoc/userguide.html#X77) gets evaluated, or remove the passthrough entirely. (You only need to use plus signs when you have monospaced text that contains inline substitutions.) What you have now renders like this:

image

Since you've removed the passthrough macro, you can just use back ticks (with no plus signs):

```
`50`
```

felixbarny:
Oh, the asciidoc preview from my editor evaluated the things inside backticks, which made things like `/VAADIN/*, (?i)/heartbeat/*, /favicon.ico, *.js, *.css, *.jpg, *.jpeg, *.png, *.webp, *.svg, *.woff, *.woff2` not seem to work. But with the build it's working fine.

Fixed it.

PR #370, opened 2018-06-04
shared/attributes: add apm-server-ref-62

axw:
Add the apm-server-ref-62 attribute for referring
to version 6.2 specifically, when another branch
is used generally.

(In the Frontend JavaScript APM Agent docs
we use "current" generally, but make specific
reference to 6.2 in one case.)


axw:
Is there a better approach to this? Feels a little clumsy - would be nice to be able to override the {branch} value where {apm-server-ref} is used. I took my cue from {ref-60}.

axw:
> LGTM! Checked the URL you added as well, works. I don't know what your use case is for using links to a specific version, but I expect you do. slightly_smiling_face

This: https://github.com/elastic/apm-agent-js-base/pull/24/files#diff-b40e969cab8a3a76759352587744c4caR12

We want to point to the more recent docs by default, but call out a requirement when using 6.2 specifically.

nrichers:
@axw Makes sense. We do the same on Cloud where we need to version links.

PR #372, opened 2018-06-06
Add note about --resources parameter

cbuescher:
After the x-pack merge, building the Elasticsearch reference documentation
requires specifying additional resources using the `--resource` parameter.
This change adds a note to check the doc_build_aliases.sh file if ommiting this
parameter leads to errors and adds an example for building the reference
documentation.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -95,6 +95,14 @@ build_docs.pl --doc path/to/index.asciidoc
 To use simplified aliases for the build commands for each book, see
 `doc_build_aliases.sh`. 
 
+NOTE: Some documentation parts, e.g. the Elasticseach reference
+documentation, may need additional resources specified with the 
+`--resource` parameter. If you get errors with a particular book,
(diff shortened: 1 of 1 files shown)


cbuescher:
Since I ran into this and it took me a bit to understand whats wrong I think some sort of note is useful. Maybe there are other places where this fits in better, please let me know.

lcawl:
@cbuescher That build command will change again shortly (i.e. the --resources parameter will no longer be required when https://github.com/elastic/elasticsearch/issues/30665 is complete).  In general I think it's best to encourage people to use that aliases file, since we keep that up-to-date with the changes.

cbuescher:
> That build command will change again shortly (i.e. the --resources parameter will no longer be required when elastic/elasticsearch#30665 is complete)

Didn't know that, I guess mentioning the alias file is suffiecient then, I just didn't know it yet. We can close this then.

PR #373, opened 2018-06-06
Add shared folder for Cloud docs

nrichers:
We started single-sourcing some content in a shared folder. Looks like we might need to add that folder for the builds to work.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -535,6 +535,10 @@ contents:
               -
                 repo:   cloud
                 path:   docs/saas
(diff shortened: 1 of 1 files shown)


lcawl:
LGTM.  You might find (e.g. if you add images in that shared folder) that you need to update the docbldece and docbldec commands in https://github.com/elastic/docs/blob/master/doc_build_aliases.sh too (to add --resource=$GIT_HOME/cloud/docs/shared)

PR #374, opened 2018-06-06
Add shared folder for Cloud docs

nrichers:
We have some images in a new shared folder that might require this PR.


lcawl:
LGTM . Thanks for updating that list!

PR #380, opened 2018-06-14
[DOCS] Creates a Legacy Documentation section

lcawl:
This PR creates a "Legacy Documentation" section, which contains the old X-Pack, Shield, Marvel, Watcher, and Reporting documentation.  Additional legacy content can be added to that section in subsequent PRs.


conf.yaml
```diff
@@ -1152,6 +1054,105 @@ contents:
                   path: /docs/kr
                   prefix: elasticsearch-extra/x-pack-elasticsearch
 
+    -
+        title:      Legacy Documentation
+        sections:
+          -
+            title:      X-Pack Reference
+            prefix:     en/x-pack
```
debadair: Do we want to add versions to the title like we did for Marvel, etc.? "X-Pack Reference for 6.0-6.2 and 5.x"

PR #390, opened 2018-07-02
[DOCS] Fix instructions for using build alias file.

Sue-Gallagher:
Correct and clarify instructions for setting GIT_HOME


debadair:
Agree with Lisa, otherwise LGTM

PR #391, opened 2018-07-05
Build docs for the 0.6 java agent branch

felixbarny:


felixbarny:
0.6 marks the beta release. The documentation should have a selection for this branch so that users don't get confused when I document features in master, which will go into 0.7.0. Having a 0.6 branch makes it clear which features are in 0.6.x. The disadvantage is that I have to cherry-pick doc changes which relate to both master and 0.6.x

PR #395, opened 2018-07-06
Add attribute for APM RUM agent.

simitt:
We need an additional attribute pointing to the real user monitoring agent, so we can refer to it from the APM Server docs.


simitt:
@dedemorton do I need to backport this to the `6.3` branch so I can reference it from the APM Server 6.3 docs?

dedemorton:
@simitt No, you don't have to backport. Bcause of how the doc build works (builds all versions), we only use master. If we need to have more than one version of a file, we create additional files rather than using branches.

PR #402, opened 2018-08-06
Beta boilerplate language update per Legal

kellyemurphy:
Per Carl Spataro, the language we use in the beta tag needs to be updated.


dedemorton:
LGTM

PR #404, opened 2018-08-10
update APM section title

jalvz:

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -754,7 +754,7 @@ contents:
         title:      APM
         sections:
           -
-            title:      Getting Started with Elastic APM
+            title:      APM overview
             prefix:     en/apm/get-started
             index:      docs/guide/index.asciidoc
             current:    6.4
--- a/conf.yaml
+++ b/conf.yaml
@@ -754,7 +754,7 @@ contents:
         title:      APM
         sections:
           -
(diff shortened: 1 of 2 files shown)


dedemorton:
@jalvz I'm concerned about renaming this document to "APM overview" because users might not realize that the document includes getting started content. We already have this problem with the Stack Overview. It was renamed hastily when X-Pack went away, but the name doesn't convey the depth and variety of content that the Stack Overview contains. We are meeting sometime soon (probably this week) to discuss renaming the book, so I would suggest waiting for a resolution on that question before renaming the APM book. 

Unless you plan to remove the installation content and have just the overview info.

jalvz:
Thanks @dedemorton 
This change was suggested by @makwarth , i think waiting for the meeting is a good idea

makwarth:
Waiting for the resolution sounds good. 

FYI, I suggested this change as we're introducing Release notes for APM, which don't feel right to place under "Getting started", so I went with "Overview", as it was used in other documents already.

dedemorton:
Go ahead and change the title  to use caps (as suggested above), and I'll merge this.

jalvz:
And im back, sorry for the long way @dedemorton, updated now!

makwarth:
thanks @jalvz

PR #419, opened 2018-08-27
Increase replica shards for Site/Docs Search

jarpy:
The public search interface currently runs on a three node cluster. This
patch increases replica shards from 1 to 2, in order to distribute
search load more evenly across the production cluster.

Relates: elastic/infra#6636


jarpy:
🔔 @drewr

PR #425, opened 2018-08-31
Add Azure Marketplace and ARM template docs

russcam:
This commit adds the Azure Marketplace and
Azure Resource Manager (ARM) template documentation
to the elastic.co documentation.

Closes https://github.com/elastic/azure-marketplace/issues/97

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -14,6 +14,7 @@ repos:
     apm-agent-js-base:    https://github.com/elastic/apm-agent-js-base.git
     apm-agent-go:         https://github.com/elastic/apm-agent-go.git
     apm-agent-java:       https://github.com/elastic/apm-agent-java.git
+    azure-marketplace:    https://github.com/elastic/azure-marketplace.git
     beats:                https://github.com/elastic/beats.git
     cloud:                https://github.com/elastic/cloud.git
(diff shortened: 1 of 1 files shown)


conf.yaml
```diff
@@ -89,6 +90,19 @@ contents:
               -
                 repo:   kibana
                 path:   docs
+          -
+            title:      Azure Marketplace and Resource Manager (ARM) template
```
lcawl: Would you be open to adding "Deploying the.." to this title?  Then it's less of a change when we eventually roll it into a Deployment Guide?  If yes, I can make that change.

conf.yaml
```diff
@@ -89,6 +90,19 @@ contents:
...
+            prefix:     azure-resource-manager
```
lcawl: Likewise I think we might want to tweak this prefix so that it makes sense once there is more than just azure in the book.  Doing so now means it's less likely that we'll have to fix broken links later.

conf.yaml
russcam: I'm good with this. How about "Deploying with..."?

conf.yaml
russcam: something like `deployment/azure-resource-manager`?

PR #438, opened 2018-10-02
ECE - Fixed incorrect product names.

kylec32:
Since Elastic Cloud Enterprise is not a public repository I figured the best way of sharing this issue with fix was to create this pull request. 

The labels in the glossary appear to be incorrect. "Elastic Cloud Enterprise" being applied to what should have been simply "Elastic Cloud" and "Elastic Cloud Enterprise Enterprise" to what should be simply Elastic Cloud Enterprise.


kylec32:
Looks like this was updated. 👍

PR #451, opened 2018-11-06
Removed the default "test build" value from the version variable.

debadair:
This removes the "test build" value from the DC.identifier tag for index pages that don't have an associated version. 

This is a bit of a hammer, but I'm not seeing anywhere were removing the default "test build" value for the version variable is going to blow things up. 

@brianjolly

### DIFF
--- a/lib/ES/Util.pm
+++ b/lib/ES/Util.pm
@@ -33,7 +33,7 @@ sub build_chunked {
     my ( $index, $dest, %opts ) = @_;
 
     my $chunk     = $opts{chunk}         || 0;
-    my $version   = $opts{version}       || 'test build';
+    my $version   = $opts{version}       || '';
     my $multi     = $opts{multi}         || 0;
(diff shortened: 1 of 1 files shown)


clintongormley:
Hi @debadair, what's the reason for removing this value? I added it for a reason (which now I can't remember :) ) but should only be present in local builds, not with build --all, unless you're seeing something different?

brianjolly:
Hi @clintongormley, We were hoping to remove that value on the index page to help solve this bug:
https://github.com/elastic/elastic-co-search/issues/34
If you do need the meta tag as-is we can look into another way to fix `elastic-co-search/issues/34`

Let me know, thanks!

clintongormley:
Ah thanks @brianjolly - I'll assign this to myself and come up with a better fix

brianjolly:
👍 Thank You @clintongormley!

PR #455, opened 2018-11-09
Add Infrastructure monitoring docs

dedemorton:
I'm open for recommendations about the prefix used to determine the URL.


debadair:
LGTM. I think the prefix is fine. The page name for the top-level page you might want to spell out “infrastructure-operations” and then introduce infraops & use that for the subpage anchors.

PR #456, opened 2018-11-09
Add journalbeat and functionbeat to the doc build

dedemorton:


debadair:
LGTM. I don't see any problems & it passes YAML validation.

PR #468, opened 2018-11-14
Update Ruby APM Agent branches

mikker:


mikker:
Thanks @kuisathaverat !

PR #472, opened 2018-11-14
[DOCS] Add aliases for APM books

lcawl:
This PR adds build commands for the APM books that were missing from the doc_build_aliases.sh file.  If there are better alias names, we can change them as needed.


debadair:
LGTM

PR #484, opened 2018-11-23
Fixed indent on line 723

krisATelastic:
The YAML format was incorrect and ssl.truststore was not indent correctly


krisATelastic:
Ahhh, sorry @nrichers - my bad. I've updated the source doc you linked and submitted a PR. This can be killed off.

PR #500, opened 2018-12-07
1240012443 fix

kcm:
Fix "1240012443" to be "12400/12443"


nrichers:
@kcm Thanks for trying to fix the docs! A couple of quick points:

* Everything deserves review, so we encourage folks not to simply merge fixes on their own. 🙂 
* You fixed the HTML output that gets overwritten every 30 mins or so. This problem needs to be fixed in the Asciidoc source, see https://github.com/elastic/cloud/edit/2.0/docs/cloud-enterprise/ce-prerequisites.asciidoc. Please create a new branch (do not commit directly) and tag @elastic/cloud-writers for review. 

Ping me on Slack if I can help with the process further.

kcm:
Sorry about that! It was such a quick and straightforward fix (I thought) that I even just used the GH editor to do it, then merged for simplicity. Will do.

PR #504, opened 2018-12-12
Allow building with asciidoctor

nik9000:
Adds support for `--asciidoctor` to build the docs with Asciidoctor[1]
instead of AsciiDoc[2]. This is the first step towards converting our
docs build to Asciidoctor which is better for a bunch of reasons:

1. Asciidoctor is under active development and AsciiDoc isn't.
2. Asciidoctor is way faster. Like 10x or 100x faster to go from
asciidoc->xml files. That is the slowest step in the docs build at this
point so we'd see a substantial performance improvement using Asciidoctor.
2. Asciidoctor supports emitting warnings on missing attributes! AsciiDoc
just drops the line containing the missing attribute which drives us
crazy.
3. Just about every search for AsciiDoc documentation lands you on
Asciidoctor's docs. And they are *slightly* different so this becomes a
big pain.
3. Asciidoc is much more pluggable.
4. We can run Asciidoc inside our Java and JavaScript based builds one
day. That'd give us very good testing of the docs build very quickly.

This change doesn't switch any of our documentation over to building
with Asciidoctor when you do `build_docs.pl --all`. We'll save that for
follow-up changes.

We pull a copy of Asciidoctor into this repo like we do with all of our
other dependencies. Unlike our other dependencies, Asciidoctor is a ruby
gem. So we pull it into the repo using `bundler`. Updating Asciidoctor
*ought* to be as simple as changing the Gemfile and running:

```
cd resources/asciidoctor
bundle install
git add .
git commit
```

So a few things don't work right now:
1. All of our custom macros for stuff like `beta` and to make sure the
edit link shows up in the heading.
2. Actually, every heading is a little broken.
4. We never run anything like a2x's "find resources" step. This is
almost certainly required. But `build_docs.pl --asciidoctor --open`
works and that is mostly what I'm looking for at this point.

Some of that comes from Asciidoctor not supporting `asciidoc.conf`
files. The rest come from `a2x` not supporting Asciidoctor. Instead we
run the two steps in the perl code directly.

[1]: https://asciidoctor.org/
[2]: http://asciidoc.org/


.gitignore
```diff
@@ -16,3 +16,4 @@ bin
 !html/sitemap.xml
 metastore_db
 html_docs/
+!resources/asciidoctor/**/bin
```
nik9000: Bundler builds some bin files that we use.

build_docs.pl
```diff
@@ -52,7 +52,7 @@ BEGIN
     'all', 'push', 'update!',    #
     'single',  'pdf',     'doc=s',           'out=s',  'toc', 'chunk=i',
     'open',    'skiplinkcheck', 'linkcheckonly', 'staging', 'procs=i',         'user=s', 'lang=s',
-    'lenient', 'verbose', 'reload_template', 'resource=s@'
+    'lenient', 'verbose', 'reload_template', 'resource=s@', 'asciidoctor'
```
nik9000: I have no idea how to write idiomatic perl. But the perl in this PR seems to work.

lib/ES/Util.pm
```diff
@@ -45,42 +45,88 @@ sub build_chunked {
...
+            "local.book.version"       => $version,
+            "local.book.multi_version" => $multi,
+            "local.page.header"        => $page_header,
+            "local.book.section.title" => "Learn/Docs/$section",
+            "local.book.subject"       => $subject,
+            "local.noindex"            => $noindex,
+    );
+
     my ( $output, $died );
-    eval {
```
nik9000: Moved into the `else` arm of the `if` statement below.

lib/ES/Util.pm
```diff
@@ -45,42 +45,88 @@ sub build_chunked {
...
-                "local.book.section.title" => "Learn/Docs/$section",
-                "local.book.subject"       => $subject,
-                "local.noindex"            => $noindex
-            ),
-            $index
-        );
-        1;
-    } or do { $output = $@; $died = 1; };
+    if ( $asciidoctor ) {
+        # TODO icons
```
nik9000: You'll see quite a few of these review. They are all for followups.

lib/ES/Util.pm
```diff
@@ -326,13 +438,13 @@ sub run (@) {
         ( $out, $err, $ok ) = capture { system(@args) == 0 };
     }
 
-    return $out if $ok;
+    return "$out\n$err" if $ok;
```
nik9000: Without this we don't get asciidoctor's warnings because they are on stderr.

resources/asciidoctor/bin/asciidoctor
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env ruby
```
nik9000: Everything below this line is asciidoctor being copied into the repo using bundle. Asciidoctor and Bundler are MIT licensed.

lib/ES/Util.pm
```diff
@@ -180,6 +267,19 @@ sub _check_build_error {
     warn join "\n", ( '', @warn, '' );
 }
 
+#===================================
+sub _xml_lint {
+#===================================
+    my ( $dest ) = @_;
+    return run(
+            'xmllint',
```
ddillinger: Calling `run` (or any subprocess work, really) probably always warrants error handling and reporting and probably even vigorous dying, in the event the world happened to be on fire that day.

resources/asciidoctor/README
```diff
@@ -0,0 +1,2 @@
+This contains a fully checked out copy of asciidoctor. You should be able to
+update it by chaning the `Gemfile` and running `bundle install`.
```
ddillinger: `s/chaning/changing/`

resources/asciidoctor/bin/asciidoctor
ddillinger: ...I'm experiencing so many emotions all at once, haha

lib/ES/Util.pm
```diff
@@ -45,42 +45,88 @@ sub build_chunked {
...
+                '--stringparam', 'base.dir', $chunks_path->absolute . '/',
+                file('resources/website_chunked.xsl')->absolute,
+                "$dest/index.xml"
+            );
+            # TODO copy_resources?
+            # TODO clean the xml files
+            1;
+        } or do { $output = $@; $died = 1; };
+    }
+    else {
```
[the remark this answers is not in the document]
nik9000: I was just doing what the rest of the script does. It looks weird to me too!

lib/ES/Util.pm
```diff
@@ -45,42 +45,88 @@ sub build_chunked {
...
-        );
-        1;
-    } or do { $output = $@; $died = 1; };
+    if ( $asciidoctor ) {
+        # TODO icons
+        my $chunks_path = dir("$dest/.chunked");
+        $chunks_path->mkpath;
+        # Emulate asciidoc_dir because we use it to find shared asciidoc files
+        # but asciidoctor doesn't support it.
+        my $asciidoc_dir = dir('resources/asciidoc-8.6.8/')->absolute;
```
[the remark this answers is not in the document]
nik9000: I'm not particularly concerned that we'll upgrade and lose it. To be honest I don't think we'll upgrade asciidoc at this point. I can't comment on whether or not the path should have a version number in it because it had one when I got here.

nik9000:
@ddillinger, I've pushed some fixes.

I don't think the whole process will get an order of magnitude faster, just about 60% of it. So the best we'll get is about 50% speedup. Which ain't bad!

nik9000:

PR #507, opened 2018-12-14
Beats: add x-pack/filebeat/docs to contents

adriansr:
This PR adds the `x-pack/filebeat/docs` to the list of directories to look for docs in Filebeat. This is needed so that we can include x-pack specific docs.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -613,6 +613,7 @@ contents:
               -
                 repo:   beats
                 path:   x-pack/filebeat/docs
+                exclude_branches:   [ 6.x, 6.5, 6.4, 6.3, 6.2, 6.1, 6.0, 5.6, 5.5, 5.4, 5.3, 5.2, 5.1, 5.0, 1.3, 1.2, 1.1, 1.0.1 ]
               -
                 repo:   beats
                 path:   CHANGELOG.asciidoc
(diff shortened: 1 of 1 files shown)


lcawl:
In case it makes a difference, we're slowly getting rid of that folder structure in the elasticsearch repo and have already removed it from kibana.  The "x-pack" content is now just mixed in with the rest.

lcawl:
My advice would be to leave the x-pack content in the regular docs folder, but if you do go this route of separation I think you'll need to make some additional changes:

1. Add a --resource parameter to your build command for master (e.g. similar to the docbldesx alias in https://github.com/elastic/docs/blob/master/doc_build_aliases.sh)
2. Add a separate build alias for 6.x and earlier releases, since they won't use that resource
3. If you don't already have it, you might also need to add an attribute that clarifies the path to that the two different directories, like the es-repo-dir and xes-repo-dir attributes here: https://raw.githubusercontent.com/elastic/elasticsearch/master/docs/reference/index.asciidoc

dedemorton:
@lcawl Which of the changes listed above need to be made before this PR is merged? The Beats docs use relative paths right now in the include statements (no attributes). We should probably improve that, but it hasn't caused build issues so far.

lcawl:
@dedemorton If I remember correctly, the first errors you'll see will likely be related to files that have images in them.

PR #509, opened 2018-12-14
Support building docs into a separate repo

nik9000:
Adds the `--target-repo` option to the `--all` mode of `build_docs.pl`
to build the docs into a separate repository. For example:
```
./build_docs.pl --all --target_repo=https://github.com/nik9000/built-docs --skiplinkcheck --push
```

That example will clone the nik9000's `built-docs` repository and build
the documentation into that repository instead of the current
repository.

I'd like this to be required at some point but for now it probably
should stay optional while we work on getting it published.

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -204,7 +204,7 @@ sub build_all {
     }
     push_changes($build_dir, $target_repo, $target_repo_checkout) if $Opts->{push};
 
-    # $temp_dir->rmtree;
+    $temp_dir->rmtree;
 }
 
 #===================================
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -49,7 +49,7 @@ use ES::Template();
 
 GetOptions(
     $Opts,    #
-    'all', 'push', 'update!',    #
+    'all', 'push', 'update!', 'target_repo=s',   #
     'single',  'pdf',     'doc=s',           'out=s',  'toc', 'chunk=i',
     'open',    'skiplinkcheck', 'linkcheckonly', 'staging', 'procs=i',         'user=s', 'lang=s',
     'lenient', 'verbose', 'reload_template', 'resource=s@', 'asciidoctor'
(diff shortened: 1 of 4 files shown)


build_docs.pl
```diff
@@ -164,18 +164,17 @@ sub build_all {
     ensure_creds_cache_enabled() || enable_creds_cache() || exit(1);
     check_github_authed();
 
-    my $repos_dir = init_repos();
+    my ($repos_dir, $temp_dir, $target_repo, $target_repo_checkout) = init_repos();
```
nik9000: I move the `$temp_dir` clean and recreation into `init_repos` because I need to clone the target repo into it.

build_docs.pl
```diff
@@ -164,18 +164,17 @@ sub build_all {
...
-
-    $build_dir = dir($build_dir);
+    if ( $target_repo ) {
+        $build_dir = dir("$target_repo_checkout/$build_dir");
+    } else {
+        $build_dir = dir($build_dir);
+    }
     $build_dir->mkpath;
 
-    my $temp_dir = $repos_dir->subdir('.temp');
```
nik9000: Moved into `init_repos`.

build_docs.pl
```diff
@@ -426,17 +457,22 @@ sub init_repos {
...
+    return ($repos_dir, $temp_dir, $target_repo, $target_repo_checkout);
 }
 
 #===================================
 sub push_changes {
 #===================================
-    my $build_dir = shift;
+    my ($build_dir, $target_repo, $target_repo_checkout) = @_;
 
+    local $ENV{GIT_WORK_TREE} = $target_repo_checkout if $target_repo;
```
nik9000: I'm not sure of a more clean way to manipulate the environment than this. It *works* but it doesn't look nice.

build_docs.pl
```diff
@@ -426,17 +457,22 @@ sub init_repos {
...
+    local $ENV{GIT_DIR} = $ENV{GIT_WORK_TREE} . '/.git' if $target_repo;
+
+    say 'Building revision.txt';
```
nik9000: I saw these steps take a while when I tested locally so I added this so I could watch.

lib/ES/BranchTracker.pm
```diff
@@ -77,6 +77,7 @@ sub write {
     my $self = shift;
     my $new  = Dump( $self->shas );
     return if $new eq $self->{yaml};
+    $self->file->parent->mkpath;
```
nik9000: This is important if you are doing this against an empty repository.

lib/ES/Repo.pm
```diff
@@ -251,7 +255,7 @@ sub all_repo_branches {
         my $shas = $repo->tracker->shas_for_repo( $repo->name );
 
         for my $branch ( sort keys %$shas ) {
-            my $log = run( qw(git log --oneline), $shas->{$branch} );
+            my $log = run( qw(git log --oneline -1), $shas->{$branch} );
```
nik9000: I noticed that this step was super slow and this seems to have sped it up a ton.

lib/ES/Util.pm
```diff
@@ -450,7 +450,6 @@ sub run (@) {
     my @args = @_;
     my ( $out, $err, $ok );
 
-    print "Run PATH=$ENV{PATH}\n";
```
nik9000: These make seeing what is going on very hard and we log this at startup anyway.

PR #520, opened 2018-12-21
[DOCS] Adds xpack-basic and xpack-pro roles

lcawl:
Related: https://github.com/elastic/docs/pull/177/files

This PR adds "xpack-basic" and "xpack-pro" roles, so that we can tag pages differently depending on the high-level subscription types (https://www.elastic.co/subscriptions).

### DIFF
--- a/resources/web/styles.css
+++ b/resources/web/styles.css
@@ -44,6 +44,38 @@
   font-weight: inherit;
 }
 
+/* Format floating role=xpack-basic titles */
+#guide h1 > span[class~="xpack-basic"],
+#guide h2 > span[class~="xpack-basic"],
+#guide h3 > span[class~="xpack-basic"],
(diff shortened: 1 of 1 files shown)


lcawl:
@kellyemurphy If you have suggestions for different role names, let me know.

kellyemurphy:
I thought we agreed on `premium`? Or would it be possible to have our own Cloud tags? Since the subscription pages are separate, maybe that would be for the best?

lcawl:
I just used "pro" in the keyword since I wanted something short.  If you think it is confusing, this PR can wait until we've agreed on a term and use that.

lcawl:
Related to https://github.com/elastic/docs/issues/542

lcawl:
These differentiations are no longer required.

PR #521, opened 2018-12-21
Begin customizing asciidoctor

nik9000:
This is the start of a likely fairly extensive effort to customize
asciidoctor so that it can consume Elastic's asciidoc files and produce
usable output. With this change asciidoctor's output is *much* closer
to asciidoc and there are far fewer errors that require `--lenient`.

At this point building the elasticsearch java client docs with asciidoctor
takes:

```
real   0m5.306s
user   0m5.067s
sys    0m0.211s
```

as compared to asciidoc:

```
real   0m51.581s
user   0m44.597s
sys    0m6.386s
```

While these numbers aren't *fair*, they are *compelling*.

In particular this adds support for the following things:

== added[some.version]

Asciidoc supports macros of any shape, but asciidoctor requires that
block macros look like `added::[some.version]`. We fix this with a
regex in a "preprocessor" and we implement the macro as a "block macro".

== include-tagged::some/path[some-tag]

Asciidoc has support for the very similar
`include:some/path[tag=some-tag]` but the tags in the source file aren't
quite the same and asciidoctor doesn't strip indentation by default while
we do. This is implemented as an "include processor" but those don't have
quite the same syntax. We catch the old syntax in a "preprocessor" and
call our "include processor" from there.

== Elastic's code listing section 

Elastic often uses syntax like:

```
["source","java",subs="attributes,callouts,macros"]
include-tagged::{doc-tests-file}[{api}-hits-singleHit-source]
```

This is required because in asciidoc `include-tagged` is a macro.
Specifying the list of subs breaks code that contains special characters
like `<`. We fix that by forcibly adding `specialcharacters` escaping
to all listing (`----`) blocks regardless of if that is included in the
`subs`. That is more consistent with asciidoc's behavior.

The callouts (`<1>`) are also broken by this because asciidoctor
recognizes callouts with a regex that recognizes that post-escaped
sequence of characters. That means that `specialcharacters` has to come
*before* `callouts` in the list of subs. This seems like a bug to me so I
filed it:
https://github.com/asciidoctor/asciidoctor/issues/2988

We fix both of these with a "tree processor".


lib/ES/Util.pm
```diff
@@ -77,7 +77,8 @@ sub build_chunked {
         my $asciidoc_dir = dir('resources/asciidoc-8.6.8/')->absolute;
         eval {
             $output = run(
-                'asciidoctor', '-v',
+                'asciidoctor', '-v', '--trace',
```
nik9000: `--trace` gives us a traceback when there is an error which is important if we're going to debug these things....

resources/asciidoctor/.bundle/config
```diff
@@ -1,3 +1,4 @@
 ---
-BUNDLE_PATH: "vendor/bundle"
-BUNDLE_BIN: "bin"
+BUNDLE_PATH: "build/bundle"
```
nik9000: This became the "development" bundle config.

resources/asciidoctor/.bundle/prod/config
```diff
@@ -0,0 +1,4 @@
+---
+BUNDLE_PATH: "vendor/bundle"
```
nik9000: And this is the "production" one.

resources/asciidoctor/spec/spec_helper.rb
```diff
@@ -0,0 +1,49 @@
...
+      :doctype => :book,
+      :attributes => {
+        'docdir' => File.dirname(__FILE__),
+      },
+    }
+  if logger.messages != [] then
+    raise ConvertError.new(logger.messages, result)
+  end
+  result
+end
```
nik9000: I'm not 100% sure where the binary file below came from. I got them after running bundler again so I suspect they are indeed part of asciidoctor but I didn't pull them in somehow the first time.

PR #522, opened 2018-12-27
Add two arguments for ES's Jenkins

nik9000:
`--reference` adds support for using a local mirror which is how
Elastic's CI infrastructure usually speeds up big clones.

`--rely_on_ssh_auth` relies on the SSH authentication already configured
to perform all tasks which is conventient because Elastic's CI already
*has* SSH authentication configured and so do many users. Mechanically
causes the script to skip the git authentication tests entirely,
presuming that SSH authentication is already configured. It also causes
the script to transform all configured repositories that happen to use
`https` into `ssh`.

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -392,8 +392,11 @@ sub init_repos {
     my $tracker_path = $Conf->{paths}{branch_tracker}
         or die "Missing <paths.branch_tracker> in config";
 
-    my $reference_dir = dir($Opts->{reference})->absolute;
-    die "Missing reference directory $reference_dir" unless -e $reference_dir;
+    my $reference_dir = dir($Opts->{reference});
+    if ( $reference_dir ) {
+        $reference_dir = $reference_dir->absolute;
+        die "Missing reference directory $reference_dir" unless -e $reference_dir;
+    }
 
     my $target_repo = 0;
     my $target_repo_checkout = 0;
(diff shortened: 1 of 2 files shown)


lib/ES/Repo.pm
```diff
@@ -60,7 +70,7 @@ sub update_from_remote {
         unless ( $self->_try_to_fetch ) {
```
nik9000: Right now if we don't check that the reference is the same as the last execution in `_try_to_fetch`. We probably should update it if it has changed. That isn't super likely, but it is probably worth doing.

build_docs.pl
```diff
@@ -390,24 +392,28 @@ sub init_repos {
     my $tracker_path = $Conf->{paths}{branch_tracker}
         or die "Missing <paths.branch_tracker> in config";
 
+    my $reference_dir = dir($Opts->{reference})->absolute;
+    die "Missing reference directory $reference_dir" unless -e $reference_dir;
```
ddillinger: Two things:

1. Maybe I am missing some context (or my perl is just too rusty) but this seems like it's a pretty non-optional option if it's going to `die` on this?
2. Looking at the diff I am guessing it was already like that but what on earth happened to this sub's indentation? It is wrong in the PR diff view, but _looks_ right in raw view, which makes me suspect tabs/spaces mixing is going on maybe?

nik9000:
> By which mainly I mean "does it produce an error that is helpful enough to a human who wants to fix it"?

It'll die with the output of the failed git command. That seems ok to me.

build_docs.pl
nik9000: RE indentation: Bleh! It looks right locally which is super odd.

RE non-optional: I'll fix it! That'd have broken everything..... I need to test this some more too.

nik9000:

PR #525, opened 2019-01-03
Asciidoctor: Work around strange attribute rules

nik9000:
Asciidoc requires that we fence attributes at the beginning of a section
like so:
```
--
:api: index
:request: IndexRequest
:response: IndexResponse
--
```

But asciidoctor drops the attribute values if they are in the block
which I've filed here:
https://github.com/asciidoctor/asciidoctor/issues/2993

That means that we have to *remove* the blocks for asciidoctor. This
does that.

### DIFF
--- a/resources/asciidoctor/spec/spec_helper.rb
+++ b/resources/asciidoctor/spec/spec_helper.rb
@@ -41,7 +41,6 @@ def convert input
       :backend    => :docbook45,
       :logger     => logger,
       :doctype    => :book,
-      :verbose    => true,
       :attributes => {
         'docdir' => File.dirname(__FILE__),
       },
--- a/resources/asciidoctor/spec/spec_helper.rb
+++ b/resources/asciidoctor/spec/spec_helper.rb
@@ -28,16 +28,20 @@ class ConvertError < Exception
   end
 end
 
+# Put asciidoctor into verbose mode so it'll log reference errors
+$VERBOSE = true
+
 ##
 # Convert an asciidoc string into docbook. If the conversion results in any
 # errors or warnings then raises a ConvertError.
 def convert input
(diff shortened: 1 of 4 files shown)


resources/asciidoctor/spec/spec_helper.rb
```diff
@@ -28,16 +28,20 @@ def initialize warnings, result
...
   result = Asciidoctor.convert input, {
-      :safe => :unsafe,  # Used to include "funny" files.
-      :backend => :docbook45,
-      :logger => logger,
-      :doctype => :book,
+      :safe       => :unsafe,  # Used to include "funny" files.
+      :backend    => :docbook45,
+      :logger     => logger,
+      :doctype    => :book,
+      :verbose    => true,
```
ddillinger: Should this use the `$VERBOSE` variable?

nik9000: Oh! Actually it doesn't do anything. Let me drop it.

nik9000: I tried this first because it feels like it should work, but it doesn't. Because :shrug:

nik9000:
> Do we have, or need to support, mixed blocks? I see tests for attribute-only, and non-attribute-having, but nothing for has-an-attribute-and-also-something-else (but I don't know if that's a thing that can happen...)

You can totally have mixed blocks and we should treat them like the non-attribute-having blocks because we want to run normal processing on them. Honestly I'd prefer not to do anything with blocks but this seems to be what asciidoctor needs right now but it is weird and I don't like it so I want to keep it to a minimum. I'll add a mixed test just in case though.

nik9000:
>  I'll add a mixed test just in case though.

Done!

nik9000:

PR #526, opened 2019-01-04
Asciidoctor: Add "Edit Me" Links

nik9000:
This adds the missing "Edit Me" links to every section. It uses a tree
processor that modifies how some parts of Asciidoctor's AST render their
titles to render the link.

### DIFF
--- a/lib/ES/Util.pm
+++ b/lib/ES/Util.pm
@@ -85,9 +85,8 @@ sub build_chunked {
                 '-d' => 'book',
                 '-a' => 'showcomments=1',
                 '-a' => "lang=$lang",
-                '-a' => 'base_edit_url=' . $edit_url,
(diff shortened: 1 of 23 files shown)


lib/ES/Util.pm
```diff
@@ -85,9 +85,8 @@ sub build_chunked {
                 '-d' => 'book',
                 '-a' => 'showcomments=1',
                 '-a' => "lang=$lang",
-                '-a' => 'base_edit_url=' . $edit_url,
                 '-a' => 'root_dir=' . $root_dir,
-                $private ? ( '-a' => 'edit_url!' ) : (),
+                $private ? () : ( '-a' => "edit_url=$edit_url" ),
```
nik9000: I've flipped this around because just the presence of the edit_url attribute is enough for asciidoctor.

PR #527, opened 2019-01-07
Asciidoctor: Fix raw xslt opts

nik9000:
My previous work on raw xslt opts added extra `'`s around every
variable. This broke all kinds of stuff like chunking and headers and,
well, almost everything! This fixes that and makes that `--verbose`
printing of commands a little more clear.


nik9000:
Escaping is seriously the hardest problem in computer science.

nik9000:

PR #529, opened 2019-01-08
Asciidoctor: Work around "cramped" include statements

nik9000:
We do a fair bit of:

```
include::resources/1.adoc[]
include::resources/2.adoc[]
```

In our asciidoc files. Asciidoctor would prefer that we do:

```
include::resources/1.adoc[]

include::resources/2.adoc[]
```

but asciidoc doesn't care. To be compatible with asciidoc this works
around asciidoctor's preference, preventing the parsing error that come
about because of using these "cramped" include statements by adding a
trailing new line every time we include an asciidoc file. This is a
little crude but it works like a charm.

With this, I can now build the Elasticsearch Java REST Client reference
without `--lenient` and the document looks great! This is how long the
build takes with asciidoctor:

```
real  0m7.652s
user  0m7.080s
sys   0m0.538s
```

And this is how long it takes with asciidoc:

```
real  0m44.987s
user  0m38.715s
sys   0m6.256s
```


nik9000:
@ddillinger I've fixed the line numbers.

resources/asciidoctor/spec/creamped_include_spec.rb
```diff
@@ -0,0 +1,73 @@
+require 'added/extension'
```
ddillinger: This filename is typo'd

nik9000: Yeah. We don't need it here. I'll drop it.

nik9000:
@ddillinger I've pushed a fix for your requested change. Thanks for finding it!

resources/asciidoctor/spec/creamped_include_spec.rb
ddillinger: No I mean literally the filename. You meant `cramped` but it is `creamped` :)

nik9000: Ah!

nik9000: Fixed!

nik9000:
Thanks for keeping me honest @ddillinger!

PR #530, opened 2019-01-08
Asciidoctor: Force callouts into source listing

nik9000:
Asciidoc automatically adds callouts to all source listing regardless of
whether you ask for them or not and we rely on this behavior in the
Elasticsearch docs. This emulates it in asciidoctor. It isn't super
clean because we have to edit the text on the way in, but that is life
sometimes.


nik9000:

PR #531, opened 2019-01-08
Asciidoctor: Handle ` as monospaced literal

nik9000:
In Asciidoctor 1.5.0 they switched ` to turn on monospaced formatting
instead of delimiting a monospaced literal. Our docs assume that `
delimits a monospaced literal. Asciidoctor has an attribute that allows
us to go back to the old behavior and this does that.


nik9000:

PR #533, opened 2019-01-12
Add infra and logs UI to overview topic

dedemorton:
The overview topic was missing links for the infra and logs UIs added in 6.5. 

@lcawl FYI. For backporting, I guess we need a new version of the settings file, no? Let me know if there is anything I'm overlooking.


lcawl:
I've actually been pondering how/if we can get rid of this list, since it's very X-Pack-specific. Until then, however, you're right, you'll need to split off another version of the file to be used for 6.2, 6.3, and 6.4.

PR #534, opened 2019-01-14
Change Cloud current

kellyemurphy:
Changing ESS (Cloud) build to MS15


nrichers:
As per Igor, MS 15 code has been deployed.

PR #535, opened 2019-01-14
Asciidoctor: Fix mismatched code block fencing

nik9000:
Asciidoc will match any line that starts with `----` and only contains
`-` as a code block fence. Asciidoctor will do that for the *start* of
the code block fence, but the *end* of the code block fence must match
the start fence perfectly. Asciidoc has no such check. It is generally
good practice for them to match so they usually do, but they don't
always. So we normalize them on load, logging a warning if we have to do
it so the author of the docs can fix them at their leisure.

This comes up enough in the Elasticsearch reference that the reference
fails to render about half with through without this fix. With this fix
it still fails to render, but for some other, hopefully unreleated
reason.


nik9000:

PR #538, opened 2019-01-15
Adds settings file for 6.2

dedemorton:
Since 6.2 currently points to settings61.asciidoc, contents here are a simple copy of the 6.1 file.


dedemorton:
Closing because I don't think I've done this correctly.

PR #539, opened 2019-01-15
Add option to force rebuilding

nik9000:
Adds a `--rebuild` option to force rebuilding all branches of all books
regardless of whether or not the source of the books has changed. This
is useful when we need to rebuild the version drop downs after a
release.

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -49,7 +49,7 @@ use ES::Template();
 
 GetOptions(
     $Opts,    #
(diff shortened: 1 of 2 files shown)


nik9000:
@ddillinger I figure I can expose this via Jenkins and then anyone can select that option after a release to rebuild the docs. That way we don't rely on folks doing this on their laptops.

nik9000:

PR #543, opened 2019-01-17
Fix the edit url with --rely_on_ssh_auth

nik9000:
The `--rely_on_ssh_auth` option designed for use with new-style CI
breaks that the "Edit Me" links by using ssh-style repos for the basis
of the links instead of https. These links obviously don't work in the
browser. This converts this ssh style links into https style links when
building the edit url.


nik9000:

PR #544, opened 2019-01-17
Add option to skip fetching source repos

nik9000:
Sometimes it is difficult to experiment with changes to the docs build
because we always fetch changes from the source repositories and those
changes can get mixed up in the changes from the experiment. This adds a
`--no_fetch` option to skip fetching changes from the source
repositories. We still fetch changes from the destination repo if there
is one, but the docs that we build are then based on whatever changes
we've pulled locally.

I suspect this'll never get used in CI, but I'll use it fairly
frequently when experimenting locally.

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -49,7 +49,7 @@ use ES::Template();
 
 GetOptions(
     $Opts,    #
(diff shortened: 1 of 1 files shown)


nik9000:

PR #545, opened 2019-01-17
Fix edit links for local books

nik9000:
The "Edit" links on each page have historically been broken when
building books with `./build_docs.pl --doc` because we don't have the
right urls. This is mostly fine, but I'd like to do some debugging of
how asciidoctor generates those links so it'd be nice if they worked
with `--doc`. This takes a reasonable stab at making them work by
looking for the git repository that contains the target of the `--doc`
parameter. It then looks for the clone that comes from the `elastic`
organization. It uses that clone as the basis for the edit url.


nik9000:
Thanks @ddillinger! :heart:

PR #546, opened 2019-01-17
Asciidoctor: Fix edit url for nested roots

nik9000:
Fixes the edit url when the "base" document directory isn't the base of
the repository.


nik9000:

PR #547, opened 2019-01-17
Asciidoctor: Drop xml file after docbook

nik9000:
Drop the giant xml file produced by asciidoctor after docbook is all done.
We don't keep the xml file we make with asciidoc so we shouldn't keep
the xml file from asciidoctor either. It is useful for debugging, but we
can always get it if we need it.


nik9000:

PR #548, opened 2019-01-17
Asciidoctor: Support added inline macro

nik9000:
Adds support for the added inline macro which would normally be
specified like this:

```
words added:[version]
```

But for compatibility with Elastic's asciidoc customizations we support
specifying it like this as well:

```
words added[version]
```

Note that we'll want to build on this to add support for `coming` and
`deprecated` as well.

### DIFF
--- a/resources/asciidoctor/spec/added_spec.rb
+++ b/resources/asciidoctor/spec/added_spec.rb
@@ -1,17 +1,15 @@
 require 'added/extension'
 
-RSpec.describe AddedBlock do
+RSpec.describe Added do
   before(:each) do
-    Extensions.register do
-      block_macro AddedBlock
-    end
+    Extensions.register Added
   end
 
   after(:each) do
     Extensions.unregister_all
   end
 
-  it "creates a note" do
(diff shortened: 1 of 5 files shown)


resources/asciidoctor/spec/added_spec.rb
```diff
@@ -40,4 +38,50 @@
...
+      <simpara>words <phrase revisionflag="added" revision="some_version">
+        more words
+      </phrase>
+      </simpara>
+      </chapter>
+    DOCBOOK
+    expect(actual).to eq(expected.strip)
+  end
+
+  it "inline version is not invoked without the ::" do
```
ddillinger: Given the PR description... does this match the behavior of `asciidoc` too?

nik9000: So, in asciidoc the inline macro *is* invoked without the `:`, but in asciidoctor it isn't. We "fix" the input to add the `:` in the compat preprocessor. This is a "doesn't it work how I think it should" test more than anything.

nik9000:

PR #564, opened 2019-01-29
Allow running the build with docker

nik9000:
This adds support for building the docs using docker with the expectation
without breaking support for building the docs *without* docker. My goal
is to make it simpler for to add dependencies. In the long term this'll
allow us to add support for automatically building diagrams and fancy
mathematical notation. In the short term this'll make it simpler to run
asciidoctor in CI.

On macOS docker is a little slower than running outside of docker. At
this point the overhead is about 20%. This isn't great, but Asciidoctor
will more than make up the difference in the long run. And, in the short
run at least, docker isn't reaquired. I'll make sure it is *never*
required to build Asciidoc.

One fairly questionable choice that this makes is to diable seccomp by
default. The story goes like this: on linux, building the docs in docker
is somewhere between *faster* than building them outside of docker and
macOS's overhead. Unless you turn on seccomp. With seccomp enabled the
docs take about twice as long to build in docker. Which is terrible! As
much as I want security, this is terrible! I'd really love to turn it on
but I can't defend that kind of hit.


Dockerfile
```diff
@@ -0,0 +1,35 @@
+# Debian builds the docs about 20% faster than alpine. The image is larger
+# and takes longer to build but that is worth it.
+FROM bitnami/minideb:stretch
```
nik9000: Alpine is much nicer to work with but 20% slower is a big deal given how often we run this locally.

README.asciidoc
```diff
@@ -17,23 +17,7 @@ on any other website without explicit permission.
 [float]
 == Requirements
 
-First, you need to make sure that you have the following installed:
```
nik9000: I'm just pointing folks right at docker by default because it *ought* to be simpler to set up.

README.asciidoc
```diff
@@ -1027,13 +1005,13 @@ footnote to a particular line of code:
 ==================================
 
 [[sense-snippets]]
-=== View in Sense
+=== View in Console
```
nik9000: These aren't related but they jumped out at me while I was here. I can break this out if folks are more comfortable with that.

build_docs.pl
```diff
@@ -127,18 +127,69 @@ sub build_local {
...
         else {
-            my $http = dir( 'resources', 'http.py' )->absolute;
-            close STDIN;
-            open( STDIN, "</dev/null" );
-            chdir $dir;
-            exec( $http '8000' );
+            if ( $running_in_docker ) {
+                # We use nginx to serve files instead of the python built in web server
+                # when we're running inside docker because the python web server performs
+                # badly there. nginx is fine.
```
nik9000: The python web server is fine some of the time but hangs for many seconds at a time. I *suspect* it'll do that on some Linux systems as well, but nginx runs just fine here.

build_docs.pl
```diff
@@ -127,18 +127,69 @@ sub build_local {
...
+}
+
+http {
+  log_format short '[\$time_local] "\$request" \$status';
+  access_log /dev/stdout short;
+  server {
+    listen 8000;
+    location / {
+      root $dir;
+      add_header 'Access-Control-Allow-Origin' '*';
```
nik9000: These are copied from the python runner.

build_docs.pl
```diff
@@ -127,18 +127,69 @@ sub build_local {
...
 }
 
+#===================================
+sub _running_in_docker {
+#===================================
+    my $root_cgroup = dir('/proc/1/cgroup');
+    return 0 unless ( -e $root_cgroup );
+    open(my $root_cgroup_file, $root_cgroup);
+    return grep {/docker/} <$root_cgroup_file>;
+}
```
nik9000: This is one of those methods that makes me actually *like* perl. Lots of magic syntax which, maybe isn't great, but stuff like `return 0 unless CONDITION` is super nice.

build_docs.sh
```diff
@@ -0,0 +1,173 @@
...
+  # Emulate a terminal so things like ctrl-c work.
+  DOCKER_RUN_ARGS+=('-it')
+fi
+
+echo "Building the docker image that will build the docs. Expect this to" \
+    "take somewhere between a hundred milliseconds and five minutes."
+# Build the docker image from stdin so we don't try to pack up everything in
+# this directory. It is huge and we don't need any of it in the image because
+# we'll mount it into the image on startup.
+docker image build -t elastic/docs_build - < "$DIR/Dockerfile"
```
nik9000: Building the docker image right here feels good to me from a development background. It feels "like gradle but for native dependencies". I know the analogy isn't quite right, but it is pretty close in this case.

build_docs.sh
```diff
@@ -0,0 +1,173 @@
...
+# performance boost and it is simple enough to do.
+DOCKER_RUN_ARGS+=('--read-only')
+DOCKER_RUN_ARGS+=('--tmpfs' '/tmp')
+
+# Mount the docs build code so we can run it!
+DOCKER_RUN_ARGS+=('-v' "$DIR:/docs_build:cached")
+
+# Seccomp adds a *devestating* performance overhead if you happen
+# to have it installed.
+DOCKER_RUN_ARGS+=('--security-opt' 'seccomp=unconfined')
```
nik9000: See the commit message for this one. I don't like it, but without it my laptop *hates* this change. I've spent a few days looking into it and I'm cutting my losses for now.

build_docs.sh
```diff
@@ -0,0 +1,173 @@
...
+DOCKER_RUN_ARGS+=('--rm')
+
+# Make sure we create files as the current user because that is what
+# folks that use build_docs.pl expect.
+DOCKER_RUN_ARGS+=('--user' "$(id -u):$(id -g)")
+
+# Running read-only with a proper tmp directory gives us a little
+# performance boost and it is simple enough to do.
+DOCKER_RUN_ARGS+=('--read-only')
+DOCKER_RUN_ARGS+=('--tmpfs' '/tmp')
```
nik9000: It turns out it isn't *that* simple to do for nginx, but it isn't too bad and it feels good to know all the places where we write.

build_docs.sh
```diff
@@ -0,0 +1,173 @@
+#!/bin/bash
```
nik9000: longform-original-14836-1414320930-10

Any advice for writing a bash script that isn't terrible. I'm quite attached to the "iterate all the args and rewrite" method, but beyond that I have no opinions.

build_docs.sh
```diff
@@ -0,0 +1,173 @@
...
+        echo "------------------------ WARNING ------------------------"
+        echo "Attempting to share ssh auth but this is unlikely to work" \
+             "outside of linux."
+        echo "------------------------ WARNING ------------------------"
+      fi
+      DOCKER_RUN_ARGS+=('-v' "$(dirname $SSH_AUTH_SOCK):$(dirname $SSH_AUTH_SOCK):ro")
+      DOCKER_RUN_ARGS+=('-e' "SSH_AUTH_SOCK=$SSH_AUTH_SOCK")
+    fi
+    # Mount our known_hosts file into the VM so it won't ask about github
+    DOCKER_RUN_ARGS+=('-v' "$(to_absolute_path ~/.ssh/known_hosts):/tmp/.ssh/known_hosts:ro,cached")
```
nik9000: I know we could use something like ` ssh-keyscan -t rsa github.com >> ~/.ssh/known_hosts` to add github's key but this feels, like, slightly better. I feel like this trusts that the user cloned from github and accepted the host. I could copy it in I suppose rather than mount. I've kind of gone mount crazy here but lots of mounts is kind of required for a project like this.

nik9000:
One thing I didn't do in *this* PR is switch to using the asciidoctor gem in the docker image. We're just using `ruby` from the image. I'd like to save that for a followup PR because it is fairly mechanical and not actually interesting from a "docker" perspective.

build_docs.sh
nik9000: I like that we could do things like `git bisect` and get the docker image specified by the Dockerfile in the repo right now. I'd be quite happy with a caching strategy that makes it so we don't have to build the image locally all the time, but this feels like a simple solution for now.

build_docs.sh
```diff
@@ -0,0 +1,173 @@
...
+      echo "$1"
+      return
+    fi
+    cd ..
+  done
+  echo "$PWD"
+}
+
+DIR="$(dirname "$(to_absolute_path "$0")")"
+
```
jarpy: You're going to hate me, but:

```
git rev-parse --show-toplevel
```

> _Show the absolute path of the top-level directory._

build_docs.sh
```diff
@@ -0,0 +1,173 @@
...
+# a list of arguments to be passed to start the docker container and a
+# list of arguments to be passed to the build_docs.pl process that is
+# started in the docker container.
+# Step 3 is to start the docker container. We start it in such a way
+# that is *should* remove itself when it is done.
+
+set -e
+
+function desymlink() {
+  FILE="$1"
```
jarpy: Is this an implementation of `realpath(1)`? OSX doesn't have that, huh? Wowsers.

It does have Python with `os.path.realpath()`, though (see above). Python (or other modern scripting language) would eliminate a bunch of these disparities that work differently between GNU and BSD.

I assume you can't dereference symlinks like this on OSX either:

```
$ stat -c%N /tmp/link | cut -d "'" -f4
/tmp/i/am/a/symlinked/file.txt
```

I feel like Bash on BSD is really punishing you here if it's making you write `realpath(1)`! Philosophically, there's lot of system-dependant code to get us into Docker so we can stop thinking about the system.

jarpy: But you can, according to the manpage, do `stat -c%Y /the/link`, which is _really_ nice but not in GNU. Arrggggg!

build_docs.sh
nik9000: I love you!

build_docs.pl
nik9000: Sooo, I have a degree in math but I'm working for the docs team. Maybe I can be both!

Usually I'm not a fan of funny little language features like this, but having the option to use `unless` instead of `if` and postfix conditionals are super readable to me.

build_docs
```diff
@@ -0,0 +1,177 @@
...
+            stdin = dockerfile,
+            stdout = subprocess.PIPE,
+            stderr = subprocess.STDOUT)
+        docker_logger = logging.getLogger('docker build')
+        start_logging_at = time.time() + DOCKER_BUILD_QUIET_TIME
+        acc = []
+        while True:
+            line = build.stdout.readline()
+            if line != '':
+                line = line.rstrip()
```
jarpy: Any reason not to simply iterate `for line in build.stdout`?

build_docs
```diff
@@ -0,0 +1,177 @@
...
+    docker_args.extend(['--read-only', '--tmpfs', '/tmp'])
+    # Mount the docs build code so we can run it!
+    docker_args.extend(['-v', '%s:/docs_build:cached' % DIR])
+    # Seccomp adds a *devestating* performance overhead if you happen
+    # to have it installed.
+    docker_args.extend(['--security-opt', 'seccomp=unconfined'])
+
+    if stdout.isatty:
+        # Emulate a terminal so things like ctrl-c work.
+        docker_args.append('-it')
```
jarpy: Very thoughtful!

build_docs
```diff
@@ -0,0 +1,177 @@
...
+
+    saw_out = False
+    resource_count = 0
+    a = 0
+    while a < len(args):
+        build_docs_args.append(args[a])
```
jarpy: I guess this technique will break if people use GNU style longopts with `=` in them, but maybe `build_docs.pl` doesn't support them anyway? (And maybe I'm the only freak who uses them).

nik9000:
> I think Python 2 is a much better [second-]lowest common denominator.

It is *so* much more readable. I'll see what I can do about gnu-style arguments. I suspect the perl script supports them and I'd prefer to be 100% transparent.

build_docs
nik9000: I recall that I did this because of something to do with buffering, but I'll recheck.

build_docs
```diff
@@ -0,0 +1,177 @@
...
+    subprocess.check_call(args)
+
+if __name__ == '__main__':
+    from sys import argv
+    try:
+        logging.basicConfig(level = logging.INFO)
+        build_docker_image()
+        run_build_docs(argv[1:])
+    except IOError, e:
+        print(e)
```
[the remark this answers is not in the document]
nik9000: Silly me.

build_docs
nik9000: Yeah, it doesn't work. And the perl script does. Let me see what I can do about that.

nik9000: It won't pick up the single character alias style options either but the perl script doesn't support any of those.

README.asciidoc
```diff
@@ -54,7 +38,7 @@ this README as follows:
 [source,bash]
 ----------------------------
 cd docs/
-./build_docs.pl --doc README.asciidoc --open
+./build_docs.sh --doc README.asciidoc --open
 ----------------------------
```
jarpy: There's no `.sh` any more, right?

README.asciidoc
```diff
@@ -54,7 +38,7 @@ this README as follows:
 [source,bash]
 ----------------------------
 cd docs/
-./build_docs.pl --doc README.asciidoc --open
+./build_docs.sh --doc README.asciidoc --open
```
jarpy: There's no `.sh` any more, right?

README.asciidoc
jarpy: Grep tells me there are two more SENSEs lurking in here if you wanna smack 'em.

```
$ grep -iFn sense README.asciidoc
1066:    // SENSE: path/to/snippet.json
1082:// SENSE:one/two/three/example_1.json <1>
```

nik9000:
The other two senses I'm not sure what they are! I'm worried that if I zap
them it'll be a lie! I have to investigate them further.

nik9000:
Good call!

nik9000:
@jarpy could you have another look when you get a chance?

build_docs
```diff
@@ -0,0 +1,258 @@
+#!/usr/bin/env python2
```
nik9000: oh no! mac doesn't like this shebang!

nik9000:
I talked to @jarpy last night. The shebang issue is that last one that we could find. Sadly, it looks like the best way to solve this is to make the script compatible with any version of python which is a little funky, but I'll work on it today. It is the one thing in this repo that'll have to deal with "funny" versions. Everything else had docker to control it.

build_docs
```diff
@@ -0,0 +1,282 @@
...
+    # Mount the docs build code so we can run it!
+    docker_args.extend(['-v', '%s:/docs_build:cached' % DIR])
+    # Seccomp adds a *devestating* performance overhead if you happen
+    # to have it installed.
+    docker_args.extend(['--security-opt', 'seccomp=unconfined'])
+    # Keep stdin open so the docs build can use closing it as a signal that
+    # it needs to die.
+    docker_args.append('-i')
+
+    # NOCOMMIT handle "tell me who you are"!
```
nik9000: Hmmm - this is a blocker! I didn't realize I hadn't got to it yet.

nik9000:
I just tested the `NOCOMMIT` - it was a leftover. I'd already handled it. I was able to build docs locally with docker and push the changes. This might just work!

nik9000:
@jarpy, could you give this another round today?

build_docs
```diff
@@ -0,0 +1,283 @@
...
+    line = decode(popen.stdout.readline())
+    while line:
+        line = line.rstrip()
+        handle_line(line)
+        line = decode(popen.stdout.readline())
+    popen.wait()
+
+
+def decode(bytes_or_str):
+    """Decode the result of reading from popen's stdout. In python 2 the
```
jarpy: Sorry it came to this :(

nik9000:
> I say ship this bad baby.

:heart:! Thanks for all of your reviewing! It was a long road, but we're getting there!

nik9000:
OK! Things are looking calmer so I've merged this. Now you can use docker! I'll see about fixing ssh auth on mac now.

PR #565, opened 2019-01-30
Asciidoctor: Log absolute path when copying images

nik9000:
Switch the `copy_images` Asciidoctor plugin to log the absolute path
when copying images and clean up some rough spots in the tests around
`copy_images` that I bumped into when making this change.


resources/asciidoctor/spec/spec_helper.rb
```diff
@@ -33,7 +33,7 @@ def convert input, extra_attributes = {}, warnings_matcher = eq('')
     }
   warnings_string = logger.messages
         .map { |l| "#{l[:severity]}: #{l[:message].inspect}" }
-        .join('\n')
+        .join("\n")
```
nik9000: This is a little oopsie I caught!

nik9000:
Thanks for reviewing @ddillinger !!

PR #573, opened 2019-02-04
Asciidoctor: Remaining change admonishments

nik9000:
Adds support for `coming` and `deprecated` which function very similarly
to `added`.


resources/asciidoctor/lib/elastic_compat_preprocessor/extension.rb
```diff
@@ -7,15 +7,23 @@
 #
 # Turns
 #   added[6.0.0-beta1]
+#   changed[6.0.0-beta1]
```
ddillinger: The code and comments disagree, the code says `coming` but the comments say `changed`

resources/asciidoctor/lib/change_admonishment/extension.rb
```diff
@@ -0,0 +1,78 @@
...
+#   Foo added:[6.0.0-beta1]
+#   Foo coming:[6.0.0-beta1]
+#   Foo deprecated:[6.0.0-beta1]
+#
+class ChangeAdmonishment < Extensions::Group
+  def activate registry
+    [
+        [:added, 'added'],
+        [:coming, 'changed'],
+        [:deprecated, 'deleted'],
```
ddillinger: One of these things is not like the others! I'm sure there is a reason BUT, to people (like me!) reading this, the word `deprecated` _absolutely does not_ mean the same thing as `deleted`, in fact it very specifically implies "definitely totally 100% still present".

Is this dictated by the docbook side? I see from the tests that the revisionFlag will contain `deleted`. If that is within our control it may be worth its own issue? I have no clear sense of how much inertia there is behind it either.

nik9000: It is dictated by the docbook side, yeah. I'll file an issue for it! I think it is a thing that we can clean up now, but I'd like to break it into its own thing.

nik9000: https://github.com/elastic/docs/issues/588

resources/asciidoctor/lib/elastic_compat_preprocessor/extension.rb
nik9000: I've pushed a fix!

nik9000:

PR #574, opened 2019-02-04
Asciidoctor: Better handling for resources attribute

nik9000:
Prevents the copy images asciidoctor extension from dying if the
`resources` attribute is an empty string and improves the error message
if it is invalid CSV.


nik9000:

PR #576, opened 2019-02-05
Drop scripts for releasing docs

nik9000:
Now that we've started building the docs into the `built-docs`
repository these scripts aren't going to work properly anyway. We now
have a `rebuild` parameter which we can click in jenkins which will do
the job for us.


nik9000:

PR #578, opened 2019-02-05
Drop self-updating

### LINKED ISSUE #575: Update --staging parameter in build_docs.pl
The --staging parameter in https://github.com/elastic/docs/blob/master/build_docs.pl currently pushes the output from a local documentation build to the staging branch. It should be updated to use the staging branch in https://github.com/elastic/built-docs

nik9000:
Drops the branch switching and self updating behavior from
`build_docs.pl` because we don't need it now that we're building the
docs using our standard build process.

Closes #575

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -49,7 +49,7 @@ use ES::Template();
 
 GetOptions(
     $Opts,    #
-    'all', 'push', 'update!', 'target_repo=s', 'reference=s', 'rely_on_ssh_auth', 'rebuild', 'no_fetch', #
+    'all', 'push', 'target_repo=s', 'reference=s', 'rely_on_ssh_auth', 'rebuild', 'no_fetch', #
     'single',  'pdf',     'doc=s',           'out=s',  'toc', 'chunk=i',
     'open',    'skiplinkcheck', 'linkcheckonly', 'staging', 'procs=i',         'user=s', 'lang=s',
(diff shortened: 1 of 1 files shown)


ddillinger:
#575 has a [comment on it](https://github.com/elastic/docs/issues/575#issuecomment-460791504) that makes me wonder if this PR is still entirely valid?

nik9000:
> #575 has a [comment on it](https://github.com/elastic/docs/issues/575#issuecomment-460791504) that makes me wonder if this PR is still entirely valid?

I think this PR is still valid. This drops the scary "self-updating" behavior of `--staging` and `--update` but keeps the change in template pulling I mention in the comment.

build_docs.pl
```diff
@@ -539,14 +537,8 @@ sub push_changes {
             run qw(git push origin HEAD );
         }
         local $ENV{GIT_DIR} = $target_repo->git_dir if $target_repo;
-        if ( $Opts->{staging} ) {
-            say "Force pushing changes to staging";
-            run qw(git push -f origin HEAD );
```
ddillinger: This `-` diff right here is the code equivalent of warding off the Evil Eye hahaha

nik9000: One of the PRs you approved two days ago also had push -f!

ddillinger: 😐

PR #584, opened 2019-02-06
Default to --rely_on_ssh_auth

nik9000:
Make `--rely_on_ssh_auth` the default behavior, turning
`--rely_on_ssh_auth` into a noop. We can drop suppor for it entirely in
a follow up change once we've dropped the argument from Jenkins.


nik9000:
Hmm - I don't think I can merge this now actually. It conflicts with the docker work in that sharing the ssh config into the docker container won't work on mac. I'd like to get docker in more than I'd like to get this in. Then I can work on fixing the ssh sharing on mac.

nik9000:
Hmmm - docker isn't going to with with interactively reading the passwords anyway.....

nik9000:
@ddillinger or @jarpy I think this is ready to look at now.

PR #587, opened 2019-02-07
Remove docker detection

nik9000:
When I first allowed running the docs build in docker I made the
incorrect assumption that any time the build is running in docker it is
running in the docker container that we manage. Lots of folks were using
docker to manage the pl dependencies of the build. This drops that
assumption and adds an explicit parameter to engage `build_docs.pl`'s
customizations that work for our docker image. The folks who were building
the docs in their own docker image have their own customizations and
won't need ours.

### DIFF
--- a/build_docs
+++ b/build_docs
@@ -94,6 +94,8 @@ def run_build_docs(args):
     # it needs to die.
     docker_args.append('-i')
 
+    build_docs_args.append('--in_standard_docker')
+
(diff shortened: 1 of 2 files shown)


nik9000:
Thanks for reviewing @ddillinger! I really do want to shift folks to our docker image, but I'd like them to take their time doing it....

PR #590, opened 2019-02-07
Drop generated docs

nik9000:
We've moved the generated documentation to the `built-docs` repository
so we may as well remove the files from the checkout. This won't help
with the size of the clone but it can certainly help with the size of
the checkout!


nik9000:
`Files changed ∞` Nice!

PR #594, opened 2019-02-11
Fix documentation_links path for Kibana

epixa:
This file moved in the Kibana repo, so the path needs to be updated here.


epixa:
I'm going to close this since @nik9000 was implementing a more robust solution.

PR #595, opened 2019-02-11
Docker infrastructure changes for jenkins

nik9000:
1. Add a packger-cache.sh script to build the docker image as part of
building the packer image.
2. Switch the mechanism we use to share the committer information to one
that is a little more precise and more friendly to CI.
3. Make sure to pipe return code from running docker into exit code.

### DIFF
--- a/build_docs
+++ b/build_docs
@@ -99,7 +99,6 @@ def run_build_docs(args):
     open_browser = False
     args = Args(args)
     saw_out = False
-    expected_return_code = 0
     resource_count = 0
     arg = args.next_arg()
     while arg is not None:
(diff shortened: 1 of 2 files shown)


build_docs
```diff
@@ -273,7 +291,8 @@ if __name__ == '__main__':
     try:
         logging.basicConfig(level=logging.INFO)
         build_docker_image()
-        run_build_docs(argv[1:])
+        if not ['--just-build-image'] == argv[1:]:
```
jarpy: I love the down-to-earth flag name. :)

nik9000:
Thanks for reviewing @jarpy!

PR #596, opened 2019-02-11
Docker: Fail gracefully if run as root

nik9000:
The docker based docs build assume that you are running it as a normal
user and will likely fail if you are running it as root. Specifically
running `--open` will fail because `nginx` will fail to switch to the
"nobody" user.


nik9000:
Thanks for reviewing @polyfractal and @jarpy!

PR #597, opened 2019-02-11
Clean up argument handling

nik9000:
Drops support for single dashed long-opts from the docker build. Drops
support for "abbreviated" options, drops support for starting options
with a `+`, and drops support for specifying options in any case from the
perl build. I haven't see folks use any of the last three options and
they are incompatible with the docker build and would be very difficult
to *make* compatible with the docker build. I *have* seen folks use
single dashed long options. This will fail those single dashed arguments
on the docker build *only* and tell the user to use the double dashed
option.

I believe this might break some folks that are relying on undocumented
options, mostly the "automatic abbreviations" that the script supported.
For command line users I don't foresee an issue though - they'll get the
usage message and be fine. The bigger problem might be scripts that
build the docs. The [kibana] script looks ok. Elasticsearch uses
doc_build_aliases.sh. Cloud's script looks ok. I *suspect* we're ok
here.

[kibana]: https://github.com/elastic/kibana/blob/master/src/docs/docs_repo.js

Edit: The original version of this added support for the single dashed options to the docker build but I ended up dropping it after the conversation below.

### DIFF
--- a/build_docs
+++ b/build_docs
@@ -30,6 +30,7 @@ from __future__ import print_function
 import logging
 from os import environ, getgid, getuid
 from os.path import basename, dirname, exists, isdir, join, realpath
+import re
 import subprocess
 from sys import platform, version_info
 import time
@@ -205,10 +206,15 @@ def run_build_docs(args):
 
 class Args:
     def __init__(self, args):
-        # Replace equals delimited arguments with the space delimiated versions
-        # so it is simpler to iterate over them
+        # Normalize the args so it is simpler to iterate over them
         self.args = []
         for arg in args:
+            # Replace `-foo` style with `--foo` style
+            m = re.match('^-[^-]+$', arg)
+            if m:
+                arg = '-' + m.group(0)
+
+            # Replace `--foo=bar` style with `--foo bar` style
             split = arg.split('=')
             if len(split) > 2:
                 raise UserError('Invalid argument [%s]' % arg)
--- a/build_docs
+++ b/build_docs
@@ -209,15 +209,14 @@ class Args:
         # Normalize the args so it is simpler to iterate over them
         self.args = []
         for arg in args:
-            # Replace `-foo` style with `--foo` style
-            m = re.match('^-[^-]+$', arg)
-            if m:
-                arg = '-' + m.group(0)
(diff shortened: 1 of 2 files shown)


nik9000:
> One specific _complaint_ I have seen with single-dash long options is that they fail to disambiguate between themselves (e.g. `-version`) and single-dash-multiple-flag options (e.g. `-auxww`) which I suppose brings us back to the land of the surprising!

Personally I dislike single dashed arguments as well and if `build_docs.pl` script didn't already support them I wouldn't implement them. As it supports them my choices are:
1. Do nothing and let the perl script support them but the docker script fail in strange ways
2. Add support to the docker script for them
3. Drop support in the perl script for them
4. Explicitly fail the docker script with them

With changes 2-4 I went with option 3 because I haven't seen anyone use them. I still fear that folks might be accidentally using the "automatic abbreviation" feature which could lead to really sneaky failures!

For the single dashed long opts I've seen folks use them. Several in the first day of attempting to use the docker script. I'm weary of dropping support for them in the perl script because I don't want to hard break folks. So I went with adding support to the docker script for them mostly because I hadn't thought of option 4. Would you be more comfortable with option 4? That is, explicitly catch the single dashed arguments in the docker script and tell the user that we don't support them? That feels like a fairly kind way to drop this support. Folks will get a nice error message when they try to switch to the docker script. And they are already making a change anyway so they'll have a chance to update their arguments.

ddillinger:
That does sound good, it's a transition toward standard rather than away, while helping out any existing users of the nonstandard behavior with what is going on.

nik9000:
> That does sound good, it's a transition toward standard rather than away, while helping out any existing users of the nonstandard behavior with what is going on.

Done!

nik9000:
Thanks for talking through this @ddillinger!

jarpy:
What bright spark thought that single-dash longopts were a good idea? In the old days, the only place that they would bite you was, like, `java -version`. Now, there seems to be a plague of the horrid little things coming out of the Go community. The _Go_ community! I thought those (we?) people were pragmatic.

nik9000:
Oh @jarpy! You are wonderful.

PR #598, opened 2019-02-12
Docker: Fix default destionation directory

nik9000:
When I first made `build_docs` I picked the wrong default output
directory when you specify `--doc`. It should be `./html_docs`. I made
it `../html_docs`. See the subtle difference? I didn't notice for a
while at first. Then I was really confused. Then I tracked down the
extra `dirname`. Then I fixed it. This is the fix.


nik9000:
Thanks for reviewing @jarpy!

PR #599, opened 2019-02-12
Drop printing the path on startup

nik9000:
This doesn't seem to be needed anymore, especially once we switch CI to
building with docker.


nik9000:
Thanks for reviewing @jarpy!

Issue #591, opened 2019-02-08
Docker based build issue tracking


nik9000:
Some of you may know that we use an end of life tool, [AsciiDoc][], to build
all of our documentation. Back in 2017 they declared they'd stop releasing
updates and we've been stuck ever since. I've undertaken a project to migrate
us from that tool to an actively developed one, [Asciidoctor][]. Since it has
different dependencies I've taken the advice of my new colleagues at infra to
manage those dependencies with docker. And I'd like some beta-tester/victims to
try out the docker based build.

To build the docs the newfangled docker way you make sure you have docker
installed and then run your usual `build_docs.pl` command but without the
`.pl`. So if you are used to running

```
~/Code/Elastic/docs/build_docs.pl --doc ~/Code/Elastic/kibana/docs/index.asciidoc --chunk 1 --out ~/Code/Elastic/built_docs/ --open --lenient
```

then you can use docker for the build by running

```
~/Code/Elastic/docs/build_docs --doc ~/Code/Elastic/kibana/docs/index.asciidoc --chunk 1 --out ~/Code/Elastic/built_docs/ --open --lenient
```

All of the usual arguments to `build_docs.pl` *ought* to work with the exception
of `--all` on macOS. I'm still [fighting][] with that but not so many people
use it and I'm already in contact with them. If you have any issues with the
docker based build please report them at the [tracking issue][] on github.

IMPORTANT: Just using the docker based build doesn't switch you from AsciiDoc
to Asciidoctor. You *can* use Asciidoctor by adding `--asciidoctor` to the
command line arguments and some books build fine with it but not all of them
do. I have an [asciidoctor tracking issue][] as well, but I that is a lot
further from being done. Feel free to try it though!

ALSO SORT OF IMPORTANT: The docker based build is fairly "simple" and put
together with as few moving parts as I could manage. I'd prefer to delay a
docker purity religious war at this stage. I'm happy to make improvements along
those lines later but I'm trying for progress over perfection here. Especially
when perfection has a few diametrically opposed definitions. If none of that
makes any sense to you then great! It is like the EMACS vs VI debate but with
docker instead. And I picked nano, probably upsetting all sides.

[AsciiDoc]: https://github.com/asciidoc/asciidoc/releases/tag/8.6.10
[Asciidoctor]: https://asciidoctor.org/
[fighting]: https://github.com/elastic/docs/pull/589
[tracking issue]: https://github.com/elastic/docs/issues/591
[...]

PR #601, opened 2019-02-12
Docker: Fix docs path mounting

nik9000:
Fix the paths that docs repos are mounted into the container when
`--doc` and `--resource` are used to allow building more books. These
books expect the all repositories to be cloned into the same root
directory which is "normal" but isn't exactly how docker did it. It *is*
how `--all` does it. But users don't always do it that way....

This fixes the mounted paths so they are always mounted into
`/doc/{reponame}` in docker. Which *should* make the paths even more
consistent then when they are built locally. This should make it
possible to build the stack reference with docker.

Relates to #591

### DIFF
--- a/build_docs
+++ b/build_docs
@@ -96,11 +96,37 @@ def run_build_docs(args):
 
     build_docs_args.append('--in_standard_docker')
 
+    mounted_doc_repo_roots = set()
+    mounted_doc_repo_names = set()
+
+    def mount_docs_repo_and_dockerify_path(repo_search_path, path):
+        """Adds a mount for the root of the repository into the docker
+        container and rewrites the path so that it is inside that mount.
+
+        If the repo happens to already be mounted we won't add two mounts.
+        """
+        repo_root = subprocess.check_output(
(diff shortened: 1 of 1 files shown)


nik9000:
I'm *so* glad the script is in python and not shell.

build_docs
```diff
@@ -96,23 +96,45 @@ def run_build_docs(args):
...
+        If the repo happens to already be mounted we won't add two mounts.
+        """
+        repo_root = subprocess.check_output(
+                ['git', 'rev-parse', '--show-toplevel'],
+                cwd=repo_search_path)
+        repo_root = repo_root.decode('utf-8').strip()
+        repo_name = basename(repo_root)
+        if repo_root not in mounted_doc_repo_roots:
+            if repo_name in mounted_doc_repo_names:
+                raise ArgError("Can't mount two repos with the same " +
```
ddillinger: This is some good thinkin' right here

nik9000:
:heart: @ddillinger!

PR #603, opened 2019-02-12
Switch MS17 to current

kellyemurphy:
Changes the ESS current to MS17, removes MS15. To be merged when MS 17 releases.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -427,8 +427,8 @@ contents:
             prefix:     en/cloud
             tags:       Cloud/Reference
             subject:    Elastic Cloud
-            current:    release-ms-15
-            branches:   [ release-ms-17, release-ms-15, saas-release, saas-release-heroku ]
+            current:    release-ms-17
+            branches:   [ release-ms-17, saas-release, saas-release-heroku ]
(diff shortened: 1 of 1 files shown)


kellyemurphy:
Do we normally do three PRs? (1. Adding new branch, 2. Changing new branch to current, and 3. Removing old branch) If so, we need to update our Cloud doc release process. Happy to do whatever you think is best, just LMK.

nrichers:
@kellyemurphy Don't merge this, yet, there's a late-breaking issue with MS 17. 

> Do we normally do three PRs? (1. Adding new branch, 2. Changing new branch to current, and 3. Removing old branch) If so, we need to update our Cloud doc release process. Happy to do whatever you think is best, just LMK.

The reason I mentioned this is that both yanking the old current and replacing it with the new current seems to cause some hiccups during deploying. This is mostly from anecdotal evidence, but it never seems to have gone smoothly and sometimes you still get the old current docs or 404s as a result. As discussed during our C! syncs, it might be safer to do this as a two-step process, but we can try this PR here as-is. If we find that flipping current this way once again creates issues, then that would the third time or so, at which point we can update our process docs.

PR #604, opened 2019-02-12
Asciidoctor: Copy admonition images

nik9000:
Copies the images used by admonitions like `WARNING` and `NOTE` and the
change admonitions like `coming`.


nik9000:
Thanks for the review @ddillinger!

PR #607, opened 2019-02-13
Add a --self-test option

nik9000:
Adds a `--self-test` option to `build_docs` to be invoked by CI for
periodic or pull request builds. It runs `make check` in the docker
container, relying on the `Dockerfile` to resolve the dependencies like
`pycodestyle`, `rspec`, and `rubocop`.

I fixed the single `pycodestyle` failure and some of the `rubocop`
failures. The rest of the `rubocop` failures I've suppressed though I
intend to fix them in a follow up change.

I also build the README.asciidoc file and assert that it produces a few
files that are expected.


Dockerfile
```diff
@@ -46,8 +49,20 @@ RUN install_packages \
...
+#   * asciidoctor-diagram
+#   * asciimath
+# * Used to check the build in CI
+#   * rubocop
+#   * rspec
 RUN gem install --no-document \
   asciidoctor:1.5.8 \
   asciidoctor-diagram:1.5.12 \
   asciimath:1.0.8 \
+  rubocop:0.64.0 \
```
jarpy: Thank you for your cooperation, Citizen.

resources/asciidoctor/.rubocop.yml
```diff
@@ -0,0 +1,116 @@
+AllCops:
+  TargetRubyVersion: 2.3
+
+Layout/AlignHash:
+  Enabled: false
+
```
jarpy: Are you intending to turn up more checks over time? How many does this leave enabled? :laughing:

nik9000: There are at least three!

PR #608, opened 2019-02-13
Asciidoctor: Fix images for change admonitions

nik9000:
I must have been taking crazy pills when I wrote support for copying the
images for change admonitions because I never checked if they worked for
real files. They don't, because we don't have images with the names I
assumed we used. We use the `note` and `warning` images built in to
asciidoc for our fancy new admonitions. This fixes the copying code to
line up with the rendered html.


resources/asciidoctor/spec/change_admonishment_spec.rb
```diff
@@ -31,26 +31,26 @@
...
-      === Some Reference
-    ASCIIDOC
-    expected = <<~DOCBOOK
-      <chapter id="_example">
-      <title>Example</title>
-      <note revisionflag="#{revisionflag}" revision="some_version">
-      <simpara>See <xref linkend="some-reference"/></simpara>
-      </note>
-      <section id="some-reference">
-      <title>Some Reference</title>
```
nik9000: The indentation was just totally wrong here....

nik9000:
>  would CI and the unit tests have caught this? Even with this change, will they, going forward?

Not as written, no. I have some ideas on how to catch them.

nik9000:
> Not as written, no. I have some ideas on how to catch them.

Well, that isn't quite fair. The unit tests as I've got them now would have caught this. I just had bad unit tests. I think an integration style test is in order. I'll work on that soon.

PR #611, opened 2019-02-14
Drop support for asciidoctor outside of docker

nik9000:
We initially supported asciidoctor in the same way that the docs build
supported everything else: by copying the source into the repo. Now that
we have docker to resolve our dependencies we don't need to do that. This
blasts asciidoctor from the repo and the infrastructure for copying it
locally.


jarpy:
This diff is so red, GitHub can't show even show it.

nik9000:
@elasticmachine, run self-test

nik9000:
Self test failed here! Interesting!

nik9000:
The self-test doesn't fail locally. I'm going to kick off an intake build and have a look at that too.

nik9000:
The self-test failure is caused by different path iteration order and I have proposed a fix in #615.

PR #615, opened 2019-02-14
Asciidoctor: Consistent sorting in warning message

nik9000:
When our copy images extension can't find an image it outputs *all* of
the paths that it searched so you can be super sure it searched all the
places that you wanted it to search. This sorts the output so that the
error message is consistent. Which is mostly nice for tests. I mean, it
is nice for stuff like reproducibility of errors locally too, but it is
super nice for tests. And it is my tests that are failing.


nik9000:

PR #616, opened 2019-02-15
Add asciidoctor to --self-test

nik9000:
Adds rendering the `README.asciidoc` with Asciidoctor in `--self-test`
as a test of Asciidoctor. These two renders can act as a basic
integration test of the docs build. Right now they just validate that
neither AsiiDoc nor Asciidoctor crash when rendering the
`README.asciidoc` file, but I expect to be able to flesh out the tests
in follow up changes.


README.asciidoc
```diff
@@ -123,7 +123,7 @@ to a single XML file for DocBook to process.  _Chunking_ is the process of
 breaking that XML file up into multiple HTML pages.
 
 By default, each _part_ (`= Part Title`) and _chapter_ (`== Chapter Title`) is
-``chunked'' into a separate HTML file. However, for bigger books, such as the
+"chunked" into a separate HTML file. However, for bigger books, such as the
```
nik9000: The ``word'' syntax is not supported by Asciidoctor. It inserted a smart double quote. The `"` just renders a dumb one which seems fine.

README.asciidoc
```diff
@@ -437,7 +437,8 @@ and
 <1> Any headers in the appendix or in the preface start
     out-of-sequence at `level 2`, not at `level 1`.
 
-[sect3]
+ifdef::asciidoctor[[section]]
+ifndef::asciidoctor[[sect3]]
```
nik9000: It is rare to have a section called `Glossary` that isn't actually the glossary. So AsciiDoc has special syntax to make it into a normal section (`[sect3]`). Asciidoctor also has special syntax, but it isn't the *same* special syntax.

README.asciidoc
```diff
@@ -550,17 +551,13 @@ ifdef::env-github[]
...
-`&#x60;single quoted'`::    `single quoted'
 `&#x5e;superscript^`::      ^superscript^
 `&#x7e;subscript~`::        ~subscript~
 endif::[]
 ifndef::env-github[]
 +&#x5f;emphasis_+::         _emphasis_
 +&#x2a;bold*+::             *bold*
 +&#x60;mono'+::             `mono`
-+&#x60;`double quoted''+::  ``double quoted''
-+&#x60;single quoted'+::    `single quoted'
```
nik9000: `'` works just fine....

README.asciidoc
```diff
@@ -1042,14 +1039,13 @@ GET /_search
 
 [NOTE]
 ================================
-The ``View in Console'' links will only work if the docs are viewed in web-browser mode, as follows:
+The "View in Console" links will only work if the docs are viewed in web-browser mode, as follows:
 
 [source,sh]
 ---------------
-build_docs -d my_doc.asciidoc --open <1> --web <2>
+build_docs -d my_doc.asciidoc --open <1>
```
nik9000: Asciidoctor doesn't support `<1>` inside the text. It has to be at the end. Also, `--web` isn't a thing any more.

lib/ES/Util.pm
```diff
@@ -191,6 +195,10 @@ sub build_single {
 
     my ( $output, $died );
     if ( $asciidoctor ) {
+        my $dest_xml = $index->basename;
```
nik9000: This was just wrong before. It happened to work because we only built books that started at a file named `index.asciidoc`.

nik9000:
And now I can go to bed!

PR #618, opened 2019-02-16
Asciidoctor: Ruby style

nik9000:
Enables many of the currently disabled rubocop checks and fixes them. The
biggest change is in the `copy_images` extension which is now split into
two classes, on that processes blocks looking for candidates to copy and
one that does the actual copying. Beyond that I've converted many tests
from blocks into guards. Rubocop seems to really like guards and I find
them pretty readble, partially because they force you to create a new
method in some cases and those methods can have nice names.

### DIFF
--- a/resources/asciidoctor/.rubocop.yml
+++ b/resources/asciidoctor/.rubocop.yml
@@ -54,31 +54,7 @@ Metrics/MethodLength:
 Metrics/PerceivedComplexity:
   Enabled: false
 
-Naming/ConstantName:
-  Enabled: false
-
-Style/BlockDelimiters:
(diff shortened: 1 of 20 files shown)


nik9000:
@ddillinger, could you have a look at this when you get a chance? It ends up being noop but it makes rubocop happy.

ddillinger:
Yeah, I wasn't about to +1 this without passing CI tests, this is a whole lot of noise for a "noop". Very glad those tests are there now.

nik9000:
> This PR would have benefited from some structured commits, it was kind of hard to digest as a big lump.

I'll do it next time!

PR #619, opened 2019-02-16
Add asciidoctor to --self-test

nik9000:
Adds rendering the `README.asciidoc` with Asciidoctor in `--self-test`
as a test of Asciidoctor. These two renders can act as a basic
integration test of the docs build. Right now they just validate that
neither AsiiDoc nor Asciidoctor crash when rendering the
`README.asciidoc` file, but I expect to be able to flesh out the tests
in follow up changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,10 @@ SHELL = /bin/bash -eux -o pipefail
 MAKEFLAGS += --silent
 
 .PHONY: check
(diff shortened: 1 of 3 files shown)


nik9000:
Since this was already approved when it was #616 I'll merge it once jenkins is happy with it.

PR #620, opened 2019-02-16
Force --rely_on_ssh_auth

nik9000:
Always function as though `--rely_on_ssh_auth` is provided when `--all`
is provided. The http-based auth isn't simple to make compatible with
the docker build and ssh-agent based auth is what we do in CI anyway.

This leaves the `--rely_on_ssh_auth` argument but as a noop so that CI
doesn't break. The argument is new enough that I don't expect anyone
other than CI to be relying on it so once I remove it from CI then I
can remove it from the script.

### DIFF
--- a/build_docs
+++ b/build_docs
@@ -112,7 +112,33 @@ def run_build_docs(args):
     arg = args.next_arg()
     while arg is not None:
         build_docs_args.append(arg)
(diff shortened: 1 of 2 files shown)


nik9000:
@ddillinger, could you take a look at this when you get a chance?

PR #624, opened 2019-02-19
Asciidoctor: Rename change admonition

nik9000:
The "change admonitions" are "admonitions" not "admonishments". They are
"authoritative council" not "firm warnings". Words are hard. For what it
is worth, the AsciiDoc source also has this error in some places.
Asciidoctor gets right every time.


nik9000:
Thanks @ddillinger!

PR #625, opened 2019-02-19
self-test: File list matches

nik9000:
Adds a test that asserts that building with `Asciidoctor` produces the
same files as building with `AsciiDoc`. It doesn't actually pass right
now which is wonderful! Tests that find problems are great. We add a
`grep -v` line to the test to ignore the parts that don't pass for now
and we'll fix the failing part in a follow up.


nik9000:
Baby steps!

nik9000:
Thanks @ddillinger! I'm working on getting it to pass now....

PR #626, opened 2019-02-20
Self-Test: Generalize wrapping and document

nik9000:
Allows you to send arbitrary arguments to make. This makes it possible to
do things like "only run rubocop" with:

```
./build_docs --self-test -C resources/asciidoctor rubocop
```

While it doesn't take *that* long to run the entire `--self-test`
sequence, it is wonderful to be able to selectively run tests.

### DIFF
--- a/build_docs
+++ b/build_docs
@@ -434,14 +434,20 @@ if __name__ == '__main__':
     try:
         logging.basicConfig(level=logging.INFO)
         build_docker_image()
-        if ['--self-test'] == argv[1:]:
+        if len(argv) > 2 and '--self-test' == argv[1]:
+            cwd = realpath('.')
+            if not cwd.startswith(DIR):
+                raise ArgError(
+                    '--self-test must be invoked from within the repo')
+            docker_cwd = '/docs_build/' + cwd[len(DIR):]
             cmd = ['docker', 'run']
             cmd.extend(standard_docker_args())
+            cmd.extend(['--workdir', docker_cwd])
             if stdout.isatty():
                 cmd.append('-t')
             cmd.extend(['elastic/docs_build', 'make'])
             cmd.extend(['--no-builtin-rules'])
-            cmd.extend(['-C', '/docs_build', 'check'])
+            cmd.extend(argv[2:])
             returncode = subprocess.call(cmd)
             exit(returncode)
         elif not ['--just-build-image'] == argv[1:]:
--- a/build_docs
+++ b/build_docs
@@ -434,7 +434,7 @@ if __name__ == '__main__':
     try:
         logging.basicConfig(level=logging.INFO)
         build_docker_image()
-        if len(argv) > 2 and '--self-test' == argv[1]:
+        if len(argv) >= 2 and '--self-test' == argv[1]:
             cwd = realpath('.')
             if not cwd.startswith(DIR):
                 raise ArgError(
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -802,11 +802,12 @@ sub restart {
 #===================================
 sub usage {
 #===================================
+    my $name = $Opts->{in_standard_docker} ? 'build_docs' : $0;
     say <<USAGE;
 
     Build local docs:
 
-        $0 --doc path/to/index.asciidoc [opts]
+        $name --doc path/to/index.asciidoc [opts]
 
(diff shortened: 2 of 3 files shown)


nik9000:
OOoh! self-test job failed! Nice. I'll check it.

build_docs
```diff
@@ -434,14 +434,20 @@ if __name__ == '__main__':
     try:
         logging.basicConfig(level=logging.INFO)
         build_docker_image()
-        if ['--self-test'] == argv[1:]:
+        if len(argv) >= 2 and '--self-test' == argv[1]:
```
ddillinger: Doesn't this mean `--self-test` must be positionally the first command-line argument then?

Adding `argparse` adds like maybe 4-5 lines of code, with the benefits of both future-proofing and extensibility.

build_docs.pl
```diff
@@ -848,4 +849,25 @@ sub usage {
...
+    you can make specific targets. Some examples:
+
+    Execute all tests:
+        $name --self-test
+
+    Execute all of the tests for our extensions to Asciidoctor:
+        $name --self-test -C resources/asciidoctor
+
+    Run rubocop on our extensions to Asciidoctor:
+        $name --self-test -C resources/asciidoctor rubocop
```
ddillinger: Full transparency, I am still pretty "ehhh" on two ways to do the same thing instead of standardizing on e.g. `make test` and directing users to it. Fully granting the distinction between the goal and the path we take to get there, of course.

build_docs
nik9000: It does indeed need to be first, yeah.

I'll look into argparse!

build_docs.pl
nik9000: `make test` won't work because it isn't in the docker container so it probably won't have the dependencies available. If it did i wouldn't have made this.

build_docs
nik9000: So I had looked into argparse when I first implemented this file and I didn't like it then because there are arguments that have to be passed unchanged into the perl script. I *think* that is still the right call for that branch. The `--self-test` stuff feels pretty similar in that my goal is to pass arguments unchanged to make. It feels a little cleaner to force `--self-test` to be the first argument so it is a little more clear what goes to make.

build_docs.pl
ddillinger: I'm not really worried about how the _tooling_ accomplishes running it (barring a case where a maintenance headache gets introduced), but I am concerned with us telling users (and this is the usage string) to do this kind of stuff. It seems nicer if users always go to the same place with the same expectations. For example, they should be able to type `make rubocop` from the top level, and have that do the expected thing they wanted. 

> `make test` won't work because it isn't in the docker container so it probably won't have the dependencies available.

I don't understand the problem. If something is missing from the container, add it to the container? At the same time I also am unsure how the container is necessarily involved here, if this is a wrapper around make then it's running make, and if that isn't what's going on then this usage string is inaccurate, because that's what it says it does.

Additionally, speaking as something a user would do, a user would check out the repo, cd into it, and type `make test`. That's on their host; running any containers as a result of that is directly within our control and also only marginally their problem -- like, they need to have docker, but that's about it. That's what I'm really getting at in this comment thread: a simplified, consistent, and unsurprising experience for contributors. 

I imagine that over time they will become completely insulated from `build_docs.pl` ever talking to them at all, at least if I understand the current plan? Which ultimately ends in `build_docs.pl` no longer existing at all?

nik9000: Right now `build_docs` runs docker which runs make to execute the tests. That is the only way to run these tests. This way the Makefiles get to "know" that they are in our standard docker container. The idea is that the only thing that needs to be portable is `build_docs`.

With this change the user doesn't have to know about `build_docs.pl`. I rename it to `build_docs` in the usage instructions if you are running in docker.  They do still have to know about `build_docs` which is weird. Folks are going to be used to make. But I *like* that I know exactly the shape of the environment that the makefile is executing within. 

If there was a way I could hook into make's startup to transparently restart itself in the docker container that'd be lovely.

ddillinger: I'm sure there's some way around it. I'm not _so_ worried that it can't be on a future PR, though. Let's think about it more, and I will stop expecting anything _here_ about it. :)

PR #628, opened 2019-02-21
Asciidoctor: Fix cramped include

nik9000:
Fixes the cramped include preprocessor to be compatible with the elastic
compat preprocessor. The trouble here is that the cramped include
preprocessor adds a frozen `''` to the stream of lines and the elastic
compat processor attempts to modify it. It is tempting to remove all of
the string mutation from the elastic compat preprocessor but that isn't
compatible with how preprocessors have to hook in to asciidoctor.
Asciidoctor expects them to mutate the strings they are given. Or, to be
more precise, it expects that it only needs to call `process` on each
line once, and that calling it will mutate the line in place. Which
isn't possible for the frozen empty line that the cramped include
preprocessor adds.

We instead fix the problem by adding a mutable empty line to the stream.
Making asciidoctor happy while slightly upsetting me.

A better integration test would have caught this, but for now we include
the compat preprocessor in the cramped include preprocessor's unit test
to check for it. I'll work on a better integration test presently.

### DIFF
--- a/integtest/included.asciidoc
+++ b/integtest/included.asciidoc
@@ -1,0 +1,1 @@
+I am tiny.
--- a/Makefile
+++ b/Makefile
@@ -19,36 +19,5 @@ asciidoctor_check:
 	$(MAKE) -C resources/asciidoctor
 
 .PHONY: integration_test
-integration_test: expected_files_check same_files_check
-
-.PHONY: expected_files_check
-expected_files_check: /tmp/readme_asciidoc
-	# Checking for expected html files
-	[ -s $^/index.html ]
-	[ -s $^/_conditions_of_use.html ]
-	# Checking for copied images
-	[ -s $^/resources/cat.jpg ]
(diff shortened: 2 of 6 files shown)


ddillinger:
> I'll work on a better integration test presently.

On this PR? I think it would be nice if it could be on this PR, to come along with the fix it checks for. :)

nik9000:
> On this PR? I think it would be nice if it could be on this PR, to come along with the fix it checks for. :)

I can do. I feel bad making the PR 30x bigger for integration testing.

nik9000:
@ddillinger take a look now. It has an integration test that would have caught this. I have to learn more make to make the test cleaner, but it shows the kind of thing I was thinking of doing.

integtest/included.asciidoc
```diff
@@ -0,0 +1 @@
+I am tiny.
```
ddillinger: i am tiny

nik9000:

PR #632, opened 2019-02-22
Asciidoctor: Freeze constants

nik9000:
Freezes all of the variables that have CONSTANT_CASE so we don't
accidentally mutate them and enables rubocop enforcement for this.


nik9000:
Thanks for reviewing @ddillinger! Mutable *global* state is ought right terrifying.

PR #633, opened 2019-02-22
Asciidoctor: Explain unary plus on string

nik9000:
Adds a comment explaining a the fairly string looking `+''` construct
in the cramped include extension.


nik9000:
Thanks for reviewing @ddillinger! I figured it would be inscrutable without `git blame`....

PR #635, opened 2019-02-22
Self-Test: Add html diffing

nik9000:
Compare the html built by AsciiDoc and Asciidoctor as part of the
self-test. Like all good tests, it doesn't pass at first. I've
temporarilly ignored all of the failures and have added them to the
main asciidoctor issue or filed them as their own issue.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,6 +14,7 @@ LABEL MAINTAINERS="Nik Everett <nik@elastic.co>"
 #   * openssh-client (used by git)
 #   * openssh-server (used to forward ssh auth for git when running with --all on macOS)
 #   * perl-base
+#   * python (is python2)
 #   * xsltproc
 # * To install rubygems for asciidoctor
 #   * build-essential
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -14,6 +14,7 @@ LABEL MAINTAINERS="Nik Everett <nik@elastic.co>"
 #   * openssh-client (used by git)
 #   * openssh-server (used to forward ssh auth for git when running with --all on macOS)
 #   * perl-base
+#   * python (is python2)
```
ddillinger: This reminds me. Is there an issue for changing to python 3, and explicitly weeding out and dropping support for python 2 wherever it may be found? Because python 2 has less than a year to live. I didn't see one on a quick issue search, but I might have just missed it.

nik9000: Not *really*, no. But, hey, I'll make [one](https://github.com/elastic/docs/issues/636).

nik9000:

PR #637, opened 2019-02-22
Mac: Nice warning if you don't have an identity

nik9000:
Supporting `--all` on mac is "fun" because we rely on ssh auth and, unlike
Linux's lovely bind mounts, macOS can't share the ssh authorization
directly. Instead we have to carefully bounce the ssh authorization into
the container by `ssh`ing into it. One of the steps in this process
requires that the user has an ssh identity configured with their agent.
This is fairly easy to do, but if you haven't done it the error message
was garbage. This catches the original message from `ssh-add` and sends
the user a nicer one that includes a link to instructions.

### DIFF
--- a/build_docs
+++ b/build_docs
@@ -37,6 +37,7 @@ import time
 import webbrowser
 
 DOCKER_BUILD_QUIET_TIME = 3  # seconds
+SSH_AGENT_HELP = 'https://help.github.com/en/articles/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent#adding-your-ssh-key-to-the-ssh-agent'  # noqa
 
 DIR = dirname(realpath(__file__))
 logger = logging.getLogger('build_docs')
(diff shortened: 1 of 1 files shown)


build_docs
```diff
@@ -37,6 +37,7 @@ import time
 import webbrowser
 
 DOCKER_BUILD_QUIET_TIME = 3  # seconds
+SSH_AGENT_HELP = 'https://help.github.com/en/articles/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent#adding-your-ssh-key-to-the-ssh-agent'  # noqa
```
ddillinger: What does `# noqa` mean?

build_docs
```diff
@@ -250,6 +251,10 @@ def forward_ssh_auth_into_container(container, docker_run):
     handle_popen(get_ssh_public_key, record_public_key)
     keys = "\n".join(public_keys) + "\n"
     if get_ssh_public_key.returncode != 0:
+        if 'The agent has no identities.' in keys:
+            raise ArgError('-all requires an identity be registered with the '
```
ddillinger: `s/-all/--all/`

build_docs
nik9000: It tells pycodestyle to ignore the super long line.

build_docs
nik9000: Yeah!

nik9000:
Thanks for reviewing @lcawl and @ddillinger!

PR #638, opened 2019-02-22
Fix skiplinkcheck

nik9000:
`--skiplinkcheck` and `--linkcheckonly` were documented as only working
with `--doc` but they are actually only used with `--all`. This fixes
the documentation and the preflight checks to reflect that reality.


lcawl:
Good catch!  Makes sense, since we only do the link checks on --all

nik9000:
> Good catch! Makes sense, since we only do the link checks on --all

I bumped into this yesterday when I tried to skip the links check!

nik9000:
Thanks for looking @ddillinger and @lcawl!

PR #639, opened 2019-02-22
Asciidoctor: Fix deprecation

nik9000:
When I added deprecations I created a `note` element but it should have
been a `warning` element. This is one of the things that the new
`html_diff`-based integration tests caught!


nik9000:

PR #640, opened 2019-02-22
Asciidoctor compatibility for part of readme

nik9000:
Our README.asciidoc file has source listing containing asciidoc code
which is rare for our books. Asciidoctor was rendering it incorrectly
because of its "automatic comment detection" that it uses with callouts.
It saw
```
-- <1>
```

And through "that is a SQL comment preceeding a callout so I will eat
the `--`". While *normally* helpful, this is wrong when commenting on
asciidoc source. I work around this by writing
```
pass:[--] <1>
```

Which spits out the desired `--` followed by a callout. To get that
working I had to poke the language declaration line some, especially
because AsciiDoc doesn't support the simpler Asciidoctor syntax for
enabling macros inside of listings.

This another thing that our `html_diff` based integration tests caught.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -372,21 +372,21 @@ Longer `partintro` blocks should be wrapped in an
 http://www.methods.co.nz/asciidoc/userguide.html#X29[_open block_]
 which starts and ends with two dashes: `--`:
 
-[source,asciidoc]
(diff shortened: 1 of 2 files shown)


nik9000:

PR #642, opened 2019-02-25
Clean up Edit Me links

### LINKED ISSUE #634: Edit me link detection not working for README.asciidoc
I added code to detect the appropriate "edit me" link when building docs locally because it makes it easier to compare things. It is broken for `./build_docs --doc README.asciidoc`. It seems to work everywhere else.

nik9000:
A minor fix and minor clean up for the "Edit Me" links that allow us to
remove the "temporary workaround" in our `html_diff`ing integration tests:
1. Switch detection of the root of the repository from a loop to a call
out to `git rev-parse --show-toplevel`, a trick Jarpy tought me when he
saw my loop in the docker code.
2. Do not add `./` to the url if the `base_dir` of the document is also
the root of the repository. This is *probably* only true of
`README.asciidoc` in the docs repo, but we test with that file so we
should get it right.

Closes #634

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -212,7 +212,7 @@ sub _guess_opts_from_file {
     $Opts->{root_dir} = $toplevel;
     local $ENV{GIT_DIR} = dir($toplevel)->subdir('.git');
     my $remotes = eval { run qw(git remote -v) } || '';
-    if ($remotes !~ /\s+(\S+[\/:]elastic\/\S+)/) {
+    if ($remotes !~ m|\s+(\S+[/:]elastic/\S+)|) {
         say "Couldn't find edit url because there isn't an Elastic clone";
         say "$remotes";
         return;
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -200,27 +200,26 @@ sub _guess_opts_from_file {
 #===================================
     my $index = shift;
 
-    my $dir = $index->parent;
-    while ($dir ne '/') {
-        $dir = $dir->parent;
-        my $git_dir = $dir->subdir('.git');
-        if (-d $git_dir) {
-            $Opts->{root_dir} = $dir;
-            local $ENV{GIT_DIR} = $git_dir;
-            my $remotes = eval { run qw(git remote -v) } || '';
-            if ($remotes !~ /\s+(\S+[\/:]elastic\/\S+)/) {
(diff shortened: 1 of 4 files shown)


resources/asciidoctor/spec/edit_me_spec.rb
```diff
@@ -89,6 +89,29 @@
...
+      <simpara>Words.</simpara>
+      </chapter>
+      <chapter id="_chapter_2">
+      <title>Chapter 2<ulink role="edit_me" url="www.example.com/docs/resources/edit_me/chapter2.adoc">Edit me</ulink></title>
+      <simpara>Words.</simpara>
+      </chapter>
+    DOCBOOK
+    expect(convert input, attributes).to eq(expected.strip)
+  end
+
```
jarpy: There's a lot of HTML here that doesn't directly support the intent of the test. Could you "zoom in" on the relevant parts with "includes" assertions, rather than a big "is" assertion?

build_docs.pl
```diff
@@ -200,27 +200,26 @@ sub _guess_opts_from_file {
...
+    chdir $original_pwd;
+    unless ( $toplevel ) {
+        say "Couldn't find edit url because the document doesn't look like it is in git";
+        $Opts->{root_dir} = $index->parent;
+        return;
+    }
+    $Opts->{root_dir} = $toplevel;
+    local $ENV{GIT_DIR} = dir($toplevel)->subdir('.git');
+    my $remotes = eval { run qw(git remote -v) } || '';
+    if ($remotes !~ /\s+(\S+[\/:]elastic\/\S+)/) {
```
jarpy: How about an alternate delimiter here, since there are slashes involved?

Like:

```
m|\s+(\S+[/:]elastic/\S+)|
```

nik9000: :+1:

resources/asciidoctor/spec/edit_me_spec.rb
nik9000: I can do that but the *file* is about testing the edit me link itself and its location in the xml is important. So I can't do that for the whole file.

nik9000:
@jarpy, could you have another look? I "zoomed in" on the two tests that only care about the links. The other tests care very much about the location of the edit me link in the docbook so I'd prefer to keep them as they are.

nik9000:
No problem at all! Thanks for reviewing!

PR #643, opened 2019-02-25
Asciidoctor: Handle skipped lines

nik9000:
Our compat-preprocessor wasn't properly handling "skipped" lines
sometimes, mostly because we didn't have any tests for them. This fixes
both problems. It is a prerequisite to switching the "Getting Started"
book to asciidoctor.

### DIFF
--- a/resources/asciidoctor/lib/elastic_compat_preprocessor/extension.rb
+++ b/resources/asciidoctor/lib/elastic_compat_preprocessor/extension.rb
@@ -124,11 +124,13 @@ class ElasticCompatPreprocessor < Asciidoctor::Extensions::Preprocessor
 
         @in_attribute_only_block = true
         line.clear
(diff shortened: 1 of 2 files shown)


nik9000:
@lcawl, I found this when looked into converting the next book in our list!

lcawl:
Thanks, if I understand correctly, this is the cause of the following error when building the 6.7 Getting Started Guide via asciidoctor:

> INFO:build_docs:/docs_build/resources/asciidoctor/lib/elastic_compat_preprocessor/extension.rb:129:in `process_line': asciidoctor: FAILED: /doc/stack-docs/docs/en/getting-started/index.asciidoc: Failed to load AsciiDoc document - undefined method `sub!' for nil:NilClass (NoMethodError)

nik9000:
@lcawl, that is the one!

PR #646, opened 2019-02-26
Add Milestone 18 to ESS builds

nrichers:
Looks like there are some items that cropped up that warrant a Milestone 18 doc build after all. So, as a first step, let's add a build and then we can haggle about ze details. 😸


debadair:
LGTM

Issue #614, opened 2019-02-14
Generalize links checking


nik9000:
Right now we have code in the docs build that checks links from the docs to the docs and we have code that checks links from Kibana to the docs. The Kibana to the docs links checking code uses regexes to grab links from bits of Kibana's code. This seems brittle because Kibana might want to move things around or build links with string concatenations or, well, all kinds of stuff. I don't really care Kibana wants to build links.

I wonder if we'd be better off having `build_docs.pl` look for an executable file in every branch of every repo that it knows about and runs it to get a list of links. This'd rely on two things:
1. Docker to resolve the dependencies that we need.
2. Kibana to write a script to dump their own links.

I'm quite happy to install some node version in the docker image but I really would like the script to be quick so I don't think it should have to resolve dependencies itself. I'm not sure if that is a think Kibana would want to do, but it feels cleaner than the regexes we have now.

nik9000:
I'd love an opinion from folks like @epixa.

PR #651, opened 2019-02-27
Add --conf option and test for --all

nik9000:
Adds support for `--conf` to specify the config file used by the docs
build and uses that to add a test for `--all`. This will allow us to add
more features with more confidence that we're not breaking `--all`. It
makes my life less terrifying.

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -57,10 +57,11 @@ GetOptions(
     'single',  'pdf',     'doc=s',           'out=s',  'toc', 'chunk=i',
     'open',    'skiplinkcheck', 'linkcheckonly', 'staging', 'procs=i',         'user=s', 'lang=s',
     'lenient', 'verbose', 'reload_template', 'resource=s@', 'asciidoctor', 'in_standard_docker',
+    'conf=s',
 ) || exit usage();
 check_args();
 
-our $Conf = LoadFile('conf.yaml');
+our $Conf = LoadFile(pick_conf());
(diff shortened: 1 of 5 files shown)


build_docs.pl
```diff
@@ -294,7 +295,7 @@ sub check_links {
 
     $link_checker->check;
 
-    check_kibana_links( $build_dir, $link_checker );
+    check_kibana_links( $build_dir, $link_checker ) if exists $Conf->{repos}{kibana};
```
nik9000: I opened #614 to talk about how it is odd we have a specific repo here. We can do better, I think.

build_docs.pl
```diff
@@ -57,10 +57,11 @@ BEGIN
     'single',  'pdf',     'doc=s',           'out=s',  'toc', 'chunk=i',
     'open',    'skiplinkcheck', 'linkcheckonly', 'staging', 'procs=i',         'user=s', 'lang=s',
     'lenient', 'verbose', 'reload_template', 'resource=s@', 'asciidoctor', 'in_standard_docker',
+    'conf=s',
 ) || exit usage();
 check_args();
 
-our $Conf = LoadFile('conf.yaml');
+our $Conf = LoadFile(pick_conf());
```
nik9000: I think it'd be nice if we didn't rely on the conf file when building a single document but that is a topic for another time. And having these tests will help when we go to do that.

nik9000:

PR #652, opened 2019-02-27
Drop --rely_on_ssh_auth

nik9000:
It has been a noop for a week or two. We don't use it in CI any more.


nik9000:

PR #654, opened 2019-02-27
Self-Test: Ignore width space after em dash

nik9000:
AsciiDoctor spits out an em dash when there is an em dash between two
words. This is kind of a pleasant surprise because it makes it clear
that we can break words after the em dash. This change teaches the
`html_diff` utility that we use for finding differences in pages that
this particular zero width space doesn't count as a difference.

### DIFF
--- a/integtest/Makefile
+++ b/integtest/Makefile
@@ -5,6 +5,7 @@ MAKEFLAGS += --silent
 check: \
 	style \
 	readme_expected_files readme_same_files \
+	emdash_expected_files emdash_same_files \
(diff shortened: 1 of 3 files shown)


nik9000:
Bleh merge conflict with myself!

nik9000:
Conflict resolved!

nik9000:
@jarpy, could you have a look at this one when you get a chance please?

nik9000:
> Invisible Unicode surgery eh? I don't envy you.

I signed up for it! Thanks for reviewing!

PR #659, opened 2019-03-01
Allow building all docs with a local repo

nik9000:
We current offer two ways to build the docs:
1. Build an arbitrary book with `--doc` and its companion arguments like
`--resource` and `--asciidoctor` and `--chunk`.
2. Build *all* of the books that Elastic publishes with `--all`.

Building an arbitrary book is nice and quick because it is have very
little overhead on top of the time it takes to build the actual book.
But it isn't always the best test because the command that you use to
build the book might not exactly match the invocation that `--all` uses
to build the book. Furthermore, building a single book is not usually
enough! When you are testing documentation changes that you've made
locally you often don't know all of the books that your changes might
effect. The Elasticsearch repository is used in about a dozen books in
combination with a hand full of other repositories. Also! If you want
to build any books that require multiple repositories then you have to
set all of them up appropriately.

Building all of the docs doesn't have any of those problems but it isn't
wasn't compatible with building from the local filesystem. This change
fixes that by adding a `--sub_dir` option to "substitute" and "directory"
in place of some branch of some repository when building all of the docs.
This overrides the standard up-to-date checks to force all books being
built with these subsitutions to be rebuilt every time. It looks like:

```
./build_docs --all --target_repo git@github.com:elastic/built-docs.git \
                --keep_hash \
                --sub_dir elasticsearch:master:~/Code/elasticsearch
```

This is nice because it:
1. Rebuilds all of the books that your local directory is involved with
2. Uses standard switches when building each of those books
3. Does not rebuild books that don't use your local directory

That third thing is actually caused by that new `--keep_hash` flag. It
causes the build's up-to-date checking to use the same hash that was
used the last time the book was built. `--keep_hash` has uses beyond
being paired with `--sub_dir`. In particular, it is useful when testing
what happens when you switch a book to Asciidoctor. You can switch the
book to Asciidoctor in `conf.yaml` and rebuild with `--keep_hash` and
the build will *only* have the asciidoctor change in it.

This also drops `--no_fetch` because `--keep_hash` is better in almost
every way.

The goal of `--sub_dir` isn't so much that people will execute it
locally, but that we can execute it on CI. I'm sure folks will execute
it locally but it pretty heavy because, just like a normal `--all` build
it has to:
1. Clone the `built-docs` repo or fetch updates if it is already cloned.
many, many files in it.
2. Check out the `built-docs` repo. This is truly slow because it has
3. Clone all of the repos used to build every book or fetch updates if
they are already cloned.

I suspect in time we can fix some of these. But for now, this is a heavy
thing.


build_docs
```diff
@@ -103,8 +104,7 @@ def run_build_docs(args):
                 '-v',
                 '%s:/doc/%s:ro,cached' % (repo_root, repo_name)
             ])
-        build_docs_args.append(
-                '/doc/' + repo_name + path.replace(repo_root, ''))
+        return '/doc/' + repo_name + path.replace(repo_root, '')
```
nik9000: I dropped automatically adding it to the list because we don't want to do that for `--sub_dir`.

build_docs.pl
```diff
@@ -489,6 +489,7 @@ sub init_repos {
         user      => $Opts->{user},
         url       => $Opts->{target_repo},
         reference => $reference_dir,
+        keep_hash => 0,
```
nik9000: You can't `keep_hash` on the target repo because the `target_repo` contains the list of hashes to keep!

build_docs.pl
```diff
@@ -532,7 +534,7 @@ sub init_repos {
         else {
             $pm->start($name) and next;
             eval {
-                $repo->update_from_remote() unless $Opts->{no_fetch};
+                $repo->update_from_remote();
```
nik9000: I drop this feature because `--keep_hash` is better in most cases.

build_docs.pl
```diff
@@ -770,7 +783,9 @@ sub usage {
           --skiplinkcheck   Omit the step that checks for broken links
           --linkcheckonly   Skips the documentation builds. Checks links only.
           --rebuild         Rebuild all branches of every book regardless of what has changed
-          --no_fetch        Skip fetching updates from source repos
+          --keep_hash       Build docs from the same commit hash as last time
+          --sub_dir         Use a directory as a branch of some repo
+                            (eg --sub_dir elasticsearch:master:~/Code/elasticsearch)
```
nik9000: I picked this format because it feels right for CI. It is a bit wordy for folks trying to do it locally but, like I say in the commit message, I don't think this is something we'll run locally too much. At least at first....

integtest/Makefile
```diff
@@ -78,34 +81,152 @@ endef
...
-.PHONY: small_all_expected_files
-small_all_expected_files: /tmp/small_all
-	[ -s $^/html/branches.yaml ]
-	grep '<a class="ulink" href="test/current/index.html" target="_top">Test book</a>' $^/html/index.html > /dev/null
-	grep '<meta http-equiv="refresh" content="0; url=current/index.html">' $^/html/test/index.html > /dev/null
-	[ -s $^/html/test/current/index.html ]
+.PHONY: simple_all_expected_files
+simple_all_expected_files:
+	# Test the simplest possible `--all` invocation
+	$(call BUILD_MINIMAL_ALL,/tmp/simple_all)
```
nik9000: I switched to these `call` invocations because they don't bump into make's broken up-to-date checking for directories.

integtest/small_conf.yaml
```diff
@@ -44,10 +44,10 @@ contents:
...
         current:    master
         branches:   [ master ]
-        index:      minimal.asciidoc
+        index:      index.asciidoc
         tags:       test tag
         subject:    Test
         sources:
             -
                 repo:   source
-                path:   .
```
nik9000: `.` isn't a thing we ever see in the real `conf.yaml` so I replaced it here.

integtest/small_conf.yaml
```diff
@@ -44,10 +44,10 @@ contents:
         prefix:     test
         current:    master
         branches:   [ master ]
-        index:      minimal.asciidoc
+        index:      index.asciidoc
```
nik9000: `index.asciidoc` allows me to reuse this file in other tests.

lib/ES/Repo.pm
```diff
@@ -128,25 +130,36 @@ sub _reference_args {
...
+    if ( $self->keep_hash ) {
+        $new = $self->_last_commit(@_);
+    } else {
+        $new = sha_for($branch) or die(
+                "Remote branch <origin/$branch> doesn't exist in repo "
+                . $self->name);
+    }
     $new .= '|asciidoctor' if $asciidoctor;
 
+    return $old ne $new if $self->keep_hash;
```
nik9000: If we're keeping the hash then we don't need to check to see if any of the files that we're using changed so we can just return here.

lib/ES/Repo.pm
```diff
@@ -165,42 +178,45 @@ sub mark_done {
...
+    }
     $new .= '|asciidoctor' if $asciidoctor;
+
     $self->tracker->set_sha_for_branch( $self->name,
         $self->_tracker_branch(@_), $new );
-
 }
 
 #===================================
-sub tree {
```
nik9000: The `tree` subroutine was not used.

lib/ES/Repo.pm
```diff
@@ -300,8 +320,13 @@ sub all_repo_branches {
         for my $branch ( sort keys %$shas ) {
             my $sha = $shas->{$branch};
             $sha =~ s/\|.+$//;  # Strip |asciidoctor if it is in the hash
-            my $log = run( qw(git log --oneline -1), $sha );
-            my ($msg) = ( $log =~ /^\w+\s+([^\n]+)/ );
+            my $msg;
+            if ( $sha eq 'local' ) {
+                $msg = 'local changes';
```
nik9000: I don't think it is worth looking at the substitution branch to try and find a commit message here. We don't even know that the substitution directory is in a git repo....

lib/ES/Repo.pm
```diff
@@ -316,6 +341,7 @@ sub checkout_to {
 #===================================
     my ( $self, $destination ) = @_;
 
+    die 'sub_dir not supported with checkout_to' if %{ $self->{sub_dirs}};
```
nik9000: This method is only used by the `target_repo` so we don't need to support `sub_dirs`. In fact, it doesn't make sense to have it!

lcawl:
This seems like it helps with some of the issues described in https://github.com/elastic/docs/issues/249

I initially thought that it could help with the scenario where we need to know which books to rebuild when we make changes to shared files like https://github.com/elastic/docs/blob/master/shared/attributes.asciidoc.  On closer reading, however, I don't think that's right--these parameters wouldn't catch that scenario and we'll still need to forcefully rebuild all.

nik9000:
> I initially thought that it could help with the scenario where we need to know which books to rebuild when we make changes to shared files like https://github.com/elastic/docs/blob/master/shared/attributes.asciidoc. On closer reading, however, I don't think that's right--these parameters wouldn't catch that scenario and we'll still need to forcefully rebuild all.

I think that is true.

> This seems like it helps with some of the issues described in #249

I think it catches the "Check cross-doc links and perform other post-build verification on specific books using the already-built versions of everything else." bit. I don't think it covers the rest though!

nik9000:
@jarpy, could you have a look at this one when you get a chance please?

nik9000:
Woops, there are a lot of conflicts! I'll see about fixing those as soon as I finish what I've got locally.

build_docs.pl
jarpy: It's fine by me, but aren't hyphens more conventional in longopts than underscores?

integtest/Makefile
```diff
@@ -117,37 +120,155 @@ missing_include_fails_asciidoctor: missing_include.asciidoc
...
-	# Builds "--all" documentation specified by by the "small_conf.yaml" file.
+.PHONY: keep_hash_expected_files
+keep_hash_expected_files:
+	# Test that `--all --keep_hash` doesn't pull new updates
+	$(call BUILD_MINIMAL_ALL,/tmp/keep_hash)
 
-	# First build a repository to use as the source.
+	# REPLACE the minimal documentation in the source repo
+	cp ../README.asciidoc /tmp/source/index.asciidoc
+	cd /tmp/source && \
```
jarpy: Would the ever be run in parallel with itself? Might need a proper tmpdir.

build_docs
```diff
@@ -208,7 +209,19 @@ def run_build_docs(args):
             resource_dir = realpath(args.next_arg_or_err())
             if not isdir(resource_dir):
                 raise ArgError("Can't find --resource %s" % resource_dir)
-            mount_docs_repo_and_dockerify_path(resource_dir, resource_dir)
+            build_docs_args.append(mount_docs_repo_and_dockerify_path(
+                    resource_dir, resource_dir))
+        elif arg == '--sub_dir':
```
jarpy: As you add features, all the manual parsing of `argv` is going to cost you more and more. `argparse` isn't great, but is there any value in considering it over full DIY?

(Normally, I would suggest Click, but I know you can't do packages.)

build_docs.pl
jarpy: If it's worth a review comment, is it worth a code comment?

build_docs.pl
jarpy: Oh, I think we've had that conversation. NVM. :(

integtest/Makefile
nik9000: It'd be difficult to get it to run in parallel with itself but I may as well make a proper tmpdir. You are right about that!

build_docs
nik9000: I've considered it on and off for a while. I think the value that DIY gets me is being able to pass through arguments that I don't know about. I only rewrite the arguments that have anything to do with paths. The cost of DIY is that I need this everywhere. At this point I think it is net simpler, but I'd be receptive to any argument to the contrary. Like, I wonder if it'd be helpful to *force* me to list all of the supported arguments in the python file because we plan to drop the perl file one day.

jarpy: > "I've considered it"

Cool. I trust your judgement.

integtest/Makefile
jarpy: I guess Docker kinda takes care of that, doesn't it? I may have been thinking too old-school.

PR #660, opened 2019-03-01
Allow running --open with --all

nik9000:
This allows you to `--open` the docs built with `--all`. This should be
useful when testing changes that effect all docs, like the redirect work
that I plan on doing soon.

There is a wrinkle though: to make serving the docs locally work we
actually built *different* docs. We'd switch from finding jQuery at
`/static/js/jquery.min.js` to getting it from
`https://www.elastic.co/static/js/jquery.min.js`. This worked, and it
was simple enough to do for the built in python web server that we use
when you run `--open` outside of docker, but we don't want to do that
when we serve all of the docs because most of the docs will have been
built with the standard template.

But now that we have docker we don't have to! In docker we're using
nginx to serve the docs and we can write a few rewrite rules to fix the
missing assets. For good measure we assets rewrite when we're running in
docker even when we're building a single doc. So now building a single
doc is just that much more real!

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -80,7 +80,7 @@ $Opts->{template} = ES::Template->new(
     %$template_urls,
     lenient  => $Opts->{lenient},
     force    => $Opts->{reload_template},
-    abs_urls => $Opts->{doc}
+    abs_urls => ! $running_in_standard_docker && $Opts->{doc},
 );
 
 $Opts->{doc}       ? build_local( $Opts->{doc} )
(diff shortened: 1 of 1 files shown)


build_docs.pl
```diff
@@ -733,6 +670,82 @@ sub pick_conf {
     die $Opts->{conf} . " doesn't exist";
 }
 
+#===================================
+sub serve_and_open_browser {
+#===================================
```
jarpy: I love you too, decomposition.

nik9000: > decomposition

Much :heart: for decomposition.

nik9000:
Thanks for reviewing @jarpy!

PR #661, opened 2019-03-04
Asciidoctor: Make Edit Me links know about repos

nik9000:
The "Edit Me" link generation in Asciidoctor was basically a copy of the
asciidoc implementation which was simple because it *had* to be simple.
The asciidoctor implementation had trouble dealing with books built from
multiple repositories because the asciidoc implementation does. The
asciidoc implementation works around it by allowing files to set their
`edit_url` whenever they include a book from a different repo. We had a
bug the asciidoctor implementation that led to cross repo links not
working that was unrelated, but it led me to notice how silly this whole
thing was.

This change replaces way that asciidoctor finds the edit url. We now
pass in an attribute with the root of each repo and the edit me url for
that root. When we want to generate the a link we pick the first repo
that contains the path of the asciidoctor file and use it's associated
url. This is nice because it dosen't need anything like `repo_root` and
it can automatically handle the edit me links, allowing us to remove the
overridden links when we shift books to asciidoctor.

### DIFF
--- a/resources/asciidoctor/lib/edit_me/extension.rb
+++ b/resources/asciidoctor/lib/edit_me/extension.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require 'pathname'
+require 'csv'
 require_relative '../scaffold.rb'
 
 ##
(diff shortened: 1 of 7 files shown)


nik9000:
I'd love to steal some of the infrastructure in #659 for an integration test for this.

resources/asciidoctor/lib/edit_me/extension.rb
```diff
@@ -12,32 +12,56 @@ class EditMe < TreeProcessorScaffold
...
+    CSV.parse edit_urls_string do |toplevel, url|
+      unless toplevel
+        logger.error message_with_context "invalid edit_urls, no toplevel"
+        next
+      end
+      unless url
+        logger.error message_with_context "invalid edit_urls, no url"
+        next
+      end
+      url = url[0..-2] if url.end_with? '/'
```
jarpy: Since you are mutating `url` anyway, you could:

```
url.chomp!('/')
```

Not a change request, just a "fun fact".

resources/asciidoctor/lib/edit_me/extension.rb
```diff
@@ -12,32 +12,56 @@ class EditMe < TreeProcessorScaffold
...
+      # || '<stdin>' allows us to not blow up when translating strings that
+      # aren't associated with any particular file. '<stdin>' is asciidoctor's
+      # standard name for such strings.
+      path = source_path || '<stdin>'
+
+      edit_urls = @document.attributes['edit_urls']
+      edit_url = edit_urls.find { |e| path.start_with? e[:toplevel] }
+      unless edit_url
+        logger.warn message_with_context "couldn't find edit url for #{path}", :source_location => source_location
+        return super
```
jarpy: Really happy to see all these new log messages.

resources/asciidoctor/spec/edit_me_spec.rb
```diff
@@ -13,17 +14,41 @@
     Asciidoctor::Extensions.unregister_all
   end
 
+  spec_dir = File.dirname(__FILE__)
+
+  it "has a nice error message if you are missing the edit url" do
```
jarpy: I love test names like this. :+1:

Totally slays `test_error_message`.

nik9000: I like this about rspec a lot! "Ruby as a DSL" can have all kinds of problems with error reporting and stuff, but here it is super nice!

nik9000:
Thanks for reviewing @jarpy!

PR #663, opened 2019-03-05
Fail the build if there are any errors 

nik9000:
We used to fail when building a book resulted in any `WARNING`s or the
process the builds the book exits with a non-0 status. This was fine
when we were only using `AsciiDoc` because `a2x` would always exit with
a non-0 status if it output any `ERROR`s. But Asciidoctor will happily
exit with a 0 status after output errors! If that happens, we should
fail to build the book. This change does exactly that. And it adds an
integration test that attempt to build a file with a missing include.
That tests this change.

### DIFF
--- a/integtest/Makefile
+++ b/integtest/Makefile
@@ -5,7 +5,8 @@ MAKEFLAGS += --silent
 check: \
 	style \
 	readme_expected_files readme_same_files \
-	includes_expected_files includes_same_files
+	includes_expected_files includes_same_files \
+	missing_include_fails_asciidoc missing_include_fails_asciidoctor
 
 .PHONY: style
 style: html_diff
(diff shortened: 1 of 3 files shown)


nik9000:
I'm rerunning a `--all` test build locally just to make sure nothing snuck in. It is pretty unlikely because ERRORs usually cause xmllint errors which will fail the build anyway, just with a bad error message. But it is certainly possible.

integtest/Makefile
```diff
@@ -70,10 +71,30 @@ experimental_expected_files: /tmp/experimental_asciidoc
...
 	done
 
+# Build the docs into the target
 define BD=
 /docs_build/build_docs.pl --in_standard_docker --out $@
 endef
 
+# Build the docs into a temporary directory
+define BD_DUMMY=
+/docs_build/build_docs.pl --in_standard_docker --out /tmp/dummy
```
ddillinger: Huh. [TIL](https://www.gnu.org/software/make/manual/html_node/Multi_002dLine.html)

nik9000:

PR #653, opened 2019-02-27
Switches X-Pack Reference to Asciidoctor builds

lcawl:
This PR updates the X-Pack Reference (https://www.elastic.co/guide/en/x-pack/index.html) such that it's built via Asciidoctor.

Related to #606

PR #664, opened 2019-03-05
Docker: Fix building x-pack references

nik9000:
Fixes building the x-pack reference books with `--doc`. They require a
special directory layout that `build_docs.pl` relied on the user to
build manually. `build_docs` builds the docker filesystem agnostic of
the user's filesystem and without this change would stuff all of the
source repos into `/doc` in the docker image.

For example, the `elasticsearch` and `x-pack-elasticsearch` repo expect
to be layed out like this:

```
+ elasticsearch
\ elasticsearch-extra
 + x-pack-elasticsearch
```

but docker was putting them in `/docs/elasticsearch` and
`/docs/x-pack-elasticsearch`.

This changes how they are mounted in the docker image so they'll build.
Now all repos with names like `x-pack-foo` are mounted in `foo-extra`.
This isn't the cleanest thing in the world, but I expect we can revert
it once we no longer have to support building the docs outside of
docker.

Relates to #653.

### DIFF
--- a/build_docs
+++ b/build_docs
@@ -93,6 +93,20 @@ def run_build_docs(args):
                 cwd=repo_search_path)
         repo_root = repo_root.decode('utf-8').strip()
         repo_name = basename(repo_root)
+        if 'x-pack-' in repo_name:
+            # x-pack-foo repositories expect to be mounted in the foo-extra
+            # directory so we should oblige them so they'll build. This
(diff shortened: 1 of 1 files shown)


nik9000:
> :scream:

It wasn't as weird as it sounds because to hack on the repo you had to check them out that way too.....

PR #666, opened 2019-03-05
Switch resiliency status to asciidoctor

nik9000:
Switches the Elasticsearch resiliency status page to asciidoctor. It
doesn't look 100% the same but the differences are minor and look ok to
me.


nik9000:
Thanks for looking this over with me @lcawl!

PR #667, opened 2019-03-05
Change nginx layout of --open

nik9000:
When you run a build with `--open` we start a web server locally and
open up a browser window for you to look at them. With the docker-based
`build_docs` that server is nginx. nginx is fancy, and one of the things
that we can do pretty easily is lay out the files that we serve so they
look fairly like the public web site. This change does just that and it
should come in handy to test future work we plan to do that generates an
nginx config to deploy to the site serving the docs.

### DIFF
--- a/build_docs
+++ b/build_docs
@@ -234,7 +234,8 @@ def run_build_docs(args):
                 # is weird. If we tell it that the browser's name is `open`
                 # we'll get whatever the user set as their default browser.
                 environ['BROWSER'] = 'open'
-            webbrowser.open('http://localhost:8000/guide', new=1, autoraise=False)
+            webbrowser.open('http://localhost:8000/guide',
+                            new=1, autoraise=False)
         if should_forward_ssh_auth_into_container:
             match = re.match('Waiting for ssh auth to be forwarded to (.+)',
                              line)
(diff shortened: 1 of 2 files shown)


build_docs
```diff
@@ -234,7 +234,7 @@ def run_build_docs(args):
                 # is weird. If we tell it that the browser's name is `open`
                 # we'll get whatever the user set as their default browser.
                 environ['BROWSER'] = 'open'
-            webbrowser.open('http://localhost:8000', new=1, autoraise=False)
+            webbrowser.open('http://localhost:8000/guide', new=1, autoraise=False)
```
nik9000: This is the new path. It mirrors the public site nicely when you do `./build_docs --all --open`

build_docs.pl
```diff
@@ -710,7 +710,7 @@ sub serve_and_open_browser {
         if ( not $running_in_standard_docker ) {
             sleep 1;
             say "Press Ctrl-C to exit the web server";
-            open_browser("http://localhost:8000/$open_path");
+            open_browser("http://localhost:8000/");
```
nik9000: This is always the path when you build the docs with `./build_docs.pl` which doesn't support `--all`.

nik9000:
Thanks for reviewing @jarpy!

PR #674, opened 2019-03-06
Add redirects to built docs

nik9000:
This causes us to build a list of redirects in the built-docs repo which
we can then pick up on the public web server. For now, this list is just
copied from the public web server and munged slightly to make it
testable locally.

As a side advantage, when you run `./build_docs --all --open` it'll
include the same redirects that we have on the public web server!


nik9000:
I'll add a simple integration test for this then remove the WIP!

nik9000:
OOH! Self test failure!

nik9000:
OK! This has what feels to me like the minimum testing. I think that is enough for playing with this in staging. I'd like to add another test with `--open` that actually hits a redirect before we cut over the public site, but I think it'd be useful to get this in for testing with staging.

brianjolly:
This PR looks good to me!

build_docs.pl
```diff
@@ -125,7 +127,7 @@ sub build_local {
 
     if ( $Opts->{open} ) {
         say "Opening: " . $html;
-        serve_and_open_browser( $dir );
+        serve_and_open_browser( $dir, 0 );
     }
```
jarpy: The second argument is `$redirects_file`. Isn't that a string? What strange autocast magic is happening here?

build_docs.pl
```diff
@@ -236,6 +239,9 @@ sub build_all {
             write_html_redirect( $build_dir->subdir( $_->{prefix} ),
                     $_->{redirect} );
         }
+
+        say "Writing nginx redirects";
+        write_nginx_redirects( $redirects, $build_dir, $temp_dir );
```
jarpy: What is your rule for whitespace inside function call parens?

Is it "extra space for dyadic and above function calls"?

nik9000: I has a somewhat shameful confession: I mostly don't care. Like, the only time I really want there to be some particular kind of whitespace is when I've got named parameters. I tend to make the whitespace look like the rest of the project around it and since this project sometimes uses `( $foo )` and sometimes uses `($foo)` I end up mixed up. Especially if I copying something. There is something to say for being consistent though. I think more of the project uses `( $foo )` I should probably stick with that.

build_docs.pl
nik9000: Think of `0` as Python's `None` or Ruby's `nil`. `$redirects_file` is "anything that's string representation looks like a file or anything falsy 0 to skip the file." Let me document the parameters.

nik9000:
@jarpy, I believe this is ready for another round whenever you are ready for it!

nik9000:
And thanks for reviewing it!

build_docs.pl
jarpy: Thanks for replying. Not a huge deal to me either, to be frank, so sorry if it sounded picky.

nik9000: Note at all!

build_docs.pl
jarpy: Ah cool. It's been a looong time since I wrote much Perl. If that's a common falsy/nulley value in Perl then I'm good.

jarpy:
Thanks for the updates and clarifications. Jarpy is rusty in Perl and appreciates it!

nik9000:
> Thanks for the updates and clarifications. Jarpy is rusty in Perl and appreciates it!

Thanks for the reviews! Especially on Saturday morning!

PR #678, opened 2019-03-07
Drop .DS_Store

nik9000:
macOS makes this thing and sometimes modifies it. That gets in the way of
the normal business of tracking changes. This removes the file and
stops git from tracking it.


nik9000:
> If I was using a Mac, I would personally just add this to [my global ignore list](https://github.com/jarpy/puppet-desktop/blob/92512d4225b4ee7b05a494c3771c8d4afac0d587/modules/git/templates/gitconfig.erb#L7).

OOOOOH! I peak into the secret heart of Jarpy. And see puppet!

nik9000:
> And see puppet!

Mine is [shell](https://github.com/nik9000/recreate)

PR #680, opened 2019-03-07
Asciidoctor: Handle snippets in definition lists

nik9000:
Our lang override processor had trouble with snippets inside of
definition lists because a method provided by Asciidoctor doesn't quite
behave well inside definition lists:
https://github.com/asciidoctor/asciidoctor/issues/3133

That method had behavior that is a little too broad anyway so we
implement what we need directly.


nik9000:
@estolfo could you have a look at this one? It is 100% ruby and small enough that I think we can iterate on some best practices things here. This in particular makes me like the whole "early return" thing so we don't end up with a ton of indentation.

I know this has issues with the magic perl variables, but I think it'd be best to clean those up separately.

resources/asciidoctor/lib/elastic_compat_tree_processor/extension.rb
```diff
@@ -67,7 +67,13 @@ def process_subs(block)
   }.freeze
 
   def process_lang_override(block)
-    next_block = block.next_adjacent_block
+    # Check if the next block is a marker for the language
+    # We don't want block.next_adjacent_block because that'll go "too far"
+    # and it has trouble with definition lists.
+    next_block_index = block.parent.blocks.find_index block
+    return if next_block_index.nil?
```
estolfo: I believe it would be more idiomatic to write `return unless next_block_index`

resources/asciidoctor/spec/elastic_compat_tree_processor_spec.rb
```diff
@@ -96,4 +96,20 @@
       expect(actual).to eq(expected.strip)
     end
   end
+
+  it "doesn't mind when the snippet inside a definition list" do
+    actual = convert <<~ASCIIDOC
```
estolfo: I think I mentioned this in the last PR I reviewed, maybe eventually you can use `context` and `let` blocks instead in the specs.
For example, this case would have a context `when the snippet is inside a definition list` and then the `it` block would be `generates the correct string` (or whatever). The `actual` and `expected` variables would be `let` blocks.
But I think that change would be part of a larger rspec refactor and perhaps out of the scope of this PR.

[This](https://github.com/elastic/elasticsearch-ruby/blob/6.x/elasticsearch-api/spec/elasticsearch/api/utils_spec.rb#L89-L130) is an example of how context blocks and let blocks can be used.

resources/asciidoctor/lib/elastic_compat_tree_processor/extension.rb
nik9000: When I wrote it I felt like "This can only be nil or a string so I should be explicit" but I've come around to your way.

nik9000:
@estolfo, I pushed a change for the `nil?` thing and a change to use `context`, `let`, and `subject` for that section. If it looks right to you I can do the other sections as I get there.

nik9000:
@estolfo, I pushed some changes to the rspec to line up a bit better with what you proposed.

PR #682, opened 2019-03-07
Asciidoctor: Make some regexes constants

nik9000:
Its been recommended to me that I should make all of my regexes constants
if possible. This grabs a few of them.


nik9000:
@estolfo, this is a small step in the kind of clean up that we've talked about. Thanks for pointing these out to me!

nik9000:

PR #685, opened 2019-03-07
Switch the perl client docs to asciidoctor

nik9000:
Switches the perl client docs to asciidoctor. `html_diff` says
asciidoctor produces exactly the same output modulo new lines.

### DIFF
--- a/doc_build_aliases.sh
+++ b/doc_build_aliases.sh
@@ -120,7 +120,7 @@ alias docbldnet='$GIT_HOME/docs/build_docs.pl --doc $GIT_HOME/elasticsearch-net/
 
 alias docbldphp='$GIT_HOME/docs/build_docs.pl --doc $GIT_HOME/elasticsearch-php/docs/index.asciidoc'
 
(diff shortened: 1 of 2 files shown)


doc_build_aliases.sh
```diff
@@ -120,7 +120,7 @@ alias docbldnet='$GIT_HOME/docs/build_docs.pl --doc $GIT_HOME/elasticsearch-net/
 
 alias docbldphp='$GIT_HOME/docs/build_docs.pl --doc $GIT_HOME/elasticsearch-php/docs/index.asciidoc'
 
-alias docbldepl='$GIT_HOME/docs/build_docs.pl --doc $GIT_HOME/elasticsearch/docs/perl/index.asciidoc --single 1'
+alias docbldepl='$GIT_HOME/docs/build_docs --asciidoctor --doc $GIT_HOME/elasticsearch/docs/perl/index.asciidoc --single'
```
nik9000: The `1` doesn't actually work with `build_docs`. Since no one has complained I'm going to assume that is ok and just drop it. The `1` didn't do anything - just specifying `--single` does the trick.

nik9000:
Thanks for reviewing @lcawl!

PR #689, opened 2019-03-08
Commit template instead of fetching it

### LINKED ISSUE #679: Allow local doc build without internet connection
The current doc build requires an internet connection, but sometimes we want to build the docs while we're offline (on a plane, on a train, in a boat).

nik9000:
This stops `build_docs` from fetching the html template used to spruce
up the docs built by docbook to look more like the rest of Elastic's
site. Instead we commit the template directly to the repository and add
a simple shell script to fetch it. I've extracted a promise from the
folks that maintain the template that they'll either open a PR to update
the template or file an issue about it.

This drops support for `--staging` which builds the docs against the
template from Elastic's staging site. I do this because we're honestly
not sure how we'd like to use it and it hasnt worked in CI ever. We'll
work something out when we know what we have to do with staging.

Closes #679


lcawl:
We used the --staging to have a way to preview changes to the docs while we were "frozen" before big releases.  Do we just use a different --target_repo value now instead?

nik9000:
I'm fairly sure I broke that behavior of `--staging` when I switched to deploying from the different repo. I *think* the best way to get that sort of thing back is to let you do something like `--target_branch staging`. I'll file an issue for that.

What this change drops is that `./build_docs --staging` pulls the template for the staging site.

dedemorton:
LGTM. The offline build now works (yay!), but of course you need a connection to see the docs properly formatted.

nik9000:
> but of course you need a connection to see the docs properly formatted.

Sadly, yes. I wonder about that though..... A thing for later I think.

dedemorton:
Being able to build to docs offline is a 1000% improvement, IMO, with or without proper formatting, so *thank you* for doing this.

build_docs.pl
```diff
@@ -55,7 +55,7 @@ BEGIN
     $Opts,    #
     'all', 'push', 'target_repo=s', 'reference=s', 'rebuild', 'no_fetch', #
     'single',  'pdf',     'doc=s',           'out=s',  'toc', 'chunk=i', 'suppress_migration_warnings',
-    'open',    'skiplinkcheck', 'linkcheckonly', 'staging', 'procs=i',         'user=s', 'lang=s',
+    'open',    'skiplinkcheck', 'linkcheckonly', 'procs=i',         'user=s', 'lang=s',
     'lenient', 'verbose', 'reload_template', 'resource=s@', 'asciidoctor', 'in_standard_docker',
```
jarpy: Not your fault, but you would you mind doing something about this whitespace when you're next in here?

fetch_template.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+# Fetch the production templates for the guide.
+# Run this from the root of the docs repo.
+
+curl https://www.elastic.co/guide_template > resources/web/template.html
```
jarpy: This is invoked manually, right?

resources/web/README
```diff
@@ -0,0 +1 @@
+template.html is fetched by fetch_template.sh
```
jarpy: If this is a manual step, I'd say "_can be fetched"_.

"_is fetched_" implies that it happens automagically.

fetch_template.sh
nik9000: Manually, yes.

resources/web/README
nik9000: ++

build_docs.pl
nik9000: How about I open a PR after this that sorts them? Some of them are "for `--all`" and some are "for `--doc`" and some of them are "common" so I shouldn't sort them blindly. But I think a followup PR is best here.

resources/web/README
nik9000: Done!

build_docs.pl
jarpy: Thanks! Yes. I really meant "if you are bored sometime in the future". <3

PR #691, opened 2019-03-08
Switch ruby client docs to asciidoctor

nik9000:
This switches building the ruby client docs from AsciiDoc to
Asciidoctor. The html that it produces isn't 100% the same, but most the
differences aren't relevant. Here is a diff of the relevant stuff:
https://gist.github.com/nik9000/92e098ca9d9ce113f5d795ee1b0017ce

That diff seems like a net positive.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -385,6 +385,7 @@ contents:
                 current:    master
                 tags:       Clients/Ruby
                 subject:    Clients
+                asciidoctor: true
                 sources:
                   -
                     repo:   elasticsearch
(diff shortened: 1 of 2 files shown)


lcawl:
@estolfo Similar to something we've noticed while migrating the other clients, I noticed that there are no explicit anchors for the pages in the books (i.e. no [[anchor_name]] for each page).  As a result, they're auto-generated, which could result in changed URLs unintentionally.  If you like, I am happy to help add IDs in the source files.

nik9000:
Thanks for reviewing @lcawl!

@estolfo, when you build these docs locally you'll need to do something like:
```
$GIT_HOME/docs/build_docs --asciidoctor --doc $GIT_HOME/elasticsearch/docs/ruby/index.asciidoc
```

Note the lack of `.pl` on the build and the new `--asciidoctor`.

estolfo:
@lcawl thanks. I'll take a look and see how I can structure the ruby client docs better.

PR #692, opened 2019-03-08
Switch the community clients book to Asciidoctor

nik9000:
Switches the community clients book in the Elasticsearch repo to
Asciidoctor. The generated html has some spacing differences but our
tuned html_diff tool doesn't think that there are any differences that
are impactful.


nik9000:
Thanks for reviewing @lcawl!

PR #698, opened 2019-03-11
Switch the Infra Monitoring Guide to asciidoctor

nik9000:
Switches the Infrastructure Monitoring Guide to Asciidoctor. There are a
few spacing changes that are noops in HTML and there are a few changes
that look like this:
```
-            <img alt="monitoring-architecture.png" src="monitoring-architecture.png"/>
+            <img alt="monitoring architecture" src="monitoring-architecture.png"/>
```

They switch the alt tag on an image from the name of the image to a
"more English" version of the name of the image. I'm not sure if this is
really an improvement because the alt tag is still being automatically
generated so it isn't going to be particularly descriptive. But I don't
*think* this is a problem either.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -836,6 +836,7 @@ contents:
             chunk:      1
             tags:       Infrastructure/Guide
             subject:    Infrastructure
+            asciidoctor: true
             sources:
               -
                 repo:   stack-docs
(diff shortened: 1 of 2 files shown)


nik9000:
Thanks for reviewing @dedemorton and @bmorelli25!

PR #702, opened 2019-03-12
Switch the ES Python Client to Asciidoctor

nik9000:
This switches the Elasticsearch Python Client's stub book to
Asciidoctor. `html_diff` claims that the html is the same modulo
whitespace. This book is small because it points folks at the canonical
docs at `readthedocs.org`.


nik9000:
Thanks for reviewing @lcawl!

PR #704, opened 2019-03-12
Asciidoctor: Remove some magic $ variable usage

nik9000:
This makes some headaway into our usage of "magic $ variables" which are
so common in perl but discouraged in ruby. It doesn't remove all of
them, but it removes that ones that are simple to get at and aren't
already being addressed in another ongoing change.

### DIFF
--- a/resources/asciidoctor/lib/copy_images/extension.rb
+++ b/resources/asciidoctor/lib/copy_images/extension.rb
@@ -25,7 +25,7 @@ module CopyImages
       'changed' => 'note',
       'deleted' => 'warning',
     }.freeze
-    CALLOUT_RX = /CO\d+-(\d+)/.freeze
+    CALLOUT_RX = /CO\d+-(\d+)/
 
     def initialize(name)
       super
--- a/resources/asciidoctor/lib/copy_images/extension.rb
+++ b/resources/asciidoctor/lib/copy_images/extension.rb
@@ -25,6 +25,7 @@ module CopyImages
       'changed' => 'note',
       'deleted' => 'warning',
     }.freeze
+    CALLOUT_RX = /CO\d+-(\d+)/.freeze
 
     def initialize(name)
       super
@@ -54,8 +55,8 @@ module CopyImages
       coids = block.attr('coids')
       return unless coids
 
-      coids.scan(/CO(?:\d+)-(\d+)/) do
-        @copier.copy_image block, "images/icons/callouts/#{$1}.#{callout_extension}"
+      coids.scan(CALLOUT_RX) do |(index)|
+        @copier.copy_image block, "images/icons/callouts/#{index}.#{callout_extension}"
       end
     end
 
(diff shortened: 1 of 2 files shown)


resources/asciidoctor/lib/copy_images/extension.rb
```diff
@@ -25,6 +25,7 @@ class CopyImages < TreeProcessorScaffold
       'changed' => 'note',
       'deleted' => 'warning',
     }.freeze
+    CALLOUT_RX = /CO\d+-(\d+)/.freeze
```
estolfo: I don't believe it's necessary to freeze this.

resources/asciidoctor/lib/copy_images/extension.rb
```diff
@@ -54,8 +55,8 @@ def process_callout(block)
       coids = block.attr('coids')
       return unless coids
 
-      coids.scan(/CO(?:\d+)-(\d+)/) do
-        @copier.copy_image block, "images/icons/callouts/#{$1}.#{callout_extension}"
+      coids.scan(CALLOUT_RX) do |(index)|
```
estolfo: There's no need to wrap `index` in`()`

resources/asciidoctor/lib/elastic_compat_preprocessor/extension.rb
```diff
@@ -150,8 +150,9 @@ def reader.process_line(line)
         line = super
         return nil if line.nil?
 
-        if SOURCE_WITH_SUBS_RX =~ line
-          line.sub! "subs=\"#{$1}\"", "subs=\"#{$1},callouts\"" unless $1.include? 'callouts'
+        SOURCE_WITH_SUBS_RX.match(line) do |m|
+          old_subs = m[1]
+          line.sub! "subs=\"#{old_subs}\"", "subs=\"#{old_subs},callouts\"" unless old_subs.include? 'callouts'
```
estolfo: Some people may disagree but I would prefer `()` here for readability. 

```ruby
line.sub!("subs=\"#{old_subs}\"", "subs=\"#{old_subs},callouts\"") unless old_subs.include?('callouts')
```

(For what it's worth, I use `()` everywhere but that's a much larger debate)

Also, what's the reason you're checking whether `old_subs` contains the substring `callouts`?

resources/asciidoctor/lib/copy_images/extension.rb
nik9000: I wrapped it so I could get the index rather than a single element array containing the index.

resources/asciidoctor/lib/elastic_compat_preprocessor/extension.rb
nik9000: > Also, what's the reason you're checking whether `old_subs` contains the substring `callouts`?

I'll leave a comment. This is a difference between Asciidoctor and AsciiDoc. AsciiDoc will add `callouts` to the list automatically. Asciidoctor doesn't. But we have zillions of places where we assume that it *is* added automatically.

resources/asciidoctor/lib/copy_images/extension.rb
estolfo: ah, makes sense.

nik9000:
@estolfo, I've pushed some fixes based on your comments. Please would you have another look?

nik9000:

PR #705, opened 2019-03-12
Switch the APM Agent for .NET to Asciidoctor

nik9000:
Switches the APM Agent for .NET's docs from the no-longer-maintained
AsciiDoc project to Asciidoctor. The only difference between the html
output is spacing which the browser ignores.


nik9000:
Thanks for reviewing @gregkalapos and @bmorelli25!

PR #706, opened 2019-03-12
self-test: remove last temporary workaround

nik9000:
Remove the last "temporary workaround" in `html_diff` for issues
building the readme. I fix the readme by using `ifdef` to write
different source text for AsciiDoc vs Asciidoctor. This would be a bad
solution if these things were common in normal books, but the constructs
that need this are when you have snippets of asciidoc inside the
asciidoc. This is a thing that will probably only happen in the readme.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -997,7 +997,8 @@ footnote to a particular line of code:
 }
 ----------------------------------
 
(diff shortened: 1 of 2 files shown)


nik9000:

PR #707, opened 2019-03-12
Switch the APM Agent for Java to Asciidoctor

nik9000:
Switches the core of the documentation build infrasturcture for the APM
Agent for Java from the no-longer-living AsciiDoc project to the active
Asciidoctor project. The HTML differences are fairly small:
1. Empty table blocks no longer contain an `<p></p>`
2. Headings are now in a `<thead>` tag.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -942,6 +942,7 @@ contents:
                 tags:       APM Java Agent/Reference
                 subject:    APM
                 chunk:      1
+                asciidoctor: true
                 sources:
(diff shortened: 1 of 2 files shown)


nik9000:
[Here](https://gist.github.com/nik9000/0562b30c51c1822edf877e4771335e92) is a gist containing the difference, ignoring things that are invisible when the html is rendered or are a net positive like a few extra zero-width-spaces.

PR #709, opened 2019-03-12
Asciidoctor: Support adoc files

nik9000:
The `adoc` extension is fairly common instead for asciidoc files. Most
of our files use `asciidoc` but a few of them, notable
elasticsearch-hadoop, uses `adoc`. This fixes support for them and adds
a test for it.


nik9000:
Thanks for reviewing @jarpy!

PR #710, opened 2019-03-12
Switch APM Agent for Go to Asciidoctor

nik9000:
Switches the core of the process that builds the docs for the APM Agent
for Go from the no-longer-supports AsciiDoc to the supported
Asciidoctor. The html output is slightly different - mostly spacing
which is generally ignored by the browser. It *does* change the spacing
in front of the go source from a leading eight spaces to a leading tab.
These look the same to me in the browser but copy-and-pasting them comes
out slightly differently.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -930,6 +930,7 @@ contents:
                 tags:       APM Go Agent/Reference
                 subject:    APM
                 chunk:      1
+                asciidoctor: true
                 sources:
                   -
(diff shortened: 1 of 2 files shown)


nik9000:
[Here](https://github.com/elastic/built-docs-nik/commit/77a2e3e35ca3c0c8a3794038223281b099ad8a97) is how the html differs. You can test the change locally using the command in the aliases file.

PR #711, opened 2019-03-12
Build tables of contents with Asciidoctor

nik9000:
We chain the books that make the all of the docs together into a web
site by making tables of content that link the whole thing together.
We make these by building asciidoc files and then building them like
single-page books. This switches from building them with AsciiDoc to
building them with Asciidoctor.


nik9000:
This produces functionally equal html output. It also does it quite a bit faster, though I expect once we get #673 it won't be as much faster.

nik9000:

PR #714, opened 2019-03-13
Switch APM Agent for RUM JavaScript

nik9000:
Switches the docs build core for the APM Agent for RUM JavaScript from
the no-longer-supported AsciiDoc to the actively maintained Asciidoctor.
The generated HTML is the same modulo whitespace which the browser
ignores.


nik9000:
Thanks for reviewing @bmorelli25  and @jahtalab ! When you next build the book please use the new command line options described in the aliases file. Thanks again!

PR #716, opened 2019-03-13
Add shared refs for APM Server 7.0

bmorelli25:
APM Server's `7.x` documentation branch is going away. See https://github.com/elastic/apm-server/issues/1993 for more details. This PR allows us to cross doc link to `7.0` documentation from `7.x`.


lcawl:
Another option might be to add definitions in each book for the appropriate APM version (e.g. kind of like how Logstash books define Elasticsearch and Kibana versions in https://raw.githubusercontent.com/elastic/logstash/master/docs/index.asciidoc).  Then in these URL attributes, you could just reference the {apm-version}.  

I think that would be more work across all the books, however, so unless there are a *lot* of APM links all over the place, your current solution is likely the best answer.

PR #718, opened 2019-03-14
Ruby Style: Treat regexps as immutable

nik9000:
In ruby regular expressions aren't frozen by default *but* they don't have
any methods to mutate them. You *can* still mutate them with standard ruby
monkey patching, but we find it more idiomatic to treat them as immutable
for style purposes.


nik9000:
Thanks @estolfo!

PR #721, opened 2019-03-15
Asciidoctor: BWC for // CONSOLE then callouts

nik9000:
Our backwards compatibility support for "OPEN IN CONSOLE"-style links did
not work with snippets followed by a callout list. That looks like this:

```
[source,js]
----
GET / <1>
----
// CONSOLE
<1> words go here
```

These is actually *super* common and I'm surprised I didn't catch it
before! This adds tests for it so it won't sneak through again.

The reason this didn't work is that our backwards compatibility worked by
transforming the `// CONSOLE` comment in `pass:[// CONSOLE]` and then
catching that when walking the tree. We do this in two passes because our
preprocessor works line by line and reading ahead can cause issues. So we
do the "action at a distance" in a tree processor where it is quite
simple.

The trouble is that the `pass:[// CONSOLE]` construct is *inline* so when
it is immediately followed by the `<1> words go here` line it gets sucked
into the same block which Asciidoctor thinks of as a paragraph. This
causes the parsing for the callout list not to happen, causing all kinds
of trouble.

The fix for this is to introduce a block macro that functions like
`pass:[// CONSOLE]` and to use that instead. This causes the callout list
to be parsed properly.

### DIFF
--- a/resources/asciidoctor/spec/elastic_compat_tree_processor_spec.rb
+++ b/resources/asciidoctor/spec/elastic_compat_tree_processor_spec.rb
@@ -1,11 +1,13 @@
 # frozen_string_literal: true
 
 require 'elastic_compat_tree_processor/extension'
+require 'lang_override/extension'
 
 RSpec.describe ElasticCompatTreeProcessor do
   before(:each) do
     Asciidoctor::Extensions.register do
       treeprocessor ElasticCompatTreeProcessor
+      block_macro LangOverride
     end
   end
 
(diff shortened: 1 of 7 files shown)


resources/asciidoctor/spec/elastic_compat_tree_processor_spec.rb
```diff
@@ -72,51 +74,82 @@
...
+    context "for a snippet with#{name}" do
+      let(:snippet) do
+        snippet = <<~ASCIIDOC
+          [source,js]
+          ----
+          GET / <1>
+          ----
+        ASCIIDOC
+        snippet += "lang_override::[#{override}]" if override
+        snippet
```
estolfo: Have you seen the `#tap` method in Ruby? It's a style thing, but you could use it here. I like using it so I never accidentally forget to return the object (in this case, `snippet`) from the let block.

resources/asciidoctor/spec/elastic_compat_tree_processor_spec.rb
```diff
@@ -72,51 +74,82 @@
...
+        end
+      end
+      context 'when it is alone' do
+        let(:converted) do
+          convert <<~ASCIIDOC
+            == Example
+            #{snippet}
+          ASCIIDOC
+        end
+        include_examples 'has the expected language'
```
estolfo: This is very nuanced, but you could also use `it_behaves_like` here. `include_examples` is best suited for what you're doing from the topic level - creating specs using different values but when you have a nested context and want to emphasize a behavior, people tend to use `it_behaves_like`. They're effectively interchangeable and I think the difference really comes down to how you want the specs to print out. Sometimes it's also awkward to find a phrase that describes the behavior. For example, in this case, it might be strange to say `it_behaves_like 'a string converter'` - or whatever. So anyway, maybe `include_examples` is best here and you can ignore this whole comment = )

nik9000: Thanks for this! I looked at `it_behaves_like` for a bunch of things. I'll play with it some more and see what I can find. I found that it doesn't work well when you want to apply the same set of shared examples multiple times with different parameters. Or, rather, it doesn't seem to properly list the parameters when you run rspec. I'll try it some more though.

resources/asciidoctor/spec/elastic_compat_tree_processor_spec.rb
nik9000: `#tap` seems pretty complex to use here because the string is frozen. Though I bet there are other places i can use it....

nik9000:
OK! I've pushed some more tests for the regex on the way in. I think this is ready for review again!

resources/asciidoctor/spec/elastic_compat_tree_processor_spec.rb
estolfo: ah, then it wouldn't work if the String is frozen..

resources/asciidoctor/spec/elastic_compat_tree_processor_spec.rb
estolfo: I see what you're saying. Ultimately, they can accomplish very similar things in the specs, so it's up to you and how you want the the specs to run and print out.

nik9000:
Thanks for reviewing @lcawl!

nik9000:
> Thanks for reviewing @lcawl!

Wow, that is embarrassing. "I meant thanks for reviewing @estolfo!" I mean, both of you review my code but for different reasons and I was just sending something to @lcawl.... But I'm not sure about why I mixed you two up there. Sorry!

PR #723, opened 2019-03-15
Switch Python APM Agent docs to Asciidoctor

nik9000:
Switches the core of the docs generation pipeline from the
no-longer-maintained AsciiDoc project to the actively-maintained
Asciidoctor project. The HTML that it produces isn't 100% the same but
it looks to only differ in three ways:
1. Spacing that the browser will ignore anyway.
2. AsciiDoc was producing `<p></p>` inside of empty table cells.
Asciidoctor does not.
3. In the 1.x docs there are some attribute references that do not
resolve and they are rendered differently by Asciidoctor. Either way,
they should probably be fixed in the 1.x branch of the docs.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -891,6 +891,7 @@ contents:
                 tags:       APM Python Agent/Reference
                 subject:    APM
                 chunk:      1
+                asciidoctor: true
                 sources:
                   -
                     repo:   apm-agent-python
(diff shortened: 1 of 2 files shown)


nik9000:
Here is the diff: https://gist.github.com/nik9000/9b67d9c8732de19d5db34185181586b8

nik9000:
Thanks @beniwohli !

PR #727, opened 2019-03-15
self-test: Ignore empty paragraphs inside of td

nik9000:
AsciiDoc spits out `<td><p></p></td>` for an empty table cell. Asciidoctor
spits out `<td></td>` which is just fine. This ignores the empty `<p>`
tags that are children of `<td>`.


nik9000:

PR #731, opened 2019-03-16
Fix handling of copy-as-curl for HEAD requests

jasontedor:
curl -X HEAD behaves in an unexpected way; namely, it behaves as a GET request so it sits there waiting for Content-Length bytes to be sent, but since the method was HEAD the server would never send these bytes. This makes curl hang forever. It is unfortunate that common sense has not prevailed so this remains unchanged in curl to this day. Instead, curl has a built-in flag for sending a HEAD request and not expecting Content-Length bytes to sent. This is -I or --head. This commit fixes the copy-as-curl behavior with HEAD requests to use the correct form of sending a HEAD request.

Relates #110
Relates elastic/elasticsearch#40114

### DIFF
--- a/resources/web/docs.js
+++ b/resources/web/docs.js
@@ -313,7 +313,12 @@ jQuery(function() {
           curlText += comment + '\n';
         } else {
           path = path.replace(/^\//, '').replace(/\s+$/,'');
-          curlText += 'curl -X ' + method + ' "' + host + '/' + path + '"';
+          if (method === "HEAD") {
+            curlText += 'curl -I ';
+          } else {
+            curlText += 'curl -X ' + method + ' ';
+          }
+          curlText += '"' + host + '/' + path + '"';
 
           if (div.data('kibana')) {
             curlText += " -H 'kbn-xsrf: true'";


nik9000:
Lgtm!

jasontedor:
Thanks for looking @nik9000!

nik9000:
Thanks for fixing it!

jasontedor:
@nik9000 Does something special need to be done to deploy this? I assumed it would happen on the next doc rebuild but while one has occurred this doesn't appear deployed?

nik9000:
> @nik9000 Does something special need to be done to deploy this? I assumed it would happen on the next doc rebuild but while one has occurred this doesn't appear deployed?

Checking.....

nik9000:
It looks like we copy this js file into every book. That means we'd need to rebuild all of the books to deploy this. We force this when the docs.js file changes because, well, we never built that. But I kicked off a [build](https://elasticsearch-ci.elastic.co/job/elastic+docs+master+build+nik/3219/console) that rebuilds all the documents because I told it to.

jasontedor:
Thanks @nik9000. I verified this is deployed now.

PR #732, opened 2019-03-18
Sort command line options

nik9000:
This tidies the list of command line options so they are always in
alphabetical order within their three groups.

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -50,15 +50,8 @@ use ES::Toc();
 use ES::LinkCheck();
 use ES::Template();
 
-GetOptions(
-    $Opts,    #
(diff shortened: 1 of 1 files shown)


build_docs.pl
```diff
@@ -680,32 +673,6 @@ sub init_env {
         or die "Please install <xsltproc>";
 }
 
-#===================================
```
nik9000: I moved this so it'd be by its buddies.

nik9000: And I sorted it.

jarpy: :+1: Command-line options should stick together in these troubled times.

nik9000:
Thanks for reviewing @jarpy!

PR #736, opened 2019-03-20
Asciidoctor: Copy images for inline image macro

### LINKED ISSUE #730: Copy images for inlinemediaobject
Related to https://github.com/elastic/docs/pull/541

If I use an inline image, it's not being copied to the output directory (thus resulting in a broken image). 

For example, if I have a block image followed by an inline image:

>image::xpack-tag.png[X-Pack]

>image:test-star.png[Star]

It generates:

`<div class="informalfigure"><div class="mediaobject">X-Pack</div></div><p><span class="inlinemediaobject">Star</span></p>`

But only the first file is copied to the output directory:

image

nik9000:
This copies images declared "inline". Asciidoctor doesn't give a good
way of hooking into these images so we have to scan for them when we
walk the tree. It isn't the best solution, but at least we don't have to
scan for them after the document is rendered! If we did that we wouldn't
get line number information when we can't find images....

Closes #730

### DIFF
--- a/resources/asciidoctor/lib/copy_images/extension.rb
+++ b/resources/asciidoctor/lib/copy_images/extension.rb
@@ -69,7 +69,7 @@ module CopyImages
     end
 
     def process_image(block, uri)
-      return if uri == ''
+      return unless uri
       return if Asciidoctor::Helpers.uriish? uri # Skip external images
 
       @copier.copy_image block, uri
--- a/resources/asciidoctor/lib/copy_images/extension.rb
+++ b/resources/asciidoctor/lib/copy_images/extension.rb
@@ -26,6 +26,7 @@ module CopyImages
       'deleted' => 'warning',
     }.freeze
     CALLOUT_RX = /CO\d+-(\d+)/
+    INLINE_IMAGE_RX = /(\\)?image:([^:\s\[](?:[^\n\[]*[^\s\[])?)\[/m
 
     def initialize(name)
       super
@@ -33,15 +34,42 @@ module CopyImages
     end
 
     def process_block(block)
-      process_image block
+      process_inline_image block
+      process_block_image block
       process_callout block
       process_admonition block
     end
 
-    def process_image(block)
+    def process_block_image(block)
       return unless block.context == :image
 
       uri = block.image_uri(block.attr 'target')
+      process_image block, uri
+    end
+
+    def process_inline_image(block)
+      return unless block.content_model == :simple
+
+      # One day Asciidoc will parse inline things into the AST and we can
+      # get at them nicely. Today, we have to scrape them from the source
+      # of the node.
+      block.source.scan(INLINE_IMAGE_RX) do |(escape, target)|
+        next if escape
+
+        # We have to resolve attributes inside the target. But there is a
+        # "funny" ritual for that because attribute substitution is always
+        # against the document. We have to play the block's attributes against
+        # the document, then clear them on the way out.
+        block.document.playback_attributes block.attributes
+        target = block.sub_attributes target
+        block.document.clear_playback_attributes block.attributes
+        uri = block.image_uri target
+        process_image block, uri
+      end
+    end
+
+    def process_image(block, uri)
+      return if uri == ''
       return if Asciidoctor::Helpers.uriish? uri # Skip external images
 
       @copier.copy_image block, uri
--- a/resources/asciidoctor/spec/copy_images_spec.rb
+++ b/resources/asciidoctor/spec/copy_images_spec.rb
@@ -6,7 +6,7 @@ require 'copy_images/extension'
 require 'fileutils'
 require 'tmpdir'
 
-RSpec.describe CopyImages::CopyImages do
+RSpec.describe CopyImages do
   RSpec::Matchers.define_negated_matcher :not_match, :match
 
   before(:each) do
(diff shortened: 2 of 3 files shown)


resources/asciidoctor/spec/spec_helper.rb
```diff
@@ -36,3 +36,24 @@ def convert(input, extra_attributes = {}, warnings_matcher = eq(''))
   expect(warnings_string).to warnings_matcher
   result
 end
+
+##
+# Convert an asciidoc string into docbook returning the result and any warnings.
+def convert_with_logs(input, extra_attributes = {})
```
nik9000: This seems like a much more rspec friendly way to check for logs. Though I still like having two flavors - one that automatically asserts that there weren't any logs so we don't have to have `it didn't log anything` assertions all over the place....

resources/asciidoctor/spec/copy_images_spec.rb
```diff
@@ -31,46 +31,94 @@ def copy_attributes(copied)
...
-        eq("INFO: <stdin>: line 2: copying #{spec_dir}/resources/copy_images/example1.png")
-    expect(copied).to eq([
-        ["copy_images/example1.png", "#{spec_dir}/resources/copy_images/example1.png"],
-    ])
+  context 'for the image block macro' do
+    let(:image_command) { "image::#{image_ref}[]" }
+    include_examples 'copies images'
+  end
+  context 'for the image inline macro' do
+    # NOCMOMIT rename image_ref to target
```
nik9000: Woops. Looks like eI need to pick up these nocommits

resources/asciidoctor/lib/copy_images/extension.rb
```diff
@@ -26,22 +26,50 @@ class CopyImages < TreeProcessorScaffold
...
+        block.document.playback_attributes block.attributes
+        target = block.sub_attributes target
+        block.document.clear_playback_attributes block.attributes
+        uri = block.image_uri target
+        process_image block, uri
+      end
+    end
+
+    def process_image(block, uri)
+      return if uri == ''
```
estolfo: is there any chance `uri` could be nil?

resources/asciidoctor/spec/copy_images_spec.rb
```diff
@@ -6,7 +6,7 @@
 require 'fileutils'
 require 'tmpdir'
 
-RSpec.describe CopyImages::CopyImages do
+RSpec.describe CopyImages do
```
estolfo: you can also just write `describe` without calling it on `Rspec`

resources/asciidoctor/spec/copy_images_spec.rb
```diff
@@ -31,46 +31,115 @@ def copy_attributes(copied)
...
-    ])
+  shared_examples 'copies the image' do
+    include_context 'convert'
+    it 'copies the image' do
+      expect(copied).to eq([
+          [resolved, "#{spec_dir}/resources/copy_images/example1.png"],
+      ])
+    end
+    it 'logs that it copied the image' do
+      expect(logs).to eq("INFO: <stdin>: line #{include_line}: copying #{spec_dir}/resources/copy_images/example1.png")
```
estolfo: The expected result is a bit long, so perhaps it can be put in a `let` block so that it's easier to parse. Also, if the expected result changes, it's easier to locate in the spec file what needs to edited.
For example:

```ruby

let(:expected_copied_image) do
  [[resolved, "#{spec_dir}/resources/copy_images/example1.png"],]
end

it 'logs that it copied the image' do
  expect(logs).to eq(expected_copied_image)
end
```

resources/asciidoctor/spec/spec_helper.rb
estolfo: Maybe you could add an expectation using `allow` or `expect`. That way, you won't get an assertion `it didn't log anything`

resources/asciidoctor/spec/copy_images_spec.rb
```diff
@@ -31,46 +31,115 @@ def copy_attributes(copied)
...
+  shared_context 'convert' do
+    let(:result) do
+      copied = []
+      attributes = copy_attributes copied
+      result = convert_with_logs input, attributes
+      result.push copied
+      result
+    end
+    let(:logs)   { result[1] }
+    let(:copied) { result[2] }
```
estolfo: I find this to be a little confusing. I think the variables you're interested in can be extracted out into their own `let` blocks and then the `!` can be used to ensure that the result is created before other variables are accessed. I'm not sure if this would work, but I'm thinking something like this that isolates the variables.

```ruby
let!(:logs) do
  attributes = copy_attributes copied
  (convert_with_logs input, attributes)[1]
end

let(:copied) do
  []
end
```

Is the `copied` object mutated by the `copy_attributes` method?

resources/asciidoctor/spec/spec_helper.rb
nik9000: Oooh! Let me do some research there.

resources/asciidoctor/spec/copy_images_spec.rb
nik9000: That causes an undefined method error for me.

resources/asciidoctor/lib/copy_images/extension.rb
nik9000: Not really, but I'll do `return unless uri` just because that feels a little more "ruby" to me.

nik9000:
@estolfo, I pushed something to switch from the tuple-returning method to a shared context that can be included that defines the things that I might assert on.

I haven't done anything about logging so if there aren't any logging assertions then logs are ignored, but that seems fine for this change. I can handle it in a follow up I think.

resources/asciidoctor/spec/copy_images_spec.rb
estolfo: hmm, maybe your version of `RSpec` is older
Anyway, it doesn't really matter how you call `describe`

resources/asciidoctor/spec/spec_helper.rb
```diff
@@ -36,3 +36,36 @@ def convert(input, extra_attributes = {}, warnings_matcher = eq(''))
...
+        :doctype    => :book,
+        :attributes => attributes,
+        :sourcemap  => true
+  end
+  let(:logs) do
+    convert_logger.messages
+      .map { |l| "#{l[:severity]}: #{l[:message].inspect}" }
+      .join("\n")
+  end
+end
```
estolfo: yes! This looks good and more rspec-y

nik9000:

PR #741, opened 2019-03-21
Fix sharing docs built with firebase

nik9000:
Sometimes we want to share the docs that we've built with firebase or a
similar too. I broke that a while back when I turns off `abs_url` when
you build with `--doc` inside docker because I *thought* `abs_url` was
only there to support the built in python web server that we use with
`--open` outside of docker. But it also supports sharing docs with
firebase.

Eventually we'd like something more automatic than firebase but firebase
is what we have right now....

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -66,7 +66,10 @@ init_env();
 
 $Opts->{template} = ES::Template->new(
     %{ $Conf->{template} },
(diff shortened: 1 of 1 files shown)


lcawl:
Thanks, this works in my local tests.

nik9000:
Thanks for looking at it @lcawl!

PR #744, opened 2019-03-22
Make x-pack tags clickable

### LINKED ISSUE #720: Add link to subscriptions page via post-processing
The X-Pack icon near page or section title is currently implemented as described in https://github.com/elastic/docs/issues/542

We would like to turn it into a link to https://www.elastic.co/subscriptions

I had intended to accomplish this via the following Asciidoctor functionality: https://github.com/elastic/docs/pull/567 but the last time I tested it, there were problems: https://github.com/elastic/docs/issues/505#issuecomment-459163236

If we can accomplish this temporarily (or permanently) via post-processing instead, that's worth considering.

nik9000:
This replaces the `x-pack` tags on the titles of pages that are about
elastic licensed features with clickable links that take you to the
subscriptions page.

This is implemented by replacing the section rendering code in docbook
with customized code that includes the link.

This also replaces the background image of the `x-pack` tag with one
that looks "more clickable" and has a hover treatment.

Closes #720


nik9000:
It used to look like:
old

Now it looks like when you don't hover:
new

And like this when you do hover:
new_hover

nik9000:
I'm leaving this in "draft" state while I write a test for it.

nik9000:
Ok - this works for *sections* that have role=xpack but it won't work for chapters with it. I'll see if we have any of those....

resources/web/styles.css
```diff
@@ -130,14 +99,6 @@
   font-weight: inherit;
 }
 
-/* Add xpack icon to floating role=xpack titles in On This Page
```
nik9000: These weren't used.

resources/web/styles.css
```diff
@@ -28,51 +28,20 @@
...
-  margin-left: inherit;
-  margin-right: inherit;
-  font-weight: inherit;
-}
-
-/* Add xpack icon to role=xpack sidebar titles */
-#guide div[class~="xpack"] > div > div > div > p > strong:after,
-
-/* Add xpack icon to role=xpack dt titles */
-#guide dt > span > span.xpack:after,
```
nik9000: Do we have any definitions that are just x-pack? If so I'll need to do something for them.

nik9000:
OK! I've done the chapters too.

resources/web/styles.css
lcawl: I didn't find any in the stack-docs, kibana, or elasticsearch repos

lcawl:
When I test this with the docbldso (Stack Overview) doc build, it fails to create an icon for this page:
https://github.com/elastic/stack-docs/edit/master/docs/en/stack/security/index.asciidoc (https://www.elastic.co/guide/en/elastic-stack-overview/master/elasticsearch-security.html)

nik9000:
> When I test this with the docbldso (Stack Overview) doc build, it fails to create an icon for this page:
> https://github.com/elastic/stack-docs/edit/master/docs/en/stack/security/index.asciidoc (https://www.elastic.co/guide/en/elastic-stack-overview/master/elasticsearch-security.html)
> Ditto for all the other pages at that level (that I tested).

Ooooh! I'll have a look at that then!

nik9000:
> Ooooh! I'll have a look at that then!

Done! @lcawl, could you have another look around?

lcawl:
The icons are appearing in the right places now, thanks!
I noticed, however, that they don't take me to a valid URL: file:///subscriptions
Is this something that's a known issue in local builds and will work when it's on the web?

nik9000:
>  I noticed, however, that they don't take me to a valid URL: file:///subscriptions

I send you to `/subscriptions` because I figured that it'd be nicer to make a relative link. I can point them to `https://www.elastic.co/subscriptions` which'll work locally. I tend to try and avoid these sorts of links because they can lead you from a test environment to a production environment without you knowing it.

lcawl:
It seems to be encountering some problems in the case of sub-sections too:

image

The right hand window is the current (desired) formatting (https://www.elastic.co/guide/en/kibana/master/working-remote-clusters.html#managing-cross-cluster-replication) and the left hand window is built with this PR. Note the lack of icon and the reduced size of the title.

nik9000:
I hit that kind of thing over and over and over again while working on it. Let me see about this instance of it....

nik9000:
Another thing that I notice is that I add the tag even when the section "above" it also has the flag. This seems wrong but something I can work out.

nik9000:
@lcawl could you have another scan on this one when you have a chance? I've gotten *further* I think. And we'll end up with a nice set of tests for when we expect this and when we don't....

PR #746, opened 2019-03-25
Switch legacy sense documentation to Asciidoctor

nik9000:
Switches the core of the docs build for the legacy sense application's
documentation from the unmaintained AsciiDoc to the actively developed
Asciidoctor. The html produced is different only by spacing.


nik9000:
There is no shell alias for this documentation so I skipped that step.

nik9000:
The HTML diff for this one is empty so I'll just merge it so we can keep moving.

PR #748, opened 2019-03-25
Explicit error message if index missing

nik9000:
Adds an explicit error message if the "index" file is missing because
asciidoctor doesn't output anything in that case. It'll fail, but it
won't tell you *why* it failed.

### DIFF
--- a/integtest/Makefile
+++ b/integtest/Makefile
@@ -27,7 +27,8 @@ check: \
 	sub_dir \
 	keep_hash_and_sub_dir \
 	multi_branch \
-	open_all
+	open_all \
+	missing_index
 
 .PHONY: style
 style: html_diff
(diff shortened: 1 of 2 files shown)


integtest/Makefile
```diff
@@ -342,6 +343,24 @@ open_all:
...
+	sed -e 's|--tmp--|$(TMP)|' \
+		-e 's|path:   index.asciidoc|path:   docs|' \
+		-e 's|index:      index.asciidoc|index:      foo/index.asciidoc|' \
+		small_conf.yaml > $(TMP)/conf.yaml
+	set +e; /docs_build/build_docs.pl --in_standard_docker --all \
+		--target_repo $(TMP)/dest.git \
+		--conf $(TMP)/conf.yaml > $(TMP)/out 2>&1; \
+		\
+		test $$? -eq 2
+	$(call GREP,'Can'"'"'t find index',$(TMP)/out)
```
ddillinger: That quoting, whoof

nik9000: Bashquotes.

Issue #757, opened 2019-03-27
Rewrite integtests in rspec


nik9000:
We've outgrown `make` for the `integtest`s. We already have `rspec` in the docker image we use to build the docs so lets use it instead!

nik9000:
Done but I never linked the PRs back to this. Ooops.

PR #758, opened 2019-03-27
Refine --keep_hash behavior for new things

### LINKED ISSUE #749: --keep_hash should just skip new books
We don't want docs tests PRs to fail whenever we add a new book. Like the failure I kicked off [here](https://github.com/elastic/apm-agent-rum-js/pull/227#issuecomment-476375885).

nik9000:
Now that we're using a combination of `--keep_hash --sub_dir` as pull
request tests we've hit some snags, namely that when we add new books or
new repos to existing books `--keep_hash` complained about those books.
This makes it so that new repos and new books are skipped if you use
`--keep_hash`. It also adds test cases for adding new repos to existing
books and adding new books *without* `--keep_hash`.

Note that this work spawned a fierce desire to replace `make` with `rspec`
that I'm tracking in #757. I think I'd like to add these tests and then have
a look at switching from `make` to `rspec`.

Closes #749

### DIFF
--- a/lib/ES/BranchTracker.pm
+++ b/lib/ES/BranchTracker.pm
@@ -7,6 +7,7 @@ use v5.10;
 use Path::Class();
 use ES::Util qw(run sha_for);
 use YAML qw(Dump Load);
+use Storable qw(dclone);
 
 my %Repos;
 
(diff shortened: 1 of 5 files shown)


lib/ES/BranchTracker.pm
```diff
@@ -75,13 +76,21 @@ sub delete_branch {
 sub write {
 #===================================
     my $self = shift;
-    my $new  = Dump( $self->shas );
+    my $to_save = dclone( $self->shas );
+    # Empty hashes are caused by new repos that are unused which shouldn't
+    # force a commit.
+    for my $repo ( keys %{ $to_save } ) {
+        unless ( keys %{ $to_save->{$repo} } ) {
+            delete $to_save->{$repo};
```
ddillinger: My (admittedly ancient) belief is that you _really_ want to use `each` to express the "deleting from a hash I am currently iterating" pattern.

https://perldoc.perl.org/functions/each.html

The relevant bit:

> If you add or delete a hash's elements while iterating over it, the effect on the iterator is unspecified; for example, entries may be skipped or duplicated--so don't do that. Exception: It is always safe to delete the item most recently returned by each, so the following code works properly:

```perl
   while (my ($key, $value) = each %hash) {
        print $key, "\n";
        delete $hash{$key};   # This is safe
    }
```

nik9000:
> Left a note about a particular perl gotcha I found buried in decades old neurons.

Thanks! I read your link and switched.

nik9000:

PR #759, opened 2019-03-28
[DOCS] Copy glossary content from Stack Docs glossary

jrodewig:
This PR copies content from the [Stack Docs glossary](https://github.com/elastic/stack-docs/blob/master/docs/en/glossary/glossary.asciidoc).

It also adds a few Elasticsearch definitions missing from the original Stack Docs glossary:
- [Cross-cluster search](https://github.com/elastic/elasticsearch/blame/master/docs/reference/glossary.asciidoc#L36)
- [Filter](https://github.com/elastic/elasticsearch/blame/master/docs/reference/glossary.asciidoc#L69)
- [Query](https://github.com/elastic/elasticsearch/blame/master/docs/reference/glossary.asciidoc#L140)

#### Why make this change?
The end goal is to use this file as a single source for both the Elasticsearch and Stack Docs glossaries.

After this PR is merged, I'll open separate PRs to:
1. Update the [Elasticsearch glossary](https://github.com/elastic/elasticsearch/blob/master/docs/reference/glossary.asciidoc) to include terms from this file.
2. Update the [Stack Docs glossary](https://github.com/elastic/stack-docs/blob/master/docs/en/glossary/glossary.asciidoc) to include terms from this file.

See elastic/elasticsearch#40560 for more discussion.

### DIFF
--- a/shared/glossary.asciidoc
+++ b/shared/glossary.asciidoc
@@ -1,0 +1,796 @@
+[[terms]]
+== Terminology
+
+ifdef::logstash-terms[]
+
+[[glossary-metadata]] @metadata ::
+
+A special field for storing content that you don't want to include in output
+<<glossary-event,events>>. For example, the `@metadata` field is useful for
+creating transient fields for use in <<glossary-conditional,conditional>>
+statements.
++
+//Source: Logstash
+endif::logstash-terms[]
+ifdef::cloud-terms[]
+
+[[glossary-admin-console]] administration console ::
+
+A component of {ece} that provides the API server for the
+<<glossary-cloud-ui,Cloud UI>>. Also syncs cluster and allocator data from
+ZooKeeper to {es}.
++
+//Source: Cloud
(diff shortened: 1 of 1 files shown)


nik9000:
I'm not a fan of putting this in the docs repo because we don't have proper up-to-date checking on it. Also, it feels like *content* instead of helpers which feels weird to have in the docs repo. I admit I've not really thought out my last point, but the up-to-date checking still makes this a less good idea than it looks on the face of it.

jrodewig:
@nik9000 I think up-to-date checking is a big and valid concern. Added the `team-discuss` label so we can talk about this more before moving forward.

I'm happy to close this PR. At the very least, it revealed some definitions that weren't synced which I can fix in a separate PR if we go another route.

jrodewig:
We discussed this internally and decide to go with a many-to-one approach (individual solution repos feeding into a combine glossary).

Closing this PR.

PR #762, opened 2019-03-29
Be a little more ruby-standard

nik9000:
This switches the way we fetch dependencies for ruby from `gem` to
`bundler` which is more standard in the ruby world. This is basically a
noop unless you are doing ruby development. But if you *are* doing ruby
development then you can use standard ruby commands like `bundle exec
rspec` locally which makes the project *much* more approachable for ruby
folks. It isn't authoritative though. Jenkins will still use docker, as
will building the docs.

I've not gone so far as to use rake for ruby stuff yet. I'll get there
in a subsequent PR probably.


Dockerfile
```diff
@@ -52,24 +52,6 @@ RUN install_packages \
 # have to be empty. So we delete them.
 RUN rm -rf /var/log/nginx && rm -rf /run
 
-# Gem inventory:
```
nik9000: I've moved these below the python deps because I think I'll manipulate them more frequently.

Gemfile.lock
```diff
@@ -0,0 +1,56 @@
+GEM
```
estolfo: Is there a reason you committed the `Gemfile.lock`?
Usually you don't have to, as it's generated when `bundle install` is run.

Dockerfile
```diff
@@ -79,3 +61,9 @@ RUN pip3 install \
   beautifulsoup4==4.7.1 \
   lxml==4.3.1 \
   pycodestyle==2.5.0
+
+# Install ruby deps with bundler to make things more standard for Ruby folks.
+RUN gem install bundler:2.0.1
```
estolfo: If you don't specify a version, RubyGems will install the latest stable version of a gem. I would recommend using the "pessimistic version constraint" or "twiddle wakka" (`~`) to specify you want the latest version in the 2 series of Bundler. Unless, of course, you specifically need to use `2.0.1`.
With the twiddle waka, you automatically get the latest version of Bundler 2.

Dockerfile
```diff
@@ -79,3 +61,9 @@ RUN pip3 install \
...
+RUN bundle config --global silence_root_warning 1
+COPY Gemfile* /
```
estolfo: Is there a reason you dockerignore the Gemfile and then copy it explicitly?

Gemfile
```diff
@@ -0,0 +1,13 @@
...
+
+gem "asciidoctor", "1.5.8"           # Used by the docs build
+gem "thread_safe", "0.3.6"           # Used by asciidoctor
+gem "asciidoctor-diagram", "1.5.12"  # Speculative
+gem "asciimath", "1.0.8"             # Speculative
+
+group :test do
+  gem "rspec", "3.8.0"
+  gem "rubocop", "0.64.0"
+end
```
estolfo: The same comment holds true here: typically, `Gemfile`s have more lenient version specifications. Unless you need to pin to the very specific versions, you might want to add some leniency with the twiddle waka (`~`) so that you'll install the latest versions. That is, if you trust the gem maintainers to follow semver ; )

resources/asciidoctor/README
```diff
@@ -1,3 +1,12 @@
...
 the actual extensions and the spec directory contains the tests. You can run
-the tests with `build_docs --self-test`.
\ No newline at end of file
+the tests with `build_docs --self-test` to reproduce the authoritative, in
+docker tests. You can also use more ruby standard stuff like:
+
+```
+bundle install
+bundle exec rspec
+bundle exec rubocop
```
estolfo: woohoo! Now I can run the tests individually = )

Dockerfile
nik9000: I dockerignore everything *but* the Gemfile. There are tons of files in the repo and they don't need to go to docker, but it needs the Gemfile and the lock.

Gemfile.lock
nik9000: I committed it because the bundler folks [suggest](https://bundler.io/v2.0/guides/faq.html#committing-lockfiles-in-libraries) it.

Dockerfile
nik9000: I want to keep the generated docker image as consistent as possible. A lot of that parnoia comes from my experience with Java but I rather like it. I'm happy to leave a comment at the top of the dockerfile about how the version numbers aren't explicit.

Gemfile.lock
nik9000: That link looks like it is only for gems, actually. And I'm not one.

Gemfile
nik9000: I want the docker image to be as consistent as possible. I'm ok with using the `~` so long as I can use the lockfile to keep it consistent though.

Dockerfile
estolfo: oops, sorry missed the `!` in the `dockerignore`.

Gemfile.lock
estolfo: Yeah, I think they recommend it if you develop a gem...though I don't do it = )

nik9000:
@estolfo, I think I have it now! Could you have a look?

Gemfile
```diff
@@ -0,0 +1,15 @@
+# IMPORTANT: If you change this file you should run `bundle lock` to
+# regenerate Gemfile.lock or building the docker image will fail.
+source "https://rubygems.org"
+
+ruby ">=2.3"
```
nik9000: Let me use the `~>` for this one too.....

Dockerfile
estolfo: You and future maintainers of this file will just have to make sure to update this line if there are newer versions of `Bundler` you'd like to use.

Gemfile
estolfo: I think this approach is good now. Although you are committing the `Gemfile.lock`, if you ever want to update all gems at once, you can just run bundle install and commit the `Gemfile.lock` again. Otherwise, you would have to keep your eye on each gem individually to know when updates are available and edit this file.

nik9000: > Otherwise, you would have to keep your eye on each gem individually to know when updates are available and edit this file.

Right! I think we'll want to periodically blast the lockfile and `bundle install` again to see what we get.

nik9000:
Thanks for all of your reviews @estolfo!

PR #765, opened 2019-04-01
Begin moving integration tests to rspec

nik9000:
This begins the process of moving integration tests from the `Makefile`
in `integtest` to spec in `integtest/spec`. It begins by switching some
of the "expected files" tests and is meant as a pattern to build on.

The interesting thing from an rspec perspective here is that we want to
run each conversion once, before any assertions. Rspec's standard `let`
based building blocks aren't quite right for this. Instead I rely on
careful placement of `before(:context)` blocks. Blocks hooked to
`before(:context)` will run before the context in which they are
defined.

Here is a fairly normal context, asserting that beta[] works as expected:
```
  context 'when the book contains beta[]' do
    def make_input
      write_source 'index.asciidoc', <<~ASCIIDOC
        #{HEADER}
        beta[]

        Words
      ASCIIDOC
    end
    include_context 'convert'

    it 'copies the warning image' do
      expect(dest_file('images/icons/warning.png')).to file_exist
    end
    context 'chapter.html' do
      include_context 'page body'
      it 'includes the warning image' do
        expect(body).to include(
          '<img alt="Warning" src="images/icons/warning.png" />'
        )
      end
      it 'includes the beta text' do
        expect(body).to include(
          'The design and code is less mature than official GA features'
        )
      end
    end
  end
```

There is a fair bit going on so I'll annotate. First we define
`make_input` which is called later to build the source files and returns
the "root" source file that is passed as the argument to `--doc`. Next
we include the `convert` context. This defines the `before(:context)`
hooks that builds a temp directory, calls `make_input` to setup the
source documents, and performs the conversion. It also defines an
`after(:context)` hook to clean up the temp directory. The important bit
here is that you can have `context`s inside of this context but that the
temp directory survives all of them.

Next we make a fairly banal assertion that we copy an image.
Interestingly, the error reporting from rspec's built in `exists`
matcher is total garbage so I make my own and use it here.

Next we make a context called `chapter.html` to make assertions about
the `chapter.html` page. The first thing we do is include the `page
body` context which adds an assertion that the page exists and defines
some `lets` which we can use to assert things about the page. We then
assert some things about the beta widget that should be on that page.

Part of #787


integtest/spec/spec_helper.rb
```diff
@@ -0,0 +1,69 @@
...
+  config.disable_monkey_patching!
+
+  config.expect_with :rspec do |c|
+    c.syntax = :expect
+  end
+end
+
+RSpec.shared_context 'tmp dirs' do
+  before(:context) do
+    @tmp = Dir.mktmpdir
```
estolfo: I always favor `let` blocks when creating variables in case some tests don't need the variable and the work to create the variable can be avoided.
Why did you choose to use an instance variable here?

integtest/spec/single_book_spec.rb
```diff
@@ -0,0 +1,209 @@
+# frozen_string_literal: true
+
+RSpec.describe 'building a single book' do
+  ##
+  # Builds a path to afile in the source.
+  def source_file(file)
```
estolfo: Would these helper methods be useful for other tests? If so, it might be worth making a `support` directory in the `spec` directory so other spec files can use it.

integtest/spec/spec_helper.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  # Disable RSpec exposing methods globally on `Module` and `main`
+  config.disable_monkey_patching!
+
+  config.expect_with :rspec do |c|
+    c.syntax = :expect
+  end
+end
+
+RSpec.shared_context 'tmp dirs' do
```
estolfo: Are you going to use this context in other files as well as you write more rspec tests?

integtest/spec/single_book_spec.rb
```diff
@@ -0,0 +1,209 @@
...
+
+  ##
+  # Include in a context named after a page to add an assertion that the page
+  # was created and to create some variables extracted from the page's contents
+  #
+  # body - the non-template body of the page
+  # title - the title of the page
+  shared_context 'page body' do
+    let(:file) do |example|
+      dest_file(example.example_group.description)
```
estolfo: hmm this is a tight and implicit requirement of the `page body` context blocks. Is there a better and more explicit way to capture the argument to `dest_file`?

integtest/spec/single_book_spec.rb
```diff
@@ -0,0 +1,209 @@
...
+    end
+    it 'is created' do
+      expect(file).to file_exist
+    end
+    let(:body) do
+      return nil unless File.exist? file
```
estolfo: This can also be `return unless File.exist? file` =)

integtest/spec/single_book_spec.rb
```diff
@@ -0,0 +1,209 @@
...
+
+      File.open(dest_file(file), 'r:UTF-8') do |f|
+        f.read
+         .sub(/.+<!-- start body -->/m, '')
+         .sub(/<!-- end body -->.+/m, '')
+      end
+    end
+    let(:title) do
+      return nil unless body
```
estolfo: It's more ruby-like to write `return unless body`

integtest/spec/single_book_spec.rb
```diff
@@ -0,0 +1,209 @@
...
+      context 'chapter.html' do
+        include_context 'page body'
+        it 'has the right tile' do
+          expect(title).to eq('Chapter')
+        end
+      end
+    end
+
+    context 'when the file ends in .asciidoc' do
+      def file_name
```
estolfo: It seems strange to me to use a method here instead of a `let` block. Is there a reason for this?

integtest/spec/spec_helper.rb
nik9000: `let` variables are scoped to a single example but these live for the entire context. I need these to live for the context because I only want to do each conversion once.

integtest/spec/single_book_spec.rb
nik9000: Probably! I put the ones that I 100% sure would be shared in spec_helper and the ones I wasn't 100% sure of I left here and figured I'd move when they turned out they were helpful.

integtest/spec/spec_helper.rb
nik9000: I think I'll use it in all of the context that do conversions, yes. I will need to document it, and probably rename it.

integtest/spec/single_book_spec.rb
nik9000: I expect I'll include this many, many times and I liked the convenience of this. I can certainly have something like

```
context 'for index.html' do
  let(:file_name) { 'index.html' }`
end
```

Instead of making a shared context I could do something like this:
```
def page_context(file_name) do
  context "for #{file_name}" do
    let(:file) do |example|
    ....
    yield
  end
end
```

That way it is a bit more explicit that you are making a context for checking on the page.

integtest/spec/single_book_spec.rb
nik9000: +1

That is a holdover from my Java days.

integtest/spec/single_book_spec.rb
nik9000: +1

integtest/spec/single_book_spec.rb
nik9000: `let` block happen too late! The `begin` block can't see the input.

I wonder if I could do this another way, maybe replacing the `convert` shared context with something like:
```
def convert_context
    before(:context) do
      from = yield
      init_repo(File.expand_path('..', from))
      # Use popen here instead of capture to keep stdin open to appease the
      # docker-image-always-removed paranoia in build_docs.pl
      _stdin, out, wait_thr = Open3.popen2e(
        '/docs_build/build_docs.pl', '--in_standard_docker',
        *convert_args(from, @dest)
      )
      status = wait_thr.value
      raise_status cmd, out, status unless status.success?

      out
    end
    ...
end
```

nik9000:
@estolfo I've pushed a few patches that add some helper methods - some of them for setting up the test context and some of them for running commands during context setup.

integtest/spec/helper/dsl.rb
```diff
@@ -0,0 +1,89 @@
...
+      # Yield to the block to add more tests.
+      class_exec(&block)
+    end
+  end
+  shared_context 'page' do |file_name|
+    let(:file) do
+      dest_file(file_name)
+    end
+    let(:body) do
+      return nil unless File.exist? file
```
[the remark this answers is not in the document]
nik9000: Silly java coming out agian.

integtest/spec/helper/dsl.rb
```diff
@@ -0,0 +1,89 @@
...
+
+      File.open(dest_file(file), 'r:UTF-8') do |f|
+        f.read
+         .sub(/.+<!-- start body -->/m, '')
+         .sub(/<!-- end body -->.+/m, '')
+      end
+    end
+    let(:title) do
+      return nil unless body
```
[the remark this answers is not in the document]
nik9000: :+1:

integtest/spec/spec_helper.rb
```diff
@@ -0,0 +1,68 @@
...
+ENV['GIT_AUTHOR_EMAIL'] = 'test@example.com'
+ENV['GIT_COMMITTER_NAME'] = 'Test'
+ENV['GIT_COMMITTER_EMAIL'] = 'test@example.com'
+
+RSpec.configure do |config|
+  # Enable flags like --only-failures and --next-failure
+  config.example_status_persistence_file_path = '.rspec_status'
+
+  # Disable RSpec exposing methods globally on `Module` and `main`
+  config.disable_monkey_patching!
```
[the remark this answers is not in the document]
nik9000: I picked this up from somewhere saying it was best practice not to use the monkey patched stuff and have brought it along every time I use rspec. It has been a while since I tracked down where the advice came from, but it seems to have served me well.

integtest/spec/spec_helper.rb
```diff
@@ -0,0 +1,68 @@
...
+    File.exist? actual
+  end
+  failure_message do |actual|
+    msg = "expected that #{actual} exists"
+    parent = File.expand_path('..', actual)
+    msg unless Dir.exist? parent
+
+    entries = Dir.entries(parent).reject { |e| e.start_with? '.' }
+    msg + " but only #{entries} exist"
+  end
```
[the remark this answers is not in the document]
nik9000: :+1:

nik9000:
> I just thumbs-up-ed this but I think if you use let!, the let block is run once at the beginning and lives for the entire context. Have you tried let!?

I'll give it another shot. I believe it invokes the method once per example though.

nik9000:
Oh yeah, `let!` doesn't work there:

```
      RuntimeError:
        let declaration `src` accessed in a `before(:context)` hook at:
          /docs_build/integtest/spec/helper/dsl.rb:53:in `block in convert_single_before_context'
      
        `let` and `subject` declarations are not intended to be called
        in a `before(:context)` hook, as they exist to define state that
        is reset between each example, while `before(:context)` exists to
        define state that is shared across examples in an example group.
```

Maybe there is a way around the instance variables, but at least for now they are fairly contained.

nik9000:
Thanks for reviewing @estolfo! I'll try and build on this for other sorts of tests soon!

PR #768, opened 2019-04-02
Adds chunk level for Installation and upgrade guide

lcawl:
Per https://github.com/elastic/docs#chunking-in-the-right-place:

> By default, each part (= Part Title) and chapter (== Chapter Title) is "chunked" into a separate HTML file. However, for bigger books, such as the Elasticsearch reference docs, this results in enormous pages. These bigger books are also chunked at the first section level (=== Section One Title). 

The "Breaking changes" page in the Installation and Upgrade Guide is getting quite large as we add more products' information.

This PR proposes to add --chunk 1 to the Installation and Upgrade Guide, which will result in the following layout:

image


debadair:
LGTM

PR #769, opened 2019-04-02
Use more standard rspec for the care admonition

nik9000:
This switches the way we write the tests for the "care admonitions" to a
more rspec-standard style using nested `context`s to describe the input.
It also adds support to the new `convert` shared context for
automatically asserting there nothing was logged during the conversion
process. This is what we want the vast majority of the time so you have
to take some action to disable the automatic assertion, namely `let`ing
a variable to `true`.


resources/asciidoctor/spec/spec_helper.rb
```diff
@@ -55,13 +57,21 @@ def convert(input, extra_attributes = {}, warnings_matcher = eq(''))
...
-    Asciidoctor.convert input,
+    converted = Asciidoctor.convert input,
         :safe       => :unsafe,  # Used to include "funny" files.
         :backend    => :docbook45,
         :logger     => convert_logger,
         :doctype    => :book,
         :attributes => attributes,
         :sourcemap  => true
+    should_assert_logs = defined?(assert_no_logs) ? assert_no_logs : true
+    if should_assert_logs && convert_logger.messages.empty? == false
```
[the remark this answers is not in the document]
nik9000: I started going that way but ended up not. I'll switch!

nik9000:
@estolfo, I've pushed a new patch. Could you have a look?

resources/asciidoctor/spec/spec_helper.rb
```diff
@@ -47,25 +63,41 @@ def convert(input, extra_attributes = {}, warnings_matcher = eq(''))
...
+    converted = internal_convert input, convert_logger, extra_attributes
+    if convert_logger.messages.empty? == false
+      raise "Expected no logs but got:\n" +
+        convert_logger.messages
+          .map { |l| "#{l[:severity]}: #{l[:message].inspect}" }
+          .join("\n")
+    end
+    converted
+  end
+end
```
estolfo: This feels clean to me now! But I now feel like raising an error in this let block should be a `it` test instead.

nik9000: So I started there but that has problems when you include the context and then nest. So, like this:
```
context 'foo' do
  include_context 'convert no logs'
  context 'bar' do
    let(:input) { ... }
    it 'blah' { ... }
  end
  context 'baz' do
    let(:input) { ... }
    it 'blah' { ... }
  end
end
```

When you do the assertion with `it` then you only get the no-log assertion in the top level.

estolfo: ah, I see. The only way to solve that is to do the following:

```ruby
context 'foo' do
  context 'bar' do
   include_context 'convert no logs'
    let(:input) { ... }
    it 'blah' { ... }
  end
  context 'baz' do
    include_context 'convert no logs'
    let(:input) { ... }
    it 'blah' { ... }
  end
end
```

But maybe that's annoyingly redundant.

nik9000: My problem with that is that it is super easy to forget!

estolfo: carry on then :)

nik9000: > carry on then :)

Does that mean "I approve of this PR?" :smiling_imp:

estolfo: Indeed!

PR #771, opened 2019-04-02
Asciidoctor: Drop perl magic variables

nik9000:
Drops the last usage of `$1` and `$2` from the ruby code because those
aren't idiomatic ruby. Nice in perl, not in ruby.


resources/asciidoctor/lib/elastic_compat_preprocessor/extension.rb
```diff
@@ -129,8 +129,12 @@ def reader.process_line(line)
 
         @in_attribute_only_block = false
         line.clear
-      elsif INCLUDE_TAGGED_DIRECTIVE_RX =~ line
-        return nil if preprocess_include_directive "elastic-include-tagged:#{$1}", $2
+      elsif (match = INCLUDE_TAGGED_DIRECTIVE_RX.match line)
+        target = match[1]
+        tag = match[2]
+        return nil if preprocess_include_directive(
```
estolfo: You can skip explicitly returning `nil` and just return =)

PR #772, opened 2019-04-02
Asciidoctor: Break up method for readability

nik9000:
Breaks up the method that we use for locating images to copy so it is
more readable. Hopefully. Rubocop identifies this method as being too
complex and long, so hopefully breaking it down into named, digestable
chunks helps.

### DIFF
--- a/resources/asciidoctor/lib/copy_images/copier.rb
+++ b/resources/asciidoctor/lib/copy_images/copier.rb
@@ -68,13 +68,11 @@ module CopyImages
       return to_check unless resources
       return to_check if resources.empty?
 
-      begin
-        to_check + CSV.parse_line(resources)
-      rescue CSV::MalformedCSVError => error
-        logger.error message_with_context "Error loading [resources]: #{error}",
-            :source_location => block.source_location
-        to_check
-      end
+      to_check + CSV.parse_line(resources)
+    rescue CSV::MalformedCSVError => error
+      logger.error message_with_context "Error loading [resources]: #{error}",
+          :source_location => block.source_location
+      to_check
     end
 
     ##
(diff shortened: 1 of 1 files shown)


resources/asciidoctor/lib/copy_images/copier.rb
```diff
@@ -66,8 +55,34 @@ def find_source(block, uri)
...
+      return to_check if resources.empty?
+
+      begin
+        to_check + CSV.parse_line(resources)
+      rescue CSV::MalformedCSVError => error
+        logger.error message_with_context "Error loading [resources]: #{error}",
+            :source_location => block.source_location
+        to_check
+      end
+    end
```
estolfo: Instead of writing individual comments, I thought it be easier to visualize if I rewrote the method and put the two comments inline:

```ruby
    def roots_to_check(block)
      to_check = [block.document.base_dir]

      resources = block.document.attr 'resources'
      # I find it clearer to use `if` to check both scenarios in which `to_check` should be returned, instead of having an `if` followed by an `unless`
      return to_check if resources.nil? || resources.empty?

      to_check + CSV.parse_line(resources)
    # There is an implicit `begin` for a method, so you can just rescue an error here. It's not necessary, I just thought I would point out some fancy Ruby. And it makes the method two whole lines shorter ;)
    rescue CSV::MalformedCSVError => error
      logger.error message_with_context "Error loading [resources]: #{error}",
          :source_location => block.source_location
      to_check
    end
```

nik9000: Hmmmm! I like being explicit about which line can through the error, but in this case it is petty obvious which one can. I'll take it!

PR #774, opened 2019-04-02
Asciidoctor: Drop a few hash rockets

nik9000:
It looks like it is more idiomatic in ruby to avoid hash rockets where
possible. This drops *some* of them.


nik9000:
> Isn't this syntax so much nicer? It's especially nice when you are dealing with json a lot.

It really does look like much more like JSON.

PR #776, opened 2019-04-02
Use more standard rspec in include_tagged spec

nik9000:
Use nested contexts when describing `include_tagged` in its spec file
because that is more idiomatic. Adds two additional asseretions that
were missing.


resources/asciidoctor/spec/elastic_include_tagged_spec.rb
```diff
@@ -13,148 +13,174 @@
     Asciidoctor::Extensions.unregister_all
   end
 
-  it "includes a tag" do
-    actual = convert <<~ASCIIDOC
+  INCLUDE_FILE = 'resources/elastic_include_tagged/Example.java'
```
estolfo: It seem strange to define a constant in a spec file. Maybe you can pass in a `include_file` variable to the method and define in a `let` block so it can also be used [here](https://github.com/elastic/docs/pull/776/files#diff-66f24cf664d01c71e63a3626975d6d7cR138)

nik9000:
@estolfo, I pushed a change that drops the constant and makes more use of `let`. Please let me know if you like it!

resources/asciidoctor/spec/elastic_include_tagged_spec.rb
```diff
@@ -135,52 +125,51 @@ def expected_include(include_body)
...
     it 'includes nothing' do
-      expect(converted).to eq(expected_include '')
+      expect(converted).to eq(expected)
     end
   end
   context 'when called more than one parameter' do
-    include_context 'convert'
-    let(:input) { include_input 'tag,otherparam' }
+    let(:tag)              { 'tag,otherparam' }
+    let(:expected_include) { ''               }
```
estolfo: strange spacing..

resources/asciidoctor/spec/elastic_include_tagged_spec.rb
```diff
@@ -91,27 +83,26 @@ def expected_include(include_body)
...
-        "ERROR: <stdin>: line 2: include file not found: #{__dir__}/#{file}"
-      )
+      expect(logs).to include(expected_log)
     end
   end
   context "when including a tag that doesn't have a start tag" do
-    include_context 'convert'
-    let(:input) { include_input 'missing_start' }
+    let(:tag)              { 'missing_start' }
+    let(:expected_include) { ''              }
```
estolfo: also strange spacing

nik9000: I got this from [here](https://rubocop-rspec.readthedocs.io/en/latest/cops_rspec/#rspecalignrightletbrace) but I'll drop it.

nik9000:

resources/asciidoctor/spec/elastic_include_tagged_spec.rb
estolfo: Oh, I thought maybe it was a typo. It looked strange but it could just be me!

PR #778, opened 2019-04-02
Switch the Transport Client docs to Asciidoctor

nik9000:
Switches the Transport Client's docs from the no-longer-maintained
AsciiDoc project to the actively maintained Asciidoctor project. In the
process it speeds up building them by a factor of three.

### DIFF
--- a/doc_build_aliases.sh
+++ b/doc_build_aliases.sh
@@ -114,7 +114,7 @@ alias docbldepi='$GIT_HOME/docs/build_docs.pl --doc $GIT_HOME/elasticsearch/docs
 
 alias docbldjvr='$GIT_HOME/docs/build_docs.pl --doc $GIT_HOME/elasticsearch/docs/java-rest/index.asciidoc --chunk 1'
 
-alias docbldejv='$GIT_HOME/docs/build_docs.pl --doc $GIT_HOME/elasticsearch/docs/java-api/index.asciidoc --chunk 1'
+alias docbldejv='$GIT_HOME/docs/build_docs -asciidoctor --doc $GIT_HOME/elasticsearch/docs/java-api/index.asciidoc --chunk 1'
 
 alias docbldejs='$GIT_HOME/docs/build_docs --asciidoctor --doc $GIT_HOME/elasticsearch-js/docs/index.asciidoc'
 
--- a/doc_build_aliases.sh
+++ b/doc_build_aliases.sh
@@ -114,7 +114,7 @@ alias docbldepi='$GIT_HOME/docs/build_docs.pl --doc $GIT_HOME/elasticsearch/docs
 
 alias docbldjvr='$GIT_HOME/docs/build_docs.pl --doc $GIT_HOME/elasticsearch/docs/java-rest/index.asciidoc --chunk 1'
 
-alias docbldejv='$GIT_HOME/docs/build_docs -asciidoctor --doc $GIT_HOME/elasticsearch/docs/java-api/index.asciidoc --chunk 1'
+alias docbldejv='$GIT_HOME/docs/build_docs --asciidoctor --doc $GIT_HOME/elasticsearch/docs/java-api/index.asciidoc --chunk 1'
 
 alias docbldejs='$GIT_HOME/docs/build_docs --asciidoctor --doc $GIT_HOME/elasticsearch-js/docs/index.asciidoc'
 
--- a/conf.yaml
+++ b/conf.yaml
@@ -286,6 +286,7 @@ contents:
                 tags:       Clients/Java
                 subject:    Clients
                 chunk:      1
+                asciidoctor: true
                 sources:
                   -
                     repo:   elasticsearch


nik9000:
The HTML output is mostly the same modulo line breaks. It differs in the following ways:

_log4j_2_logger.html:
```patch
- AsciiDoc
+ Asciidoctor
@@ -614,6 +614,7 @@
            <pre class="programlisting prettyprint lang-properties">appender.console.type = Console
 appender.console.name = console
 appender.console.layout.type = PatternLayout
+appender.console.layout.pattern = [%d{ISO8601}][%-5p][%-25c] %marker%m%n
 
 rootLogger.level = info
 rootLogger.appenderRef.console.ref = console</pre>
```

This one is an improvement.

java-compound-queries.html:
```patch
- AsciiDoc
+ Asciidoctor
@@ -603,7 +603,14 @@
              </span>
             </dt>
             <dd>
-             A query which wraps another query, but executes it in filter context. All matching documents are given the same “constant”
+             A query which wraps another query, but executes it in filter context. All matching documents are given the same
+             <span class="quote">
+              “
+              <span class="quote">
+               constant
+              </span>
+              ”
+             </span>
              <code class="literal">
               _score
              </code>
~~~

I think this one is safe to ignore.

indexed-scripts.html:
```patch
- AsciiDoc
+ Asciidoctor
@@ -583,23 +583,23 @@
           </p>
           <div class="pre_wrapper">
            <pre class="programlisting prettyprint lang-java">PutIndexedScriptResponse = client.preparePutIndexedScript()
-                         .setScriptLang("painless")
-                         .setId("script1")
-                         .setSource("script", "_score * doc['my_numeric_field'].value")
-                         .execute()
-                         .actionGet();
+			 .setScriptLang("painless")
+			 .setId("script1")
+			 .setSource("script", "_score * doc['my_numeric_field'].value")
+			 .execute()
+			 .actionGet();
 
 GetIndexedScriptResponse = client.prepareGetIndexedScript()
-                            .setScriptLang("painless")
-                            .setId("script1")
-                            .execute()
-                            .actionGet();
+			    .setScriptLang("painless")
+			    .setId("script1")
+			    .execute()
+			    .actionGet();
 
 DeleteIndexedScriptResponse = client.prepareDeleteIndexedScript()
-                            .setScriptLang("painless")
-                            .setId("script1")
-                            .execute()
-                            .actionGet();</pre>
+			    .setScriptLang("painless")
+			    .setId("script1")
+			    .execute()
+			    .actionGet();</pre>
           </div>
           <p>
            To store templates simply use "mustache" for the scriptLang.
```

There are tabs in the source file. I expect they aren't hurting anything
to leave.

delete-by-query.html:
```
- AsciiDoc
+ Asciidoctor
@@ -587,11 +587,7 @@
              Deprecated in 1.5.3.
             </h3>
             <p>
-             "Delete by Query will be removed in 2.0: it is problematic since it silently forces a refresh which can quickly cause OutOfMemoryError during concurrent indexing, and can also cause primary and replica to become inconsistent. Instead, use the
-             <a class="link" href="search.html" title="Search API">
-              scroll/scan API
-             </a>
-             to find all matching ids and then issue a bulk request to delete them.
+             "Delete by Query will be removed in 2.0: it is problematic since it silently forces a refresh which can quickly cause OutOfMemoryError during concurrent indexing
             </p>
            </div>
           </div>
```

This one will have to be fixed. We need to escape the `,` because it
*looks* like another parameter to the macro.


java-term-level-queries.html:
```patch
- AsciiDoc
+ Asciidoctor
@@ -1427,11 +1427,7 @@
               Deprecated in 7.0.0.
              </h3>
              <p>
-              Types are being removed, prefer filtering on a field instead. For more information, please see
-              <a class="ulink" href="https://www.elastic.co/guide/en/elasticsearch/reference/7.x/removal-of-types.html" target="_top">
-               Removal of mapping types
-              </a>
-              .
+              Types are being removed
              </p>
             </div>
```

I'll also have to fix this one. Same issue as above with the `,` being
escaped.

nik9000:
> I'll also have to fix this one. Same issue as above with the `,` being
> escaped.

I'll see if I can fix this one now....

nik9000:
Both of the last two issues are only fixable by wrapping the parameter in `"`s which causes AsciiDoc to render the quotes. I could do a fancy `if this is asciidoctor` style thing but I don't think it is worth it here. I think it'd be simpler to push the fixes for these after switching to asciidoctor.

nik9000:
@lcawl, what do you think about my plan [here](https://github.com/elastic/docs/pull/778#issuecomment-479217348)?

doc_build_aliases.sh
```diff
@@ -114,7 +114,7 @@ alias docbldepi='$GIT_HOME/docs/build_docs.pl --doc $GIT_HOME/elasticsearch/docs
 
 alias docbldjvr='$GIT_HOME/docs/build_docs.pl --doc $GIT_HOME/elasticsearch/docs/java-rest/index.asciidoc --chunk 1'
 
-alias docbldejv='$GIT_HOME/docs/build_docs.pl --doc $GIT_HOME/elasticsearch/docs/java-api/index.asciidoc --chunk 1'
+alias docbldejv='$GIT_HOME/docs/build_docs -asciidoctor --doc $GIT_HOME/elasticsearch/docs/java-api/index.asciidoc --chunk 1'
```
lcawl:
```suggestion
alias docbldejv='$GIT_HOME/docs/build_docs --asciidoctor --doc $GIT_HOME/elasticsearch/docs/java-api/index.asciidoc --chunk 1'
```

nik9000:
:+1:

lcawl:
I created two PRs to fix the deprecations: 
https://github.com/elastic/elasticsearch/pull/40981
https://github.com/elastic/elasticsearch/pull/40980
I don't think we need to wait til after the migration to fix those. They were too verbose in my opinion.

nik9000:
Thanks for reviewing @lcawl!

PR #779, opened 2019-04-02
Fix ecs url

dedemorton:
Fixes broken doc build.


dedemorton:
@lcawl FYI that this is a temporary fix. Karen should work with you to come up with a better solution for handling ECS versions, which do not track to Stack versions.

PR #781, opened 2019-04-03
Asciidoctor: Drop hash rockets

nik9000:
Drops the hash rockets that are not already being covered by #775.


nik9000:

PR #782, opened 2019-04-03
Asciidoctor: Fix extra spacing

nik9000:
Turns on the ExtraSpacing cop and removes two spaces to make it happy.


nik9000:

PR #790, opened 2019-04-04
Use more standard rspec for change admonition

nik9000:
Switches the rspec in the change admonition spec to nested contexts
which is more idiomatic. Also renames the file.


nik9000:

PR #792, opened 2019-04-04
Use more standard rspec in cramped include spec

nik9000:
Replaces the more junit like tests in the cramped include spec with
descriptive contexts which is more idiomatic rspec.


resources/asciidoctor/spec/cramped_include_spec.rb
```diff
@@ -8,7 +8,6 @@
   before(:each) do
     Asciidoctor::Extensions.register do
       preprocessor CrampedInclude
-      preprocessor ElasticCompatPreprocessor
```
nik9000: This wasn't needed.

resources/asciidoctor/spec/cramped_include_spec.rb
```diff
@@ -17,60 +16,40 @@
...
-    DOCBOOK
-    expect(actual).to eq(expected.strip)
+        include::resources/cramped_include/colist1.adoc[]
+        include::resources/cramped_include/colist2.adoc[]
+      ASCIIDOC
+    end
+    it 'renders both callout lists' do
+      expect(converted).to include('<callout arearefs="CO1-1">')
+        .and(include('<callout arearefs="CO2-1">'))
+    end
```
estolfo: Is there a reason you didn't write it like the following?

```ruby
expect(converted).to include('<callout arearefs="CO1-1">')
expect(converted).to include('<callout arearefs="CO2-1">')
```

nik9000: I suppose the error message is a *little* more descriptive the way I have it, but not much. Either way is fine by me. If you prefer your way I'll switch it.

nik9000:

PR #799, opened 2019-04-09
Asciidoctor: Change invocation rules for macros

nik9000:
Changes the invocation rules for legacy style admonition macros to only
invoke the block version if it is alone on a line. This more closely
mirrors the customizations that Elastic did to AsciiDoc.


nik9000:
I'm running a full [rebuild](https://elasticsearch-ci.elastic.co/job/elastic+docs+master+build+nik/) into the `built-docs-nik` repo that we do not deploy to make sure this doesn't have any unintended side effects.

nik9000:
I think this may actually cause trouble. I'll investigate!

nik9000:
I tracked down the issue and pushed a fix. Then I started [another](https://elasticsearch-ci.elastic.co/job/elastic+docs+master+build+nik/4389/console) build.

nik9000:
> I tracked down the issue and pushed a fix. Then I started [another](https://elasticsearch-ci.elastic.co/job/elastic+docs+master+build+nik/4389/console) build.

That build says that this change doesn't change anything unexpected. :success:

nik9000:

PR #800, opened 2019-04-10
Use more standard rspec for image copying spec

nik9000:
Replace most of the specification for copying images with nested
contexts which are more idiomatic rspec. Use the opportunity to add more
extensive testing.


resources/asciidoctor/spec/spec_helper.rb
```diff
@@ -65,12 +65,17 @@ def internal_convert(input, convert_logger, extra_attributes)
 #   logs             - lines logged
 RSpec.shared_context 'convert with logs' do
   let(:convert_logger) { Asciidoctor::MemoryLogger.new }
-  let!(:converted) do
-    # We use let! here to force the conversion because it populates the logger
+  let(:converted) do
+    # Using let! here would stop us from having to explicitly evaluate
+    # `converted` in the let for `logs` but it'd cause `converted` to be
+    # evaluated before `before(:example)` blocks
```
estolfo: This is not quite related but I just looked up what the rspec ordering was exactly for `before`, `let!`, etc and I found this https://relishapp.com/rspec/rspec-core/v/3-8/docs/configuration/overriding-global-ordering
I've never heard about that before, or used it, but I thought it was interesting and would possibly to useful to you at some point.

resources/asciidoctor/spec/copy_images_spec.rb
```diff
@@ -29,23 +29,41 @@ def copy_attributes(copied)
...
   shared_context 'convert intercepting images' do
-    let(:assert_no_logs) { false }
+    # TODO: once we've switched all of the examples in this file we can probably
+    #  drop this.
     include_context 'convert with logs'
 
     # [] is the initial value but it is mutated by the conversion
-    let(:copied) { [].dup }
-    let(:convert_attributes) { copy_attributes(copied) }
+    let(:copied_storage) { [].dup }
```
estolfo: I don't follow why you are duping the Array?

resources/asciidoctor/spec/spec_helper.rb
nik9000: I'll take a look at it!

resources/asciidoctor/spec/copy_images_spec.rb
nik9000: The array is mutated by the conversion process. We stick tuples for everything that we would copy into it instead. I think I'd like to replace it with a mock object eventually. Let me play around with not duping it and see if that breaks this.

nik9000:
@estolfo, I've pushed a patch that removes the `dup` because we don't need it. We still mutate that array which isn't a thing I'd like to keep forever, but at least the funny `dup` isn't there.

estolfo:
Yeah, the `dup` didn't seem necessary because the `let` block is only executed once. `[]` is the same as `Array.new` so you'll get a new array when the `let` block is executed.

nik9000:

PR #801, opened 2019-04-10
Asciidoctor: Enable cyclomatic complexity cop

nik9000:
Breaks up two large methods and enables that cyclomatic complexity cop.


resources/asciidoctor/spec/elastic_include_tagged_spec.rb
```diff
@@ -118,8 +118,7 @@
     end
     it 'logs a warning about the missing tag' do
       expect(logs).to eq(
-        'WARN: <stdin>: line 5: elastic-include-tagged missing ' \
-        'start tag [missing_start]'
+        'WARN: <stdin>: line 5: missing start tag [missing_start]'
```
nik9000: The `elastic-include-tagged` bit in the error message wasn't really doing anyone any good and just made it longer.

resources/asciidoctor/lib/elastic_include_tagged/extension.rb
```diff
@@ -11,73 +11,101 @@
...
-      warn Asciidoctor::Reader::Cursor.new(path, relpath, relpath, start_of_include),
-          "missing end tag [#{tag}]"
-    end
-    reader.push_include included_lines, path, relpath, start_of_include, attrs
-  end
 
-  def warn(cursor, message)
-    logger.warn message_with_context "elastic-include-tagged #{message}",
-      source_location: cursor
+    def warn(message, cursor = @reader.cursor)
```
nik9000: I feel like this is much cleaner!

resources/asciidoctor/lib/elastic_compat_preprocessor/extension.rb
```diff
@@ -119,77 +119,115 @@ class ElasticCompatPreprocessor < Asciidoctor::Extensions::Preprocessor
...
   def process(_document, reader)
-    reader.instance_variable_set :@in_attribute_only_block, false
-    reader.instance_variable_set :@code_block_start, nil
-    def reader.process_line(line)
+    reader.extend ReaderExtension
+  end
+
+  ##
+  # Extensions to the Reader object that implement the conversions.
+  module ReaderExtension
```
nik9000: I feel like this makes it much more readable, mostly because I can break things into named methods.

resources/asciidoctor/lib/elastic_compat_preprocessor/extension.rb
```diff
@@ -119,77 +119,115 @@ class ElasticCompatPreprocessor < Asciidoctor::Extensions::Preprocessor
...
+      line.clear
+    end
+
+    ##
+    # Process a start block when, potentially shifting into the "attribute only"
+    # block state if the block that is starting only contains attributes. If
+    # we enter into that state then we eat the block delimiter to work around
+    # a scoping difference between AsciiDoc and Asciidoctor.
+    def process_start_block(line)
+      lines = self.lines
```
estolfo: What is the intention of this line?

resources/asciidoctor/lib/elastic_compat_preprocessor/extension.rb
```diff
@@ -119,77 +119,115 @@ class ElasticCompatPreprocessor < Asciidoctor::Extensions::Preprocessor
...
+      lines.shift
+      while Asciidoctor::AttributeEntryRx =~ (check_line = lines.shift)
+      end
```
estolfo: This looks strange. I would instead write it as something like:
`check_line  = lines.shift while Asciidoctor::AttributeEntryRx =~ lines[0]`

resources/asciidoctor/lib/elastic_compat_preprocessor/extension.rb
```diff
@@ -119,77 +119,115 @@ class ElasticCompatPreprocessor < Asciidoctor::Extensions::Preprocessor
...
+      # about it so we should do the asciidoctor standard thing and keep
+      # it intact. This is how we do that.
+      @look_ahead += 1
+      line
+    end
+
+    ##
+    # Process lines after they've been processed by the reader.
+    def postprocess(line)
+      return if line.nil?
```
estolfo: I would write instead `return unless line`

resources/asciidoctor/lib/elastic_compat_preprocessor/extension.rb
nik9000: This snapshots the lines remaining in the file so it can look ahead. `self.lines` returns a copy of the buffer so I can manipulate it while I go.

resources/asciidoctor/lib/elastic_compat_preprocessor/extension.rb
nik9000: :+1:. I keep getting mixed up about empty lines being truthy. In perl it is falsey. Javascript too.

resources/asciidoctor/lib/elastic_compat_preprocessor/extension.rb
nik9000: Would it be more clear if I gave it a different name? Like `readahead` or something?

nik9000:
@estolfo, I think this is ready for another round!

resources/asciidoctor/lib/elastic_compat_preprocessor/extension.rb
estolfo: Is `self.lines` something coming from `asciidoctor` (`Asciidoctor::Extensions::Preprocessor`)?

resources/asciidoctor/lib/elastic_compat_preprocessor/extension.rb
estolfo: Welcome to Rubyland =)

resources/asciidoctor/lib/elastic_compat_preprocessor/extension.rb
nik9000: Right! it is from Asciidoctor::Reader.

estolfo: Is there a reason you're use `self.lines`? Is the intent of the line to get a copy of the attribute?

nik9000: Right. I want the copy. I could just use `lines` if I didn't name the variable that though.

estolfo: oh, right. You are using `self.lines` because the variable is `lines`. That's fine.

nik9000:

PR #802, opened 2019-04-10
Share rubocop configuration

nik9000:
Moves the rubocop configuration to the root of the project so it can
be shared by the two places in the Makefiles that expclitly invoke
rubocop.

Also it adds a cop that detects `return nil` because I can't seem to
stop myself from writing it.


nik9000:

PR #805, opened 2019-04-11
integtest: Implement Asciidoctor dueling in rspec

nik9000:
Implements a test that compares the output of converting the docs using
Asciidoctor to using AsciiDoc and adds that to all "single book" test
cases. This allows us to entirely move the tests that used a few files
from the Makefile to rspec. Because we write those files to a tmpdir
directly from the rspec file we can now remove the asciidoc files in the
integtest folder.

Also moves the test for the `experimental` admonition into rspec to be
with its brother the `beta` admonition.


nik9000:
> looks good!

Thanks for the review! This one is super important for blasting things from the Makefile.

PR #806, opened 2019-04-11
Use standard rspec for edit_me spec

nik9000:
This replaces the large examples I started using with nested contexts
and smaller examples. It manages to remove most of the copy-and-paste
as well with two shared_examples.


nik9000:

PR #807, opened 2019-04-11
Drop unused stuff

nik9000:
This removes the unused `web` directory. Since that was the last thing
using the `.run` directory this drops that too. Finally, the `web`
directory was the last thing using the `Proc::PID` perl module so this
drops that.


nik9000:
I seem to have accidentally merged this when fixing up gitignore. It *should* be ok though.

PR #808, opened 2019-04-11
Drop kibana submodule

nik9000:
I think we added it accidentally during one of the old docs force pushes.
It doesn't seem to do anything and we haven't updated it since we added
it.


nik9000:
Thanks for reviewing @jarpy! This one confused me too!

PR #810, opened 2019-04-11
Integtest: Add tests for building all books

nik9000:
Adds a basic test for building all books. I expect we'll want to refactor
it significantly as we port more tests for Make to rspec, but this is a
start!


integtest/spec/all_books_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    convert_all_before_context do |src|
+      src.write 'source/index.asciidoc', <<~ASCIIDOC
+        = Title
+
+        == Chapter
+
+        Some text.
+      ASCIIDOC
+      src.init_repo 'source'
+      src.write 'conf.yaml', <<~YAML
```
nik9000: I think I'll end up having to make generating this file more parameterized and automatic but I'd like to wait for a follow up for that.

integtest/spec/helper/sh.rb
```diff
@@ -15,18 +15,6 @@ def sh(cmd)
     out
   end
 
-  ##
-  # Init a git repo in root and commit any files in it.
-  def init_repo(root)
```
nik9000: I've moved this to the `Source` object because we'll want to initialize many repos when testing `--all`

integtest/spec/helper/dsl/file_contexts.rb
```diff
@@ -0,0 +1,65 @@
...
+        # Yield to the block to add more tests.
+        class_exec(&block)
+      end
+    end
+    shared_context 'Dsl_file' do |file_name|
+      let(:file) do
+        dest_file(file_name)
+      end
+      let(:contents) do
+        return unless File.exist? file
```
estolfo: It looks strange to me to have a `return` in a `let` block. Maybe you can just write the line in a `if File.exist? file`

nik9000: It does feel weird? I'm not really sure the `if File.exist?` block is super clear either.  I think for me coming from Java the `return unless` is more clear but I can't speak to idiomatic ruby.

estolfo: I guess I think of `let` blocks not quite as methods, more like procs. But this is not a critical thing and if you prefer using `return`, that's no problem.

nik9000: Thanks!

Let me see if I can resolve the merge conflicts here.....

nik9000:
Thanks for reviewing @estolfo! This is a big help!

PR #814, opened 2019-04-12
Switch legacy graph docs to Asciidoctor

nik9000:
Switches the core of the docs build for the legacy graph docs from the
unmaintained AsciiDoc project to the actively maintained Asciidoctor
project.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -1346,6 +1346,7 @@ contents:
             index:      docs/public/graph/index.asciidoc
             branches:   [ 2.4, 2.3 ]
             private:    1
+            asciidoctor: true
             sources:
               -
                 repo:   x-pack


nik9000:
Here is the diff:

graph-limitations.html
```patch
- AsciiDoc
+ Asciidoctor
@@ -587,11 +587,15 @@
              a different field
             </em>
            </span>
-           held in another type or index. A good example of where that might be useful is if an ip address found in the ‘remote_host` field of an index called "weblogs20160101" might want to be followed up by looking for this value in the
+           held in another type or index. A good example of where that might be useful is if an ip address found in the
+           <code class="literal">
+            remote_host
+           </code>
+           field of an index called "weblogs20160101" might want to be followed up by looking for this value in the
            <code class="literal">
             ip_address
            </code>
-           field of an index called "knownthreats". For the Graph API to support this, extra mapping would be required to understand that the weblogs’
+           field of an index called "knownthreats". For the Graph API to support this, extra mapping would be required to understand that the weblogs'
            <code class="literal">
             remote_host
            </code>
```
These seem ok. We use some odd quotes in some of our docs.

nik9000:
Hearing no objections and given that this is a net-0 change to the rendered docs I've merged.

PR #816, opened 2019-04-12
Switch legacy reporting docs to Asciidoctor

nik9000:
Switches the core of the build for the legacy reporting docs from
AsciiDoc to Asciidoctor.


nik9000:
Thanks for reviewing @lcawl! I forgot to leave a diff on this one because it was prety much empty iirc.

PR #819, opened 2019-04-16
Integtest: Move all single book tests to rspec

nik9000:
Drops all of the single book tests from the integtest Makefile and adds
them to rspec, adding many more assertions in the process.


nik9000:
I'm going to merge this to get myself unblocked because it is "just" test code. It is a ton of test code and I'll be building on it but I'd still really appreciate some extra eyes on it at some point.

PR #820, opened 2019-04-16
Rewrite preprocessor rspec for care admonitions

nik9000:
Rewrites the rspec for the preprocessor's handling of care admonitions
to be more standard. Actually, it merges the testing of care admonitions
and change admonitions into a single `shared_examples`.


nik9000:

PR #822, opened 2019-04-17
Switch Logstash VPR to Asciidoctor

nik9000:
Switches the core of the docs build for the Logstash Versioned Plugin
Reference from the unmaintained AsciiDoc project to the actively
developed Asciidoctor project.


nik9000:
I've created this one as a draft because we'll need some changes before we're ready, but I'd like to make a spot to share the diff.

nik9000:
We're waiting on https://github.com/elastic/logstash-docs/pull/688 and probably a few followup changes before removing "draft" from this PR.

nik9000:
@karenzone I've opened https://github.com/elastic/logstash-docs/pull/691 for something that I saw but at this point the diff looks like a net positive! I'm still going over the diff to see if I see anything, but I think merging this now will fix more things than it breaks.

PR #824, opened 2019-04-19
Asciidoctor: Copy inline images in lists

nik9000:
Adds support for copying inline images declared in lists. Sadly,
Asciidoctor doesn't make this easy. They pre-convert lists so we have to
detect the image in docbook. I feel like it might have been a mistake to
try and copy images by traversing the tree, mostly because Asciidoctor
works in strange ways.


nik9000:
I noticed this when working on converting Kibana's book to Asciidoctor.

resources/asciidoctor/lib/copy_images/extension.rb
```diff
@@ -48,12 +49,17 @@ def process_block_image(block)
     end
 
     def process_inline_image(block)
-      return unless block.content_model == :simple
+      process_inline_image_from_source block, block.source if block.content_model == :simple
+      process_inline_image_from_converted block, block.text if
+        block.context == :list_item && block.parent.context == :olist
+    end
```
estolfo: You might want to reverse the syntax and write the `if` statement first.
I typically only use the post-fix `if` statement if it fits on one line and is fairly simple logic.

resources/asciidoctor/lib/copy_images/extension.rb
```diff
@@ -68,6 +74,22 @@ def process_inline_image(block)
       end
     end
 
+    ##
+    # Scan the inline image from the generated docbook. It is not nice that
+    # this is required there isn't much we can do about it. We *could* rewrite
```
estolfo: Maybe missing the word 'so'..?

resources/asciidoctor/lib/copy_images/extension.rb
nik9000: Got it!

nik9000:
@esdocs, I've fixed this up from your comments and reworked it to be a bit more consistent with itself.

nik9000:
@estolfo, I've fixed this up from your comments and reworked it a big to be more consistent. Could you have another look?

nik9000:
Thanks so much for reviewing @estolfo!

PR #830, opened 2019-04-23
Use sparse checkout to speed up building all docs

nik9000:
This begins to use sparse checkout to speed up building `--all` docs. It
does so in the simplest way possible at this point: instead of checking
out the entire target repo before checking out other repos we checkout
all of the target repo *except* the built docs before checking out the
other repos. This allows us to read the "tracker" file from the built
docs repo to pick which hashes to use for the other repos. We then
checkout the built docs in parallel with the other repos.

The bulk of changed code is around breaking out handling of the
"target" repo into a new class `TargetRepo` and creating a `BaseRepo`
that is shared between `TargetRepo` and `Repo`.

This doesn't save *that* much time locally but I expect it'll be
somewhat more substantial in CI because CI tends to spend more time
pulling changes. In addition, I think we can do better later by using
sparse checkout on the built docs. In most cases we're not actually
rebuilding all of the docs so we don't need to check them all out.

### DIFF
--- a/lib/ES/BaseRepo.pm
+++ b/lib/ES/BaseRepo.pm
@@ -1,0 +1,132 @@
+package ES::BaseRepo;
+
+use strict;
+use warnings;
+use v5.10;
+
+use Path::Class();
+use URI();
+use Encode qw(decode_utf8);
+use ES::Util qw(run);
+
+#===================================
+sub new {
+#===================================
+    my ( $class, %args ) = @_;
+
+    my $name = $args{name} or die "No <name> specified";
+    my $url  = $args{url}  or die "No <url> specified for repo <$name>";
+    if ( my $user = $args{user} ) {
+        $url = URI->new($url);
+        $url->userinfo($user);
+    }
+
+    my $dir = $args{dir} or die "No <dir> specified for repo <$name>";
+
+    my $reference_dir = 0;
+    if ($args{reference}) {
+        my $reference_subdir = $url;
+        $reference_subdir =~ s|/$||;
+        $reference_subdir =~ s|:*/*\.git$||;
+        $reference_subdir =~ s/.*[\/:]//g;
+        $reference_dir = $args{reference}->subdir("$reference_subdir.git");
+    }
+
+    bless {
+        name          => $name,
+        git_dir       => $dir->subdir("$name.git"),
+        url           => $url,
+        reference_dir => $reference_dir,
+        sub_dirs      => {},
+    }, $class;
+}
+
+#===================================
+sub update_from_remote {
+#===================================
+    my $self = shift;
+
+    my $git_dir = $self->git_dir;
+    local $ENV{GIT_DIR} = $git_dir;
+
+    my $name = $self->name;
+    eval {
+        unless ( $self->_try_to_fetch ) {
+            my $url = $self->url;
+            printf(" - %20s: Cloning from <%s>\n", $name, $url);
+            run 'git', 'clone', '--bare', $self->_reference_args, $url, $git_dir;
+        }
+        1;
+    }
+    or die "Error updating repo <$name>: $@";
+}
+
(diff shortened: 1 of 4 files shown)


nik9000:
Neat! Go CI go!

lib/ES/Repo.pm
```diff
@@ -5,49 +5,25 @@ use warnings;
 use v5.10;
 
 use Path::Class();
-use URI();
 use Encode qw(decode_utf8);
 use ES::Util qw(run sha_for);
 
+use base qw( ES::BaseRepo );
```
olksdr: It would be better to use a new [parent](https://metacpan.org/pod/parent) directive (more lightweight) instead of the old-school `base`.

lib/ES/Repo.pm
```diff
@@ -5,49 +5,25 @@ use warnings;
...
-
-    my $reference_dir = 0;
-    if ($args{reference}) {
-        my $reference_subdir = $url;
-        $reference_subdir =~ s|/$||;
-        $reference_subdir =~ s|:*/*\.git$||;
-        $reference_subdir =~ s/.*[\/:]//g;
-        $reference_dir = $args{reference}->subdir("$reference_subdir.git");
-    }
+    my $self = shift->SUPER::new(%args);
```
olksdr: Am I mistaken or this should be: 
```perl
 my $self = $class->SUPER::new($args);
```

lib/ES/TargetRepo.pm
```diff
@@ -0,0 +1,97 @@
...
+# `checkout_minimal`. The first call interacts with the network and can take
+# a long time, especially if the repo doesn't exist locally. The second call
+# is generally much quicker.
+#===================================
+sub new {
+#===================================
+    my ( $class, %args ) = @_;
+
+    $args{name} = 'target_repo';
+    my $self = shift->SUPER::new(%args);
```
olksdr: I think it is the same here `$class->SUPER`

lib/ES/TargetRepo.pm
```diff
@@ -0,0 +1,97 @@
...
+
+    $self->{destination} = $args{destination}
+        or die die "No <destination> specified for repo <target_repo>";
```
olksdr: like `die` I said `die`  -  I think one would be enough :)

lib/ES/BaseRepo.pm
```diff
@@ -0,0 +1,132 @@
...
+    my $reference_dir = 0;
+    if ($args{reference}) {
+        my $reference_subdir = $url;
+        $reference_subdir =~ s|/$||;
+        $reference_subdir =~ s|:*/*\.git$||;
+        $reference_subdir =~ s/.*[\/:]//g;
+        $reference_dir = $args{reference}->subdir("$reference_subdir.git");
+    }
+
+    bless {
```
olksdr: To be honest I would use explicit `return` instead of relying on implicit one

```perl
my $self = bless {
...

return $self;
```

nik9000: :+1:

lib/ES/Repo.pm
nik9000: `parent` look great! I'll switch to it

lib/ES/Repo.pm
nik9000: Yeah, it should be.

lib/ES/TargetRepo.pm
nik9000: :+1:

nik9000:
@olksdr, I pushed a fix for all of the great stuff you mentioned. Thanks again for `use parent`!

nik9000:
> I think it's good to go +1
> But of course service need to be tested if it does what must

Thanks!

We have quite a few integration tests that hit this particular behavior.

PR #832, opened 2019-04-23
Update APM meta tags

bmorelli25:
I've just learned that some of the fields in `conf.yaml` control _some_ of the meta tags for our documentation. I'm not an SEO expert. I have no idea how important this is, but we might as well update these to something/anything better.

Unfortunately, it looks like the meta description is shared by all documentation. It does not mention APM, Observability, or other newer products: https://github.com/elastic/docs/blob/fd91e0e31650751f787a61e84617cb2d30a6b97e/resources/website-l10n.xsl#L54

`conf.yaml`-`tags` --> `DC.type`
`conf.yaml`-`subject` --> `DC.subject`
`conf.yaml`-`version` --> `DC.identifier`
Screen Shot 2019-04-23 at 10 39 52 AM

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -904,8 +904,8 @@ contents:
             current:    7.0
             branches:   [ master, 7.0, 6.7, 6.6, 6.5, 6.4, 6.3, 6.2, 6.1, 6.0 ]
             chunk:      1
-            tags:       APM Server/Reference
-            subject:    APM
+            tags:       APM/Elastic APM/APM Overview/APM Getting Started/Install APM/Run APM
+            subject:    Elastic APM Overview
             asciidoctor: true
             sources:
               -
@@ -921,8 +921,8 @@ contents:
             current:    7.0
             branches:   [ master, 7.0, 6.7, 6.6, 6.5, 6.4, 6.3, 6.2, 6.1, 6.0 ]
             chunk:      1
-            tags:       APM Server/Reference
-            subject:    APM
+            tags:       APM Server/APM Reference/Elastic APM
+            subject:    Elastic APM Server Reference
             sources:
               -
                 repo:   apm-server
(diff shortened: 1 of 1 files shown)


conf.yaml
```diff
@@ -904,8 +904,8 @@ contents:
             current:    7.0
             branches:   [ master, 7.0, 6.7, 6.6, 6.5, 6.4, 6.3, 6.2, 6.1, 6.0 ]
             chunk:      1
-            tags:       APM Server/Reference
-            subject:    APM
+            tags:       APM/Elastic APM/APM Overview/APM Getting Started/Install APM/Run APM
+            subject:    Elastic APM Overview
```
lcawl: I poked around a bit just now and it seems like the "subject" is used in the "Product name" filter on our search results, so you might want to consider whether this particular value is appropriate here.

conf.yaml
```diff
@@ -921,8 +921,8 @@ contents:
             current:    7.0
             branches:   [ master, 7.0, 6.7, 6.6, 6.5, 6.4, 6.3, 6.2, 6.1, 6.0 ]
             chunk:      1
-            tags:       APM Server/Reference
-            subject:    APM
+            tags:       APM Server/APM Reference/Elastic APM
+            subject:    Elastic APM Server Reference
```
lcawl: Ditto

conf.yaml
bmorelli25: Ahh crap. That explains why every subject is so simple. Sounds like a better solution might be having a different yml key map to DC.subject. But that's not the easiest of fixes. OK, I'll play around a bit more. Might have to cancel this idea for now. Thanks for your expertise.

bmorelli25:
Sadly, this will cause more harm than good.

PR #834, opened 2019-04-23
Asciidoctor: Enforce more cops

nik9000:
Removes a few cop suppressions and fixes that failures.


nik9000:
Thanks for the review @estolfo!

PR #835, opened 2019-04-23
Switch curator docs to Asciidoctor

nik9000:
Switches the curator docs from the unmaintained AsciiDoc project to the
Asciidoctor project.


nik9000:
I ran a diff on the result and the changes are fairly minor:
1. In super old versions we change the url of a single page. For non-deprecated software I think this is a big deal, but for deprecated stuff like curator I think we're ok. Especially for old versions.
2. Many anchors change. Again, for deprecated stuff this is almost certainly fine.

This is pretty much it....

PR #838, opened 2019-04-24
Improve migration warnings

nik9000:
This improves support for `MIGRATION` warnings by allowing books to
suppress invididual warnings by setting an attribute. That looks like
this:
```
== Example
:migration-warning-override-snippet: false
[source,console,snippet=snippet.console]
----
GET /
----
```

This will allow books to disable these warnings until they are fixed.

To implement this I've migrated handling of the migration warnings from
`perl` to `ruby`.


nik9000:
@estolfo, could you have a look at this when you get a chance?

nik9000:

PR #840, opened 2019-04-25
Fix up-to-date check with asciidoctor

nik9000:
It turns out that we've had a broken up-to-date check for books build
with asciidoctor because we were accidentally passing the "this book is
built with asciidoctor" metadata into git when it was expecting just a
hash. This caused *any* change in a repo to force rebuilding every book
that uses that repo that is built with asciidoctor. Ooops.

The vast, vast, vast majority of this commit is building the
infrastructure needed to test it. That is because we don't have the
infrastructure to use rspec to test building all books more than once.


integtest/spec/helper/dest.rb
```diff
@@ -9,11 +9,17 @@
 class Dest
   include Sh
 
+  ##
+  # Stdout and stderr of running the conversions in convert_single
+  # or convert_all.
+  attr_reader :convert_outputs
```
nik9000: This allows me to access the output from the conversions without using the member variables that are oh-so-funky to use with rspec. Rather, we end up forwarding all access through a single member variable read that we have tight control over.

integtest/spec/helper/dest.rb
```diff
@@ -39,9 +45,13 @@ def convert_all(conf)
       --target_repo #{bare_repo}
       --conf #{conf}
     ]
-    out = run_convert cmd
+    @convert_outputs << run_convert(cmd)
+  end
+
+  ##
+  # Checks out the results of the last call to convert_all
+  def checkout_conversion
```
nik9000: The new tests run `convert_all` multiple times and only need to checkout after the last time.

integtest/spec/helper/dsl.rb
```diff
@@ -7,6 +7,34 @@
...
 # Defines methods to create contexts and shared examples used in the tests.
 module Dsl
+  def convert_before
+    include_context 'source and dest'
+    before(:context) do
+      yield @src, @dest
+    end
+  end
+
+  RSpec.shared_context 'source and dest' do
```
nik9000: It feels right to move this now.

integtest/spec/helper/dsl/convert_all.rb
```diff
@@ -9,31 +9,56 @@ module ConvertAll
     # 1. Create source repositories and write them
     # 2. Configure the books that should be built
     def convert_all_before_context
-      include_context 'source and dest'
-      before(:context) do
-        yield @src
-        @src.init_repos
-        @out = @dest.convert_all @src.conf
+      convert_before do |src, dest|
```
nik9000: I like how there are fewer member variable accesses when I do it this way!

integtest/spec/helper/dsl/convert_single.rb
```diff
@@ -9,21 +9,19 @@ module ConvertSingle
...
         # this conversion with the results of the `Asciidoctor` conversion.
-        @asciidoc_out = @dest.convert_single from, 'asciidoc',
-                                             asciidoctor: false
+        dest.convert_single from, 'asciidoc', asciidoctor: false
       end
       include_examples 'convert single'
     end
     shared_context 'convert single' do
-      let(:out) { @asciidoctor_out }
+      let(:out) { outputs[0] }
```
nik9000: Most users of this context won't know or care that there are is more than one output.

integtest/spec/multiple_builds_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+          end
+        end
+      end
+
+      context 'because there are no changes to the source repo' do
+        build_one_book_twice {}
+        include_examples 'second build is noop'
+      end
+
+      context 'because there are unrelated changes source repo' do
```
nik9000: This is the context that tests the perl change. After all that infrastucture the whole thing is 7 lines of ruby.

integtest/spec/spec_helper.rb
```diff
@@ -28,24 +28,6 @@
   config.include Sh
 end
 
-RSpec.shared_context 'source and dest' do
```
nik9000: I moved this into `dsl.rb`.

lib/ES/Repo.pm
```diff
@@ -52,34 +52,37 @@ sub has_changed {
     return 1 if exists $self->{sub_dirs}->{$branch};
 
     local $ENV{GIT_DIR} = $self->git_dir;
-    my $old = $self->_last_commit_info(@_);
+    my $old_info = $self->_last_commit_info(@_);
```
nik9000: And the actual fix it `-7+10`. Tests be like that sometimes.....

nik9000:
Thanks for reviewing @estolfo and @olksdr !

PR #841, opened 2019-04-26
Switch legacy shield docs to asciidoctor

nik9000:
Switches the legacy shield docs from the unmaintained AsciiDoc project
to the actively maintained Asciidoctor project.


nik9000:
Asciidoctor makes fairly minimal changes to the rendered docs:
1. A few auto-generated ids on the page change. For legacy documentation this isn't a big deal to me so long as we aren't actively linking to it.
2. Some sections change indentation slightly, usually moving "outward" slightly. These have different HTML but are barely visually distinct.

PR #842, opened 2019-04-26
Switch legacy marvel docs to asciidoctor

nik9000:
Switches the legacy marvel docs from the unmaintained AsciiDoc project
to the actively maintained Asciidoctor project.


nik9000:
For the most part asciidoctor's changes to the generated output are positive. It breaks an ordered list but given **how** legacy these docs are I'm ok with that.

PR #843, opened 2019-04-29
Switch build aliases script to use build_docs

nrichers:
There are many of us who use the build aliases script, but many of the aliases still seem to be using the about-to-be-deprecated `build_docs.pl` build script rather than the newfangled and preferred `build_docs`. This PR fixes that. Tested to work with the ESS and ECE docs.

@nik9000 and @lcawl I assume we want this change, unless the build aliases script is also about to be deprecated?

### DIFF
--- a/doc_build_aliases.sh
+++ b/doc_build_aliases.sh
@@ -12,31 +12,31 @@
 
 
 # Elasticsearch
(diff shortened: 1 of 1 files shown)


nik9000:
> An important thing: merging this will switch Elasticsearch's [PR builds](https://elasticsearch-ci.elastic.co/view/Elasticsearch/job/elastic+elasticsearch+pull-request+docbldesx/) to using the docker based build. This should be fine because it is pinned to a node that has docker.

Let me know when you merge this. I'll watch out to make sure the PR builds don't start failing. I'm 99% sure they'll be fine.

nik9000:
The PR builds I mentioned look good!

PR #850, opened 2019-04-30
Enable line length cop

nik9000:
Enables the line length cop on the asciidoctor customizations. I'd
entended to do this slowly while reworking other things but not having
this enabled has started getting in the way.

### DIFF
--- a/resources/asciidoctor/lib/elastic_include_tagged/extension.rb
+++ b/resources/asciidoctor/lib/elastic_include_tagged/extension.rb
@@ -103,21 +103,6 @@ class ElasticIncludeTagged < Asciidoctor::Extensions::IncludeProcessor
         @start_of_include = lineno
       end
     end
-<<<<<<< HEAD
-    if start_of_include.nil?
-      warn reader.cursor, "missing start tag [#{tag}]"
-      return path
-    end
-    if found_end == false
-      cursor = Asciidoctor::Reader::Cursor.new(
-        path, relpath, relpath, start_of_include
-      )
-      warn cursor, "missing end tag [#{tag}]"
-    end
-    reader.push_include included_lines, path, relpath, start_of_include, attrs
-  end
-=======
->>>>>>> master
 
     def warn(message, cursor = @reader.cursor)
       logger.warn message_with_context message, source_location: cursor
--- a/resources/asciidoctor/lib/elastic_include_tagged/extension.rb
+++ b/resources/asciidoctor/lib/elastic_include_tagged/extension.rb
@@ -8,7 +8,7 @@ require 'asciidoctor/extensions'
 #
 # Usage
 #
-#   include:elastic-include-tagged:{doc-tests}/RestClientDocumentation.java[rest-client-init]
+#   include:elastic-include-tagged:{doc-tests}/Foo.java[tag]
 #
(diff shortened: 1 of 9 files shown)


resources/asciidoctor/spec/copy_images_spec.rb
```diff
@@ -532,8 +541,11 @@ def copy_attributes(copied)
       %w[coming note],
       %w[deprecated warning],
   ].each do |(name, admonition)|
-    it "copies images for the block formatted #{name} change admonition when requested" do
+    it "copies images for the block formatted #{name} change admonition" \
+        " when requested" do
       copied = []
+      relative_path = 'images/icons'
```
nik9000: I expect to remove this repetition in a follow up change where I switch this to more standard rspec.

resources/asciidoctor/spec/elastic_compat_preprocessor_spec.rb
```diff
@@ -245,31 +245,21 @@
...
-    expected = <<~DOCBOOK
-      <chapter id="_header">
-      <title>Header</title>
-      <note revisionflag="added" revision="some_version">
-      <simpara></simpara>
-      </note>
-      </chapter>
-      <chapter id="test">
-      <title>Header</title>
+  context 'when a block contains things other than attributes' do
```
nik9000: I reworked this one rather than just chopping the line because I'd forgotten the goal.

resources/asciidoctor/spec/edit_me_spec.rb
```diff
@@ -81,7 +82,8 @@ def spec_dir_link(file)
       end
     end
 
-    shared_examples 'standard document part' do |type, title_start = '<title>', title_end = '</title>'|
+    shared_examples 'standard document part' \
+        do |type, title_start = '<title>', title_end = '</title>'|
```
nik9000: This one kind of bothers me but it is the way that make rubocop happy.

resources/asciidoctor/lib/elastic_include_tagged/extension.rb
```diff
@@ -28,7 +28,8 @@ def process(_doc, reader, target, attrs)
     start_match = /^(\s*).+tag::#{tag}\b/
     end_match = /end::#{tag}\b/
 
-    path, target_type, relpath = reader.resolve_include_path target, attrs, attrs
+    path, target_type, relpath =
+      reader.resolve_include_path target, attrs, attrs
```
estolfo: You might as well put each assignment on a separate line, as using parallel assignment isn't as clear when it extends to >1 lines.

resources/asciidoctor/spec/edit_me_spec.rb
estolfo: looks a bit strange to me. Maybe assign `title_start` and title_end` in the block below.

resources/asciidoctor/spec/edit_me_spec.rb
```diff
@@ -115,7 +117,8 @@ def spec_dir_link(file)
...
```
estolfo: same here, I think this could be a bit more readable.

nik9000:
@estolfo, I've acted on your recommendations and merge in master. Could you have another look when you get a chance? Thanks so much for your reviews!

nik9000:

PR #854, opened 2019-04-30
Tests: Drop redundant returns

nik9000:
Enables and fixes the redundant return cop. In one case I was able to
remove the entire method.


nik9000:
@estolfo, could you have a look at this one too please?

nik9000:
Thanks again for reviewing @estolfo!

PR #855, opened 2019-04-30
Drop template.md5 file

nik9000:
Now that we're not automatically fetching the template there isn't a
nead to store its md5 with every book. Instead we leave a note to users
that they'll have to manually force a rebuild after updating the
template. This saves us a few files which is nice but it allows us to
significantly simplify the code around up-to-date checking.

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -363,7 +363,7 @@ sub build_entries {
             temp_dir => $temp_dir,
             %$entry
         );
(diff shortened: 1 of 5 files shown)


nik9000:
@olksdr, could you have a look at this one? It is all perl!

nik9000:
Thanks for reviewing @olksdr!

PR #856, opened 2019-05-01
Add basic test for docs.js

nik9000:
Adds a single, super basic test for docs.js using nodejs, yarn, and
jest. The majority of the complexity here is in getting the nodejs
environment set up in the docker container. We do it similarly to how we
set up ruby:
* Install nodejs with that package manager
* Use a nodejs specific tool to fetch dependencies: yarn
* Lock the dependencies with a lockfile to prevent them from wandering
around when we rebuild the docker image

I chose not to use nvm here because yarn wanted to be installed using
apt and it felt right to install nodejs the same way.

I use the same version of jquery that we use on the public site even if
it is super old.


nik9000:
@jarpy, could you have a look at the Docker stuff for this one?

resources/web/__tests__/docs-test.js
```diff
@@ -0,0 +1,99 @@
...
+          -d'
+          {
+          "user" : "kimchy",
+          "post_date" : "2009-11-15T14:12:12",
+          "message" : "trying out Elasticsearch"
+          }
+          '
+        `));
+      });
+      it('has a trailing newline', () => {
```
cjcenizal: Is this assertion a duplicate of the one on line 49?

resources/web/__tests__/docs-test.js
```diff
@@ -0,0 +1,99 @@
...
+      }
+    `
+    pageWithConsole('a snippet with a body', withBody, () => {
+      it('includes the method', () => {
+        expect(document.copied).toMatch(/-X PUT/)
+      });
+      it('includes the url', () => {
+        expect(document.copied).toMatch(/"localhost:9200\/twitter\/_doc\/1"/);
+      });
+      it('include the body', () => {
```
cjcenizal: Nit: "include" -> "includes"

resources/web/__tests__/docs-test.js
```diff
@@ -0,0 +1,99 @@
...
+          </div>
+          <div class="console_widget" data-snippet="snippets/getting-started-cluster-health/1.json"></div>
+        </div>
+      `;
+
+      docs.init_console_widgets('localhost:5601/app/kibana#/dev_tools/console');
+      copyAsCurl = jQuery('.copy_as_curl');
+    });
+
+    it('to exist', () => {
```
cjcenizal: Nit: this could just be "exists"

nik9000:
Thanks for reviewing @jarpy and @cjcenizal!

PR #860, opened 2019-05-02
Fix build when pointed at a worktree

### LINKED ISSUE #833: --doc doesn't like worktrees
It looks like when you point `./build_docs --doc` to a worktree it'll fail to find the root of the repo and fall over. It really should just work. At a minimum it shouldn't crash.

nik9000:
When you do `build_docs --doc ../foo/index.asciidoc` and `foo` is a git
worktree then we blew up. This prevents that.

Closes #833


nik9000:
Thanks for looking @olksdr and @jdconrad !

PR #864, opened 2019-05-06
Fix the location of "on this page"

nik9000:
When I made docs.js more testable I accidentally moved it. Tests coming
in a followup change but I'd like to get this fix out now.

Relates to https://github.com/elastic/cloud/issues/32260


resources/web/docs.js
```diff
@@ -637,6 +638,7 @@
 
   function init_headers() {
     // Add on-this-page block
+    var this_page = jQuery('<div id="this_page"></div>').appendTo(right_col);
```
cjcenizal: Do we want to append this element to `right_col`, or insert it somewhere near the top or in the middle of the `right_col` element's children? I ask because in the linked issue, it looks like the "On the page" block appears above the table of contents:

image

And when I tested locally, it appears below the table contents, which pretty much buries it in terms of user experience:

image

nik9000:
I thought it was intended to be below. I'll check on that one!

nik9000:
@cjcenizal, I've moved it to the top. I think the old code had it there too but I switched it when fixing it. I pushed a patch to switch it back.

PR #867, opened 2019-05-06
Fix up-to-date check when switching asciidoctor-ness

nik9000:
When I added support for asciidoctor I mistakenly caused all books built
by asciidoctor to be considered out of date when there was *any* change
made to any repo that builds them, even if the changes were to unrelated
files.

When I fixed that I mistakenly caused remove asciidoctor-ness from
consideration in the up-to-date check entirely. Programing is like that,
I guess. This fixes that by explicitly checking if asciidoctor-ness
changed.

### DIFF
--- a/lib/ES/Repo.pm
+++ b/lib/ES/Repo.pm
@@ -77,7 +77,7 @@ sub has_changed {
 
     return $old_info ne $new_info if $self->{keep_hash};
     return if $old_info eq $new_info;
-    # If the asciidoctor-ness of the previous build dosen't match this one then
+    # If the asciidoctor-ness of the previous build doesn't match this one then
     # we've changed. It'd be nice if build-info were a class but we'll be
     # dropping asciidoctor as soon as we've migrated all of the books and this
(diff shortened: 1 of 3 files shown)


nik9000:
@olksdr, could you have a look at this one?

nik9000:
The perl is only one line and the tests are more, but they are in the same vein as before.

lib/ES/Repo.pm
```diff
@@ -77,6 +77,11 @@ sub has_changed {
 
     return $old_info ne $new_info if $self->{keep_hash};
     return if $old_info eq $new_info;
+    # If the asciidoctor-ness of the previous build dosen't match this one then
```
olksdr: dosen't => doesn't

nik9000:
> there is a small comment fix and otherwise lgmt

Thanks!

nik9000:
Thanks so much for reviewing @olksdr!

nik9000:
Merged! Thanks for reviewing @olksdr!

PR #868, opened 2019-05-06
Integtest: rename change detection spec

nik9000:
One of our spec files is entirely about change detection so I think it
should *look* like that.


integtest/spec/all_books_change_detection_spec.rb
```diff
@@ -0,0 +1,275 @@
+# frozen_string_literal: true
+
+##
+# Assertions about when books are rebuilt based on changes in source
+# repositories or the book's configuration.
+RSpec.describe 'building all books' do
+  describe 'change detection' do
```
nik9000: I really just renamed the file, added this line and indented everyone once.

estolfo: yeah, that works!

nik9000:
Super :heart:

PR #873, opened 2019-05-07
Add test for 'On This Page' section

nik9000:
Adds a test for the 'On This Page' section which I've previously broken
accidentally. This test would have caught such breakage.

Relates to #864


nik9000:
@elasticjeff, could you have a look at this one as well?

PR #877, opened 2019-05-08
Support --target_branch

nik9000:
Adds support for specifying the branch to which to build docs in the
target repo. This will make release times smoother because we can
continue to build the docs but send them to the `staging` branch. It
*might* allow more fancy pull request testing by allowing us to send
pull requests to their own branch.

If the branch already exists we just check it out and build the docs
onto it and then push it. If it doesn't exist then things are a little
more complicated. First we check if the `master` branch exists. It won't
in an entirely fresh repo which we often use for testing. If it doesn't
exist we create an empty commit in the master branch to "anchor" it.
Once we're sure that the master branch exists we fork the new target
branch from the master branch.

Pushing the built docs gets a bit more complex in two ways:
1. If the target branch is new we always push it, even if there haven't
been any changes. This keeps things consistent, both for easy testing
and for the potential PR testing work that I mentioned above.
2. If we had to create an empty master branch we push that too. This
makes sure that non-empty repos have a master branch which is a fairly
useful invariant.

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -555,6 +555,11 @@ sub push_changes {
     $build_dir->file('revision.txt')
         ->spew( iomode => '>:utf8', ES::Repo->all_repo_branches );
 
+    # Check if there are changes before committing and save that because if
+    # there are changes we'll commit them. Once we've committed them checking
+    # if there are changes will say "no" because there aren't changes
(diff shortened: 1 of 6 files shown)


nik9000:

PR #884, opened 2019-05-10
Only mount and clean the /run/nginx directory

Crazybus:
On Sles-12 and potentially other distros some extra stuff is
automatically mounted into `/run` making it impossible to remove them
and ultimately causing the build to fail as seen in https://elasticsearch-ci.elastic.co/job/elastic+elasticsearch+pull-request+docbldesx/12978/console

```
/ # mount | grep /run
tmpfs on /run/secrets/credentials.d type tmpfs (ro,relatime)
tmpfs on /run/secrets/SUSEConnect type tmpfs (ro,relatime)
```

However it looks like only `/run/nginx` is actually being used so there
doesn't seen to be a need to mount and wipe the whole of `/run`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -64,7 +64,7 @@ RUN install_packages \
 
 # We mount these directories with tmpfs so we can write to them so they
 # have to be empty. So we delete them.
-RUN rm -rf /var/log/nginx && rm -rf /run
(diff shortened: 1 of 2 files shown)


nik9000:
Lgtm. Sorry for the mess!

nik9000:
Thanks for the fix @Crazybus! Sorry I didn't merge it when you proposed it.

PR #887, opened 2019-05-14
New value for {searchprofiler} attribute

schersh:
The {searchprofiler} attribute prints as "Query Profiler" in the docs, but the name of this tool in the UI is "Search Profiler". Suggesting that we replace the value for this attribute to match what is in the UI. 

Screen Shot 2019-05-14 at 3 26 11 PM


nik9000:
I'm going to close this as part of some (very late) spring cleaning.

PR #891, opened 2019-05-21
Set noindex on the legacy docs to block Google from indexing

debadair:
This will prevent folks from getting directed to the obsolete docs over the current docs. 

This does not set noindex on the Definitive Guide. We'll wait to do that until we've incorporated equivalent content into the ES docs.

@jonasll Need to verify that this will not prevent our site search from indexing these pages. Since we already set noindex on the Logstash Versioned Plugins doc, I think we're ok?

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -852,6 +852,7 @@ contents:
             current:    1.3
             branches:   [ 1.3, 1.2, 1.1, 1.0.1 ]
             chunk:      1
+            noindex:    1
             tags:       Legacy/Topbeat/Reference
             subject:    Topbeat
             sources:
@@ -1257,6 +1258,7 @@ contents:
             title:      X-Pack Reference for 6.0-6.2 and 5.x
             prefix:     en/x-pack
             chunk:      1
+            noindex:    1
             tags:       Legacy/XPack/Reference
             current:    6.2
             subject:    X-Pack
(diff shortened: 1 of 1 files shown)


jonasll:
@debadair what does the final outputed tag look like?

lcawl:
@jonasll This is an example of a page that has it already: https://www.elastic.co/guide/en/logstash-versioned-plugins/current/input-plugins.html

`<meta name="robots" content="noindex,nofollow" />`

debadair:
As I mentioned in https://github.com/elastic/search-www.elastic.co/issues/1, after looking at the results for a few key search terms, I'm totally okay if this change means those pages won't get surfaced in our site search, either. They are still surfaced through hits on our doc landing page, and accessible through our doc nav for folks who still need them.

jonasll:
This tag will prevent our crawler to index those pages. A Swiftype-specific tag can be added to command the crawler to proceed:
 ```
<meta name="robots" content="noindex, nofollow">
<meta name="st:robots" content="follow, index">
```

debadair:
Thanks for the info @jonasll! It's great to know that we have a way to override the noindex tag for our own search if we need to. For now, our thinking is that we don't need to in this case, so I'm going to go ahead and merge this PR as-is. We'll keep an eye on the impact & keep this in mind as we're talking about tuning the doc search results in general.

PR #899, opened 2019-05-24
Only push to new target branches if there changes

nik9000:
Before this commit if you add `--target_branch` and the branch is *new*
then we pushed to it regardless of if there were changes. This felt nice
and consistent but we're going to end up with a lot of branches that
way.


nik9000:
Thanks @olksdr !

PR #904, opened 2019-05-29
Create a docker image for previewing

nik9000:
This modifies build_docs.pl to support `--preview` which serves preview of
any branch of the `--target_repo`. It works by forking nginx to handle
the incoming traffic, pulling the target repo locally, and then forking
a nodejs process that can read directly from the git repo without
checking it out. You can test this locally with `./preview/test.sh`.

It picks the branch to preview by grabbing the first "chunk" of the
requested host. So if you run `./preview/test.sh` you should be able to
go to `master.localhost:8000/guide/index.html` to see the docs from the
master branch. When you first run `./preview/test.sh` it'll 503 but
*eventually* we'll have cloned the branch locally and we'll be able to
serve it.

`./preview/test.sh` works by booting building a docker image for the
preview and running it locally. This mirrors the way we intend to run
this in production with k8s. The idea is that k8s can deploy the docker
image as a pod and run `GET` `localhost:8000` for a "liveness" check
and `GET` `master.localhost:8000/guide/index.html` for its readiness
check. k8s will be responsible for configuring an ingress from a
convenient url, say `http://*.docs-preview.app.elstc.co/guide/index.html`,
that resolves to port 8000 on the node. Folks can replace the `*` with
the branch that they want to see.

This combines nicely with a recent change we made to the pull request
configuration which pushes the results of pull request to the
`{repo_name}_{pr_number}` branch. So you can see a preview of you pull
request, at, for example,
`http://stack-docs_332.docs-preview.app.elstc.co/guide/index.html`.

I decided on using a separate process to serve the pages instead of
using a git work tree because it saves a huge amount of disk space and
makes the process of receiving new branches much faster. It seems like
the right tradeoff because I expect we'll rarely need *all* of the files
in the preview. And it is fast enough. Like 20ms. And! It gives us the
ability to do fancy stuff with git on the way in like add the commit
hash or the message to in a header. We don't do that now, but we can.

I decided not to expose `--preview` in the `build_docs` python script
that folks run to build docs. Instead, the only way to run this is with
the docker image that it builds. This feels like the realest sort of
testing.

I include some fairly bare-bones testing in
`integtest/spec/preview_spec.rb` that mostly just verifies that we
successfully serve both html and images and appropriately pick up
changes to the `--target_repo`, adding branches when they are created
and deleting them when they are removed.

Left to do is to handle redirects. Every branch contains its nginx
config file containing redirects. When we test with `--all --open`
locally we *do* handle that redirects file but this preview *doesn't*
yet handle it. That feels complex enough that it is worth saving for a
follow up change.

### DIFF
--- a/lib/ES/Util.pm
+++ b/lib/ES/Util.pm
@@ -23,6 +23,7 @@ our @EXPORT_OK = qw(
     write_html_redirect
     write_nginx_redirects
     write_nginx_test_config
+    write_nginx_preview_config
 );
 
 our $Opts = { procs => 3, lang => 'en' };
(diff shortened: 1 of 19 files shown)


nik9000:
I've pinged a bunch of folks for reviews:
* @estolfo for rspec
* @jarpy for docker/k8s stuff
* @haywoood for the js
* @olksdr for the perl
* @tylerjl because we talked about this a few weeks ago

nik9000:
>  ** [jarpy](/jarpy) ** approved these changes [ 38 minutes ago ](#pullrequestreview-243567584)

Thanks! If you can think of anything that'd make it better let me know!

lib/ES/Util.pm
```diff
@@ -568,6 +569,66 @@ CONF
...
+#===================================
+sub write_nginx_preview_config {
+#===================================
+    my ( $dest ) = @_;
+
+    # TODO pull redirects from branches
+
+    # We log the X-Opaque-Id which is a header that Elasticsearch uses to mark
+    # requests with an id that is opaque to Elasticsearch. Presumably this is
+    # a standard. Either way we follow along. We us it in our tests so we can
```
olksdr: `we us it ..` -> `we use it ..`

nik9000: :heart:

preview/preview.js
```diff
@@ -0,0 +1,87 @@
...
+};
+const catOpts = {
+  'cwd': '/docs_build/.repos/target_repo.git',
+  'encoding': 'buffer',
+  'max_buffer': 1024 * 1024 * 20,
+};
+
+const requestHandler = (request, response) => {
+  parsedUrl = url.parse(request.url);
+  if (false == parsedUrl.pathname.startsWith('/guide')) {
```
haywoood: nit: `!parsedUrl`. Also, always default to strict equals aka `===` in JS, as `==` is a bit looser e.g. `10 == '10'` => true

nik9000:
I think I've received sufficient review at this point. Thanks all!

PR #907, opened 2019-05-31
Move `--open` tests from Makefile to rspec

nik9000:
This generalizes the infrastructure for testing the docs preview so that
it can test `--open` as well. Then it ports the test for `--all --open`
from `integtest/Makefile` to rspec and adds a test for `--doc --open`.
all this in an effort to remove tests from `integtest/Makefile` so they
are more readable.


nik9000:
@estolfo, could you have a look at this when you get a chance?

PR #911, opened 2019-06-03
Convert remaining change detection tests to rspec

nik9000:
This drops all of the remaining tests related to change detection from
`integtest/Makefile` and adds them to
`integtest/spec/all_books_change_detection_spec.rb` which should make
them *much* easier to read.


nik9000:
@estolfo, please have a look at this one when you get a chance!

integtest/spec/all_books_change_detection_spec.rb
```diff
@@ -45,15 +55,37 @@ def self.build_one_book_out_of_one_repo_twice(
...
 
           # Allow the caller to customize the source
           before_first_build.call src, config
         end,
         before_second_build: before_second_build
       )
+      include_context 'build one book twice'
+    end
+
+    def self.build_one_book_out_of_one_repo_and_then_out_of_two(
```
estolfo: I'm assuming rubocop has already warned you about this method name length...?

nik9000: It didn't! It is certainly really long though....

PR #915, opened 2019-06-04
Remove edit link from links to bridgeheads

### LINKED ISSUE #903: Extra problematic edit link
There's an extra “edit” link at the very top of this file: https://www.elastic.co/guide/en/elasticsearch/reference/master/ml-put-job.html

image

If you click it, it takes you to an inappropriate location.  If I click the next lower link, it takes me to the right place (https://raw.githubusercontent.com/elastic/elasticsearch/master/docs/reference/ml/apis/put-job.asciidoc).

The pages before and after this one don’t seem to have the problematic edit link in the title section.  There are no overrides specified for this file (it’s included here: https://raw.githubusercontent.com/elastic/elasticsearch/master/docs/reference/ml/apis/ml-api.asciidoc).

NOTE: The same problem occurs when you use the --asciidoctor build option too.

nik9000:
We had a bug where links to floating titles (aka bridgeheads) would
include the `edit_me` link in its contents like this:

```
<a class="xref" href="chapter.html#floater" title="Floateredit">
  Floater
  <a href="https://github.com/elastic/docs/edit/master/index.asciidoc" class="edit_me" title="Edit this page on GitHub" rel="nofollow">
    edit
  </a>
</a>
```

Because this is invalid html it'd cause the `edit` button to float to
the top of the page, and, confusingly, take you to edit the source of
the *link target* rather than the current page. Madness.

This removes the extra link.

Closes #903

### DIFF
--- a/integtest/spec/single_book_spec.rb
+++ b/integtest/spec/single_book_spec.rb
@@ -145,7 +145,7 @@ RSpec.describe 'building a single book' do
         end
       end
     end
-    context 'for a book that a reference to a floating title' do
+    context 'for a book that has a reference to a floating title' do
       convert_single_before_context do |src|
         src.write 'index.asciidoc', <<~ASCIIDOC
           #{HEADER}
--- a/integtest/spec/single_book_spec.rb
+++ b/integtest/spec/single_book_spec.rb
@@ -61,7 +61,7 @@ RSpec.describe 'building a single book' do
     page_context 'chapter.html' do
       it 'has an "unknown" edit url' do
         expect(body).to include(<<~HTML.strip)
-          <a href="unknown/edit/master/index.asciidoc" class="edit_me"
+          <a href="unknown/edit/master/index.asciidoc" class="edit_me" title="Edit this page on GitHub" rel="nofollow">edit</a>
         HTML
       end
     end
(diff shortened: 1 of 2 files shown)


nik9000:
@lcawl, could you check this one when you have a chance? I don't think we really have a code reviewer for the xslt bit of the change. It is sort of a evolutionary dead end. Especially for working with such large bits of xml.

We'll have to manually rebuild all the docs after merging this to fix all of the funny edit links.

integtest/spec/single_book_spec.rb
```diff
@@ -145,6 +145,25 @@
         end
       end
     end
+    context 'for a book that a reference to a floating title' do
```
lcawl:
```suggestion
    context 'for a book that has a reference to a floating title' do
```

lcawl:
@alaudazzi Found what seem to be similar problems in https://www.elastic.co/guide/en/ecs/current/ecs-guidelines.html and https://www.elastic.co/guide/en/elastic-stack-overview/current/how-watcher-works.html

nik9000:
I've merged this. I'll kick off a full rebuild of the docs later on. I can't do it right now because we're upgrading jenkins and it is in "shutting down" mode and won't take new jobs.

PR #919, opened 2019-06-05
Create branch cleaner for preview

nik9000:
Creates a fairly simple script that removes the branches created by pull
request builds when the pull requests have closed. The plan is to
schedule this once a day or something in jenkins so it can keep the
`build-docs` repo from growing too large.

The docs preview service will notice the deleted branches and clean them
up on the ma


nik9000:
I think I should give this a shot in a local CI instance before I remove "draft" from it.

nik9000:
@haywoood, could you have a look at this one?

preview/__test__/clean.test.js
```diff
@@ -0,0 +1,156 @@
...
+        pullRequest: {
+          closed: is_closed
+        }
+      }
+    }
+  });
+}
+
+afterAll(() => {
+  return new Promise((resolve, reject) => {
```
haywoood: Any reason to not use async/await in the places you're using `new Promise`? If it's just to catch an error the equivalent would be wrapping the await in a `try {} catch {}`

preview/clean.js
```diff
@@ -0,0 +1,181 @@
...
+const dedent = require('dedent');
+const https = require('https');
+const os = require('os');
+const path = require('path');
+
+module.exports = Cleaner;
+module.exports.exec_git = exec_git;
+
+function Cleaner(token, repo, cache_dir, tmp_dir) {
+  this.repo = repo;
```
haywoood: Doesn't really make a difference, but just wanted to note that you don't really need to use `this` with `new Cleaner`.

The function arguments are captured in the closure, so you could define the functions inside the body, and return an object at the end which would have the same effect. Possibly even better, move the functions out of the body, and modify their definitions so you call them with the data they need.


```Javascript
const clone = (localPath, cacheDir, repoName) =>
  exec_git([
      'clone',
      '--bare',
      '--reference', `${cacheDir}/${repoName}`,
      repo, localPath])

const getRepoName = repo => {
  let repo_name = path.basename(repo)
  if (!repo_name.endsWith('.git')) {
    repo_name += '.git';
  }
  return repoName
}

export const Cleaner = (token, repo, cache_dir, tmp_dir) => {
  const repoName = getRepoName(repo)
  const localPath = `${tmp_dir}/${repoName}`
  
  const showHeads = () => exec_git(['show-ref', '--heads'], {cwd: localPath})

  ;; snipped

  return {
    run: () =>
      clone(localPath, cache_dir, repoName)
        .then(show_heads)
        .then(heads_to_prs)
        .then(cleanup_closed_prs),
    showHeads: showHeads ;; just an example of exporting a function that's leveraging data int the closure.
    
  }
}
```

It could make things a little easier to test if you made the inner functions standalone while keeping the ease-of-use API the same.

This comment isn't an action item, just wanted to bring up this strategy.

preview/clean.js
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  this.show_heads = () => {
+    return exec_git(['show-ref', '--heads'], {cwd: this.local_path});
+  }
+
+  this.heads_to_prs = (heads) => {
+    return heads
+      .split('\n')
+      .map(line => {
```
haywoood: Would reduce be better here? I'm assuming the subsequent `filter` is to remove random `undefined`s generated by this map. That way you can conditionally push items on to the accumulator.

preview/clean.js
```diff
@@ -0,0 +1,181 @@
+'use strict';
+
+const child_process = require('child_process');
+const dedent = require('dedent');
+const https = require('https');
+const os = require('os');
+const path = require('path');
+
+module.exports = Cleaner;
+module.exports.exec_git = exec_git;
```
haywoood: Might be slightly better to export an object and destructure it when requiring vs attaching `exec_git` to the Cleaner function.

```Javascript
module.exports = { Cleaner, exec_git }

// elsewhere
const { Cleaner, exec_git } = require("../clean.js")
```

But not sure how important being able to call `Cleaner.exec_git` is. ES6 modules are supported in Node.js 12 not sure if we're using that, but it would allow you to have your cake and eat it too.

preview/__test__/clean.test.js
nik9000: I talked to @haywoood about this. There isn't actually a promise oriented API here. I'm going to see if I can switch to `rmfr` which has one.

nik9000:

PR #923, opened 2019-06-11
Add Azure Marketplace 7.1 to docs

russcam:
This commit adds Azure Marketplace 7.1 to the docs and also updates the title to make it more succinct

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -149,11 +149,11 @@ contents:
                 repo:   ecs
                 path:   docs/
           -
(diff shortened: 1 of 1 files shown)


russcam:
I'm waiting on 7.1 approval for the Azure Marketplace, which should be happening shortly. Once this comes through, this PR can be merged, if we're happy with it.

russcam:
Merging this in as new template is now live on the Azure Marketplace

PR #928, opened 2019-06-13
Remove legacy saas-release from ESS doc builds

nrichers:
Last year, we switched to the `release-ms-nn` doc builds for `current`. At the time, we decided to keep `saas-release` around for a while to prevent breaking bookmarks and links for people; a header encourages people to move to current docs. We've now been operating under the new milestone release builds for some time and I think it's time to remove the old legacy `saas-release` docs after giving folks a heads-up in the next weekly update. 

Just today, I looked at a recent doc issue that referenced some docs from `saas-release`, but at this point they are at least six or seven months out of date. I think it's time to rip off the bandaid.

@elastic/cloud-writers please approve--we'll discuss tomorrow as well.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -498,7 +498,7 @@ contents:
             tags:       Cloud/Reference
             subject:    Elastic Cloud
             current:    release-ms-21
-            branches:   [ release-ms-21, saas-release, saas-release-heroku ]
+            branches:   [ release-ms-21, saas-release-heroku ]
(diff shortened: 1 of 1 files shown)


nrichers:
We now have a wildcard redirect in place for these old docs via https://github.com/elastic/website-www.elastic.co/issues/4117. I think it's safe to merge this PR and get rid of the old docs.

PR #929, opened 2019-06-17
Move master branch to appear as current

alaudazzi:
Change the order of branch appearance, to have `master` as the current branch.


conf.yaml
```diff
@@ -536,7 +536,7 @@ contents:
             tags:       Kubernetes/Reference
             subject:    ECK
             current:    0.8
```
lcawl:
```suggestion
            current:    master
```

Issue #937, opened 2019-06-19
[Meta] Create API reference templates


KOTungseth:
In the future, all APIs deserve reference material. For customers to effectively use the APIs, they need reference material.

- [x] Choose a place for a API reference template to exist, ideally a public location (`Docs` repo, `shared` folder)

- [x] Choose best practices from existing or other content

Good API reference page has:
        * Title
        * Short description/Intro
        * Endpoint/Request
        * Description (more detailed)
        * Endpoint/Request
        * Path parameters
        * Query parameters
        * Request body
        * Response body
        
Some APIs also require:
        * Authorization/security details
        * Definition lists
        * Examples
        * Default values

- [x] Draft template for API page
      * Something that is useful, helpful to customers
      * Can be generated using tools that we have
      * Not too much glitter 

- [x] Draft template for definition page

Tests:
- https://github.com/elastic/elasticsearch/pull/43402
- https://github.com/elastic/elasticsearch/pull/43427
- https://github.com/elastic/elasticsearch/pull/43459
- https://github.com/elastic/elasticsearch/pull/43503
- https://github.com/elastic/elasticsearch/pull/43546
- https://github.com/elastic/elasticsearch/pull/43610
- https://github.com/elastic/elasticsearch/pull/43711
- https://github.com/elastic/elasticsearch/pull/43948
- https://github.com/elastic/elasticsearch/pull/43952

lcawl:
The formatting of the parameters in the API is still a little problematic. In particular, the "Required/Optional" is unnecessarily eye-catching in my opinion. For example:
![image](https://user-images.githubusercontent.com/26471269/60917410-c1d64900-a245-11e9-8095-31b1a058272c.png)
I did a quick test of some of the other options we discussed. For example:
![image](https://user-images.githubusercontent.com/26471269/60917617-4032eb00-a246-11e9-9e15-6f6941dc12cf.png)

jrodewig:
Here's my order of preference:

1.  `allow_no_match`
2. `size`
3. `from`

I believe the `from` example introduces visual clutter with the bulleted list. It will also have more overhead for Asciidoc maintenance.

The `size` example is okay, but I find the construction a little awkward. It's also not very clear to non-writers that they should always include "Optional" or "Required" as the first word of the definition.

I don't feel strongly though. Any of these options are fine with me.

lcawl:
I agree. I'll draft that change in the template unless @kotungseth disagrees.

lcawl:
[...]

PR #940, opened 2019-06-19
Drafts a template for API object definitions

lcawl:
Related to https://github.com/elastic/docs/issues/937

This PR creates a template for the pages that contain definitions for API objects. Similar to pages like https://www.elastic.co/guide/en/cloud-enterprise/current/definitions.html#AllocatorBuildInfo and https://www.elastic.co/guide/en/elasticsearch/reference/master/ml-job-resource.html

### DIFF
--- a/shared/api-definitions-ex.asciidoc
+++ b/shared/api-definitions-ex.asciidoc
@@ -6,9 +6,13 @@
 // ***************************************
 // These pages complement API reference pages. They provide details about the
 // objects in API request bodies or response bodies.
+// They do not need to map 1:1 to API response bodies unless these pages are
+// automatically generated. 
 // These pages are typically nested under an API definitions page.
-// For example, see
+// For an example of definitions that map specific objects to APIs, see:
 // https://www.elastic.co/guide/en/cloud-enterprise/current/definitions.html
+// For an example of definitions that apply to multiple APIs, see:
+// https://www.elastic.co/guide/en/elasticsearch/reference/master/api-definitions.html
 // ***************************************
 
 //Provide a brief description
--- a/shared/api-definitions-ex.asciidoc
+++ b/shared/api-definitions-ex.asciidoc
@@ -41,3 +41,30 @@ For example:
   lowercase alphanumeric characters (a-z and 0-9), hyphens, and underscores. It
   must start and end with alphanumeric characters.
 ////
+
+[float]
+[[sample-object-example]]
+==== Example
+// Optional. Be aware that if you add examples they need to be kept up-to-date.
+
+////
+[source,js]
+----
+{
+      "job_id": "total-requests",
+      "analysis_config": {
+        "bucket_span": "10m",
+        "detectors": [
+          {
+            "detector_description": "Sum of total",
+            "function": "sum",
+            "field_name": "total",
+            "detector_index": 0
+          }
+        ],
+        "influencers": [ ]
+      },
+      ...
+    }
+----
+////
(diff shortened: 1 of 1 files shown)


shared/api-definitions-ex.asciidoc
```diff
@@ -0,0 +1,30 @@
...
+[float]
+==== Properties
+
+////
+For example:
+
+`analysis_config`::
+  (object) The analysis configuration, which specifies how to analyze the data.
+  See <<ml-analysisconfig, analysis configuration objects>>.
+////
```
jrodewig: Let's include an example, similar to how it's handled in the Cloud Docs: https://www.elastic.co/guide/en/cloud-enterprise/current/definitions.html#AllocatedInstancePlansInfo

```suggestion
////

==== Example
---
{
   "pending" : true,
   "version" : "string",
   "zone_count" : 0
}
---
// NOTCONSOLE
```

shared/api-definitions-ex.asciidoc
```diff
@@ -0,0 +1,30 @@
...
+// ***************************************
+// * Use a definition list.
+// * Each property should be marked as Optional or Required.
+// * Include the data type.
+// * Include default values as the last sentence of the first paragraph.
+// * Include a range of valid values, if applicable.
+// * For nested objects, link to a separate definition list.
+// ***************************************
+
+[float]
```
jrodewig: Adding an anchor

```suggestion
[float]
[[sample-object-properties]]
```

shared/api-definitions-ex.asciidoc
```diff
@@ -0,0 +1,30 @@
+[[sample-object]]
+=== Object1 API objects
```
jrodewig:
```suggestion
=== `Object1` API objects
```

shared/api-definitions-ex.asciidoc
```diff
@@ -0,0 +1,30 @@
...
+++++
+<titleabbrev>Object1</titleabbrev>
```
jrodewig:
```suggestion
=== <titleabbrev>`Object1`</titleabbrev>
```

shared/api-definitions-ex.asciidoc
lcawl: I'm not sure how I feel about having formatting in the titles/headings.  I don't think I've seen this other than in the Cloud example.

shared/api-definitions-ex.asciidoc
lcawl: I think if we add examples like this, they'll get out-dated quickly unless we add ways to test them. That's the content that gets out-dated fastest on the APIs I've worked on.

shared/api-definitions-ex.asciidoc
```diff
@@ -0,0 +1,70 @@
...
+// ***************************************
+
+//Provide a brief description
+
+Is a cool thing.
+
+// Guidelines for API object documentation
+// ***************************************
+// * Use a definition list.
+// * Each property should be marked as Optional or Required.
```
lcawl: I think we shouldn't put "Optional" or "Required" here, since it will depend on which API is using the object. E.g. it might be required in the context of a PUT xxx API but not for a GET xxx API

shared/api-definitions-ex.asciidoc
jrodewig: Good point. I'm okay with leaving them out for now. I can't think of any way to make these testable without a lot of overhead.

shared/api-definitions-ex.asciidoc
jrodewig: 👍 Not a big deal to me.

shared/api-definitions-ex.asciidoc
jrodewig: Good point. +1 for removing.

PR #943, opened 2019-06-20
More edits to the API template

lcawl:
This PR updates the API reference template as follows:
* adds an attribute for the "Properties" title
* removes float attributes from the template (so the "On this page" navigation works)
* uncomments the Response body since it's the only optional one we were commenting out that way
* changes the template to use "Examples" instead of "Example" by default

### DIFF
--- a/shared/api-ref-ex.asciidoc
+++ b/shared/api-ref-ex.asciidoc
@@ -11,7 +11,7 @@ Use the appropriate heading levels for your book.
 Add anchors for each section.
 FYI: The section titles use attributes in case those terms change.
 ////
-[float]
+
 [[sample-api-request]]
 ==== {api-request-title}
 ////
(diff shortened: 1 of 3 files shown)


shared/api-ref-ex.asciidoc
```diff
@@ -97,15 +97,28 @@ index.
 (string) the name of the index in the leader cluster to follow.
 ////
 
-////
+
 [[sample-api-response-body]]
 ==== {api-response-body-title}
-Response body is only required for detailed responses.
 ////
+Response body is only required for detailed responses.
```
jrodewig: Nice add!

shared/attributes.asciidoc
```diff
@@ -154,4 +154,6 @@
 :api-response-codes-title: Response codes
 :api-response-body-title:  Response body
 :api-example-title:        Example
```
jrodewig: I think we should consider axing the `api-example-title` attribute or just changing its value to `Examples`.

Not a big deal though. That might be better as a separate PR.

lcawl: I think we shouldn't remove it (yet), since I have some pages that are using it. I agree that it likely makes sense to change it to "Examples" eventually but for now I think it's fine as-is, since the templates will no longer mention it.

PR #947, opened 2019-06-24
Fixes 6.x redirects

lcawl:
This PR changes the redirects for 6.x URLs from 6.7 to 6.8.


bmorelli25:
Would I be able to use this same approach in this PR: https://github.com/elastic/docs/pull/958?

PR #955, opened 2019-06-27
Update .NET client docs current to 7.x

russcam:
This commit updates the .NET client documentation to make 7.x the current version.


russcam:
Merging in as .NET client 7.0.0 is now released.

PR #963, opened 2019-07-02
Updates data frame attributes

lcawl:
This PR adds some shared attributes for data frame analytics.

Related to https://github.com/elastic/elasticsearch/pull/43875/files

NOTE: The dataframe-job attributes are commented out in this PR but if they're not used, they should be removed to avoid confusion.

### DIFF
--- a/shared/attributes.asciidoc
+++ b/shared/attributes.asciidoc
@@ -130,7 +130,6 @@
 :dataframes:                 data frames
 :dataframe-cap:              Data frame
 :dataframes-cap:             Data frames
-//:dataframe-job:              {dataframe} analytics job
 //:dataframe-jobs:             {dataframe} analytics jobs
 //:dataframe-jobs-cap:         {dataframe-cap} analytics jobs
 :dataframe-transform:        {dataframe} transform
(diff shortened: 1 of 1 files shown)


shared/attributes.asciidoc
```diff
@@ -130,13 +130,19 @@
 :dataframes:                 data frames
 :dataframe-cap:              Data frame
 :dataframes-cap:             Data frames
-:dataframe-job:              {dataframe} analytics job
-:dataframe-jobs:             {dataframe} analytics jobs
-:dataframe-jobs-cap:         {dataframe-cap} analytics jobs
-:dataframe-analytics-config: {dataframe} analytics config
+//:dataframe-job:              {dataframe} analytics job
```
szabosteve:
```suggestion

```

shared/attributes.asciidoc
```diff
@@ -130,13 +130,19 @@
...
+//:dataframe-jobs:             {dataframe} analytics jobs
```
szabosteve:
```suggestion

```

shared/attributes.asciidoc
```diff
@@ -130,13 +130,19 @@
...
+//:dataframe-jobs-cap:         {dataframe-cap} analytics jobs
```
szabosteve: I agree that we should get rid of these to avoid confusion.

```suggestion

```

PR #976, opened 2019-07-09
Automatically clean up book metadata

### LINKED ISSUE #973: Removing a branch from a book doesn't clear its metadata
We use branch metadata to figure out if a book needs to be rebuilt. When you remove a branch from a book we don't clear the metadata. This leads to weird things like not automatically rebuilding books when the branch is re-added.

nik9000:
We keep information about which branches contribute to which books but
before this change we didn't automatically remove the metadata when
a branch is no longer involved in building a book we never cleared that
metadata. For the most part this just left lines in a YAML file that
aren't used except when we remove a branch from a book and then re-add
it. In that case since we only use the branch data to figure out if we
should rebuild a book we *wouldn't* start to build the book when it is
added back unless the branch that it is built from changes. This happened
to us recently so it makes sense to clean up these files automatically
so it never happens again.

Closes #973

### DIFF
--- a/lib/ES/BranchTracker.pm
+++ b/lib/ES/BranchTracker.pm
@@ -64,7 +64,6 @@ sub prune_out_of_date {
     while ( my ($repo, $branches) = each %{ $self->{shas} } ) {
         my $allowed_for_repo = $allowed{$repo} || '';
         unless ($allowed_for_repo) {
-            say "Pruning for $repo";
             delete $self->{shas}->{$repo};
             next;
         }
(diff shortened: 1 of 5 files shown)


nik9000:
@olksdr could you have a look at this one please?

lib/ES/BranchTracker.pm
```diff
@@ -56,39 +48,71 @@ sub sha_for_branch {
...
+            delete $self->{shas}->{$repo};
+            next;
+        }
+        foreach my $branch ( keys %{ $branches } ) {
+            # We can't clear the link check information at this point safely
+            # because we need it for PR builds and we don't have a good way
+            # tell if it'll be needed again. It is a problem, but not a big one
+            # right now.
+            unless ($allowed_for_repo->{$branch} || $branch =~ /^link-check/) {
+                say "Pruning for $repo $branch";
```
nik9000: This is leftover debugging info and I'll remove before committing.

lib/ES/BranchTracker.pm
```diff
@@ -56,39 +48,71 @@ sub sha_for_branch {
...
+                delete $branches->{$branch};
+            }
+        }
+        # Empty can show up because there is a new book that weren't not
```
olksdr: I think one `not` too many here -> `book that weren't not`

nik9000:

PR #977, opened 2019-07-09
Add new main nav and footer for 2019 redesign

brianjolly:
This PR is to add a dynamic nav and footer to the docs for the upcoming site redesign.

### DIFF
--- a/resources/web/template.html
+++ b/resources/web/template.html
@@ -239,26 +239,6 @@
           }
         });
     
-        // if($(window).width() < 769){
-        //   var $item = $('.scrollable-panel li'), 
-        //       visible = 3, //Set the number of items that will be visible
-        //       index = 0, //Starting index
-        //       endIndex = ( $item.length / visible ) - 1; //End index (NOTE:: Requires visible to be a factor of $item.length... You can improve this by rounding...)
-    
-        //   $('#right-arrow').click(function(){
-        //     if(index < endIndex ){
-        //       index++;
-        //       $item.animate({'left':'-=300px'});
-        //     }
-        //   });
-    
-        //   $('#left-arrow').click(function(){
-        //     if(index > 0){
-        //       index--;            
-        //       $item.animate({'left':'+=300px'});
-        //     }
-        //   });
-        // }
       });
       $(window).resize(function () {
         $('.horizon-swiper').horizonSwiper({
--- a/resources/web/template.html
+++ b/resources/web/template.html
@@ -166,23 +166,6 @@
     
     <div id='elastic-nav'></div>
 
-    <header class="header-wrapper">
-      <div class="header-search-wrapper open-search"> 
-        <div class="container">
-          <div class="big-search">
-            <i class="big-search-icon"></i>
-            <form id="searchfrm" name="searchForm" autocomplete="on" action="/search" method="get">
-              <ul class="tags-wrapper">
-                <li class="search-field"><input type="text" name="q" id="autocomplete" class="form-control global-input" autocomplete="off"></li>
-              </ul>
-            </form>
-            <a class="header-search-cancel" href="#"></a>
-          </div>
-          <div class="nav-auto-complete"></div>
-        </div>
-      </div>
-    </header>
-
     <link rel="stylesheet" type="text/css" href="/static/css/horizon-swiper.min.css">
 
     <style> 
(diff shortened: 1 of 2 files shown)


resources/web/template.html
```diff
@@ -161,1115 +101,314 @@
...
-		</div>
-	</div>
-</div>
-  </div>
-</footer>
+<div id='elastic-footer'</div>
 <!-- Footer Section end-->
 
+<!--
+<script src='https://elastic-co-redesign.app.elstc.co/elastic-nav.js'></script>
```
nik9000: Could you drop these before merging?

resources/web/template.html
```diff
@@ -161,1115 +101,314 @@
...
-							<li class="icon-kibana-white"><a href="https://www.elastic.co/webinars/getting-started-kibana?baymax=default&elektra=docs&storm=top-video">Introduction to Kibana</a></li>
-							<li class="icon-logstash-white"><a href="https://www.elastic.co/webinars/getting-started-logstash?baymax=default&elektra=docs&storm=top-video">Logstash Starter Guide</a></li>
-						</ul>
-						</div>
-					</div>
-				</div>
-			</div>
-		</div>
-	</section>
+          <!-- <div class="pageheader">
```
nik9000: Is there a good reason we should keep this? If it is commented out I think maybe we should trash it too

resources/web/template.html
```diff
@@ -161,1115 +101,314 @@
...
-      </div>
-    </div>
-  
-
-
-<style type="text/css">
-
-  
-</style>
+        // if($(window).width() < 769){
```
nik9000: Should We pitch this before merging? it looks like debuggin leftovers.

resources/web/template.html
```diff
@@ -161,1115 +101,314 @@
...
+    
+        //   $('#left-arrow').click(function(){
+        //     if(index > 0){
+        //       index--;            
+        //       $item.animate({'left':'+=300px'});
+        //     }
+        //   });
+        // }
+      });
+      $(window).resize(function () {
```
nik9000: Should we move this into docs.js?

nik9000: Ah. I see. It was in there before. We *probably* should move it to docs.js but it can totally wait in that case.

resources/web/template.html
```diff
@@ -161,1115 +101,314 @@
...
-    $('body').on('click', function(event){
-      if (!event.target.matches('.dropdown-btn')) {
-
-        var dropdowns = $(".dropdown-content");
-        var i;
-        for (i = 0; i < dropdowns.length; i++) {
-          var openDropdown = dropdowns[i];
-          if (openDropdown.classList.contains('show')) {
-            openDropdown.classList.remove('show');
+        $('body').on('click', function(event){
```
nik9000: These should probably also move to docs.js but they can wait.

resources/web/template.html
```diff
@@ -161,1115 +101,314 @@
...
-  }
-
-  console.log('apm pageLoadName:', pageLoadName, parts)
-  elasticApm.setTags({env:"production"});
-  elasticApm.setInitialPageLoadName(pageLoadName);
-</script>
-
-<!-- Elastic APM -->
-
+    <!--
```
nik9000: Do we mean to have this commented out?

nik9000: Could we just remove it?

brianjolly: I'm noticing an error when I enable APM, but we can launch without and uncomment latter in the week when it is working.

resources/web/template.html
brianjolly: ++ I'll plan on moving this later this week.

resources/web/template.html
brianjolly: ++ I'll plan on moving this later this week.

resources/web/template.html
brianjolly: ++ just deleted in `6d2458a4af0`

resources/web/template.html
brianjolly: ++ deleted in `26ca0956f26`

resources/web/template.html
brianjolly: ++ deleted in `f64dd29741d`

brianjolly:
Merged! Thanks @nik9000!

PR #980, opened 2019-07-11
[DOCS] Adds continuous data frame related shared attributes.

szabosteve:


lcawl:
I think we'll want to wait and chat with @sophiec20 before merging this

PR #982, opened 2019-07-11
Updates to API template

lcawl:
Related to https://github.com/elastic/docs/issues/937

This PR updates the API reference template to change the location of the Required/Optional parameter qualifier (hopefully for the last time!).

I also modified the comment about the definitions list, since I think we should have some leeway for whether we create separate pages for all objects or not.  i.e. For small objects it might annoy users to be sent to a separate page.

### DIFF
--- a/shared/api-ref-ex.asciidoc
+++ b/shared/api-ref-ex.asciidoc
@@ -46,7 +46,7 @@ Guidelines for parameter documentation
 ***************************************
 * Use a definition list.
 * End each definition with a period.
-* Include the data type and whether the parameter is Optional or Required.
+* Include whether the parameter is Optional or Required and the data type.
 * Include default values as the last sentence of the first paragraph.
 * Include a range of valid values, if applicable.
 * If the parameter requires a specific delimiter for multiple values, say so.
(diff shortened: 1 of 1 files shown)


shared/api-ref-ex.asciidoc
```diff
@@ -46,13 +46,12 @@ Guidelines for parameter documentation
...
 * End each definition with a period.
-* Include the data type.
-* Include whether the parameter is Optional or Required.
+* Include the data type and whether the parameter is Optional or Required.
 * Include default values as the last sentence of the first paragraph.
 * Include a range of valid values, if applicable.
 * If the parameter requires a specific delimiter for multiple values, say so.
 * If the parameter supports wildcards, ditto.
-* For objects or nested objects, link to a separate definition list.
+* For large or nested objects, consider linking to a separate definition list.
```
jrodewig: Oohhh. Good catch.

shared/api-ref-ex.asciidoc
```diff
@@ -46,13 +46,12 @@ Guidelines for parameter documentation
 ***************************************
 * Use a definition list.
 * End each definition with a period.
-* Include the data type.
-* Include whether the parameter is Optional or Required.
+* Include the data type and whether the parameter is Optional or Required.
```
jrodewig: Should we reverse the order here since the current template includes optional/required THEN data type.

lcawl: Done, thanks!

PR #984, opened 2019-07-11
Add an option to comment about a preview

nik9000:
Adds an option to the build that causes it to comment that a preview of
the docs changes are available after it pushes a change. We'll use this
on the pull request builds to make it obvious where the previews are.


nik9000:
Could you have a look at this one @olksdr?

nik9000:
Thanks @olksdr!

PR #985, opened 2019-07-12
Always fork from master when targeting a branch

nik9000:
Previously when you specified `--target_branch` we'd fork that branch
from master if it didn't exist and otherwise we'd build "on top" of that
branch. This isn't quite right for pull request testing. The trouble is
that you often want to merge master into your pull request branch. Doing
that introduces dependencies on other repos that touch the same books
you do. But if your target branch is old we'll reuse old commits from
that branch which might not have those dependencies. It can make a
somewhat simple PR totally fail to build if it sits for a week or two.

This change makes it so we always fork the `--target_branch` from the
master branch. We fork it from master even if it already exists. This
both fixes the "out of date" issue above and it better simulates when
merging the PR will be like because, after all, the docs built from the
PR will be built "on top of" the current master.

### DIFF
--- a/lib/ES/TargetRepo.pm
+++ b/lib/ES/TargetRepo.pm
@@ -51,7 +51,7 @@ sub checkout_minimal {
         my $out = run qw(git clone --no-checkout),
             $self->git_dir, $self->{destination};
 
-        # This if stement handles empty repositories in a way that works with
+        # This if statement handles empty repositories in a way that works with
         # different target branches. It always checks out the master
         # branch. If the target branch is `master` then it will return early.
         # If the target branch isn't master it'll delete the existing copy
--- a/lib/ES/TargetRepo.pm
+++ b/lib/ES/TargetRepo.pm
@@ -43,15 +43,19 @@ sub checkout_minimal {
 #===================================
     my ( $self ) = @_;
 
+    # Whether or not we'll need to force push the target branch.
+    $self->{rebuilding_target_branch} = 0;
+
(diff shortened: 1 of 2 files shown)


nik9000:
@olksdr, could you have a look at this one? It fixes a sneaky problem I found while working with the universal PR tests.

nik9000:
@olksdr, could you have a look at this one when you get the chance please? Thanks for all of your reviews!

olksdr:
ah, I'm sorry - it's slipped from me, having a look now.

lib/ES/TargetRepo.pm
```diff
@@ -43,15 +43,19 @@ sub checkout_minimal {
...
     my $original_pwd = Cwd::cwd();
     eval {
         my $out = run qw(git clone --no-checkout),
             $self->git_dir, $self->{destination};
 
-        # This if statement will *either* put the repo on the branch we want
-        # and exit or the branch we want doesn't exist and we need to fork it
-        # from master. In that case it'll put the repo on the master branch
-        # and fall out.
+        # This if stement handles empty repositories in a way that works with
```
olksdr: `s/stement/statement/` 😄

nik9000: Words are hard!

nik9000:
> ah, I'm sorry - it's slipped from me, having a look now.

Thanks for looking! I trying and ping folks if it has been a few days because things totally slip through the cracks.

PR #986, opened 2019-07-12
Preview: More closely mimick nginx on dir request

nik9000:
When you request a directory on the public web site we redirect your
request to `{path}/index.html`. The preview *was* just serving that
index file instead. This sounds good but breaks relative links in the
`index.html` file. Sadly, we have some of those. This change teaches the
preview to do the same redirect as the public site.


nik9000:
@haywoood, could you have a look at this one pretty please!

nik9000:

PR #994, opened 2019-07-16
Edit me link improvements

nik9000:
Makes two changes to the generation of "edit" links for asciidoctor:
1. Prefers the longest matching repository. This helps when repository
is `logtstash` and another is `logstash-docs`.
2. Create a setting to enable overriding `edit_me` links using an
attribute. Historically we've overridden this attribute in cases where
it is no longer needed for Asciidoctor and overriden it *incorrectly*.
But there are some cases where we really do want ot override it.
Specifically for logstash. So this allows us to turn it on when we want
it.

### DIFF
--- a/resources/asciidoctor/lib/edit_me/extension.rb
+++ b/resources/asciidoctor/lib/edit_me/extension.rb
@@ -66,6 +66,7 @@ class EditMe < TreeProcessorScaffold
     def edit_url
       if @document.attributes['respect_edit_url_overrides']
         url = @document.attributes['edit_url']
+        return false if url == ''
         return url if url
       end
 
--- a/resources/asciidoctor/lib/edit_me/extension.rb
+++ b/resources/asciidoctor/lib/edit_me/extension.rb
@@ -57,20 +57,26 @@ class EditMe < TreeProcessorScaffold
   # Extension to blocks that need an "edit me" link.
   module WithEditLink
     def title
-      url = edit_url
-      return super unless url
-
-      "#{super}<ulink role=\"edit_me\" url=\"#{edit_url}\">Edit me</ulink>"
+      if (url = edit_url)
+        "#{super}<ulink role=\"edit_me\" url=\"#{url}\">Edit me</ulink>"
+      else
+        super
+      end
     end
 
     def edit_url
       if @document.attributes['respect_edit_url_overrides']
         url = @document.attributes['edit_url']
-        return false if url == ''
-        return url if url
+        if url == ''
+          false
+        elsif url
+          url
+        else
+          edit_url_by_path
+        end
+      else
+        edit_url_by_path
       end
-
-      edit_url_by_path
     end
 
     def edit_url_by_path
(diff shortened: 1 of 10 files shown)


nik9000:

@karenzone, I believe I need this one before converting logstash to asciidoctor.

resources/asciidoctor/lib/edit_me/extension.rb
```diff
@@ -28,50 +28,73 @@ def process(document)
       url = url[0..-2] if url.end_with? '/'
       edit_urls << { toplevel: toplevel, url: url }
     end
-    edit_urls.reverse!
+    # Prefer the longest matching edit url
+    edit_urls = edit_urls.sort_by { |e| [-e[:toplevel].length, e[:toplevel]] }
```
estolfo: you could also use [max](https://ruby-doc.org/core-2.6.3/Enumerable.html#method-i-max), using the length as the comparable attribute in the supplied block.

resources/asciidoctor/lib/edit_me/extension.rb
```diff
@@ -28,50 +28,73 @@ def process(document)
...
+        source_location.file
+      end
+    end
+  end
+
+  ##
+  # Extension to blocks that need an "edit me" link.
+  module WithEditLink
+    def title
+      url = edit_url
```
estolfo: Is there a reason you set the `url` variable here?
You can do `return super unless edit_url`
Then use the `edit_url` in the string as you've done in line 63.

resources/asciidoctor/lib/edit_me/extension.rb
```diff
@@ -28,50 +28,73 @@ def process(document)
...
+    def edit_url
+      if @document.attributes['respect_edit_url_overrides']
+        url = @document.attributes['edit_url']
+        return false if url == ''
+        return url if url
+      end
+
+      edit_url_by_path
+    end
+
```
estolfo: The `if` block will return the last evaluated statement so you could also write this:
```ruby
def edit_url
  if @document.attributes['respect_edit_url_overrides']
    url = @document.attributes['edit_url']
    return false if url == ''
    url
  end || edit_url_by_path
end
```

resources/asciidoctor/lib/edit_me/extension.rb
```diff
@@ -28,50 +28,73 @@ def process(document)
...
-        # the base_dir which we use when we build books from many repos.
-        document.source_location.file
-      end
-    else
-      def block.source_path
-        # source_location.path doesn't work for relative includes outside of
-        # the base_dir which we use when we build books from many repos.
-        source_location.file
-      end
+      url + path[entry[:toplevel].length..-1]
```
estolfo: I wouldn't use so many `return`s here and instead use `if-else` like this:

```ruby
if entry
  url = entry[:url]
  if url == '<disable>'
    false
  else
    url + path[entry[:toplevel].length..-1]
  end
else
  logger.warn#...etc
  false
end
```

nik9000:
> Hi Nik, I made some suggestions for the code to be more idiomatic Ruby.

Thanks for reviewing! I'll do it!

resources/asciidoctor/lib/edit_me/extension.rb
nik9000: I really want to sort the whole list so the longest are first so we check them first.

estolfo: ah, ok. I didn't see the big picture - you want a sorted list, not just the longest. Nevermind!

nik9000:
@estolfo, I've pushed an update with the changes you requested. Please have another look when you get the chance!

estolfo:
looks good!

nik9000:
Thanks for reviewing @estolfo !

PR #996, opened 2019-07-17
Enable ECK doc build for 0.9

pebrc:
Adds the new 0.9 branch to the docs build.


pebrc:
We are running a partial docs build as part of each PR build so we should be OK 🤞 

https://github.com/elastic/cloud-on-k8s/blob/2b96e0d1a2828b20f00d800d7966061eb7984d0e/build/ci/pr/Jenkinsfile#L39-L44

PR #1001, opened 2019-07-18
Introduce build step for docs.js, enable code splitting for easier testing

haywoood:
This is not a rewrite of the existing code. The goal of this effort is to break apart the original docs.js file so what we're left with is just the code we care about. By introducing a simple web application bundler, we can remove the various bits of code that existed to protect against the global scope—a necessary and wise precaution that is now taken care of by said bundler.

Without the need to close over our code and isolate it from the global scope, we can export top-level functions into tests, greatly increasing code coverage, as well as enhancing the readability of our intent.

Some refactors that were done:
- Remove the global `Strings` variable into a function that returns the correct phrase. The benefit of this over a map is that we can add some error handling to guard against typos when developing and give us some flexibility in the future if we need to expand it. Added tests for each phrase and language.
- Move the `base_url` regex to a function with tests


lib/ES/Template.pm
```diff
@@ -73,6 +76,19 @@ my $Autosense_RE = qr{
...
     }xs;
 
+#===================================
+sub _build_js {
+#===================================
+    my ( $self ) = (@_);
+
+    if ($self->{compiled_JS} == 0) {
+        print "Building docs.js\n";
+        `node_modules/parcel/bin/cli.js build resources/web/docs_js/index.js --no-minify -d resources/web -o docs.js`;
```
nik9000: I'd prefer you use the `run` method from `Util.pm` here so it'll log its output in the same way as all of the other subprocesses.

lib/ES/Template.pm
```diff
@@ -13,6 +13,7 @@ sub new {
     my ( $class, %args ) = @_;
 
     my $self = bless {
+        compiled_JS  => 0,
```
nik9000: Would `compiled_js` be ok here?

haywoood: Yea

lib/ES/Template.pm
haywoood: Ah right, I have that written down from our convo, thanks for the reminder.

lib/ES/Template.pm
```diff
@@ -73,6 +77,19 @@ my $Autosense_RE = qr{
...
     }xs;
 
+#===================================
+sub _build_js {
+#===================================
+    my ( $self ) = (@_);
+
+    if ($self->{compiled_js} == 0) {
+        print "Building docs.js\n";
+        run '/node_modules/parcel/bin/cli.js', 'build', '/build_docs/resources/web/docs_js/index.js', '-d', '/build_docs/resources/web', '-o', 'docs.js';
```
nik9000: Could we move this into `/tmp/` instead of writing it to the project directory? Most of the rest of the docs build uses `/tmp` for scratch work which is super safe because the docker container will vanish at the end of the run.

haywoood: Ah, well, everything expects `resources/web/docs.js` to exist, and this creates it.

lib/ES/Template.pm
```diff
@@ -73,6 +77,19 @@ my $Autosense_RE = qr{
         :(?:CONSOLE|AUTOSENSE|KIBANA):
     }xs;
 
+#===================================
+sub _build_js {
+#===================================
+    my ( $self ) = (@_);
+
+    if ($self->{compiled_js} == 0) {
```
nik9000: I think this doesn't work with `--all` because it [forks](https://github.com/elastic/docs/blob/master/lib/ES/Book.pm#L264). I ran this to build a preview:
```
./build_docs --all --keep_hash --target_repo git@github.com:elastic/built-docs.git --target_branch haywood_1001     --push --rebuild
```

And that emits `Building docs.js` super frequently. I think you should make `build_js` public and call it around [here](https://github.com/elastic/docs/blob/master/build_docs.pl#L105) and [here](https://github.com/elastic/docs/blob/master/build_docs.pl#L242). I hate leaking the abstraction when we don't strictly have to but I think that it is important not to build the file a bunch of times, especially in different processes in parallel. I think we're bound to half built files if we do that.

resources/web/styles.css
```diff
@@ -656,15 +656,14 @@ p.remark {
 }
 
 #guide a.console_settings:hover,
+#guide a.kibana_settings:hover,
```
nik9000: HA!

nik9000:
I've checked the scripts for copy as curl and open in console and they look great to me.

PR #1004, opened 2019-07-19
Preview: Add a diff endpoint

nik9000:
I've found that it can be kind of annoying trying to find out which
files changed when looking up a docs preview. This adds a page at
`/diff` that contains links to all of the modified pages.

I used chunked transfer and stream parsing the result from git because
that felt like the right thing to do here but it did make the parsing a
bit more complex.

I also chose to use the `-z` flag when calling git because it emits a
format that isn't jumbled when files move. It has many ascii-nul
terminated records which means I couldn't really use any off the shelf
nodejs parsing libraries.

### DIFF
--- a/preview/preview.js
+++ b/preview/preview.js
@@ -1,3 +1,5 @@
+'use strict';
+
 /*
  * Little server that listens for requests in the form of
  * `${branch}.host/guide/${doc}`, looks up the doc from git and streams
@@ -9,7 +11,9 @@
  */
 
 const child_process = require('child_process');
+const dedent = require('dedent');
 const http = require('http');
+const stream = require('stream');
 const url = require('url');
 
 const port = 3000;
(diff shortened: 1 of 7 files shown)


nik9000:
The page is totally unstyled but does the job:
2019-07-19-094822_1463x640_scrot

nik9000:
If you want to make your laptop lift off do
```
preview/test.sh
<wait for it to come online>
curl staging.localhost:8000/diff
```

That makes `git` take a few seconds to get the results to you:
```
22054 root      20   0 1137816 346088 119020 R  99.7   2.1   0:34.74 git diff-tree -z --find-renames --numstat staging --                                    
```

nik9000:
I'll also want to bump the docker image version in this PR. I'll push an update in a few minutes.

nik9000:
> I'll also want to bump the docker image version in this PR. I'll push an update in a few minutes.

Done.

@haywood, could you have a look at this when you get a chance? I think it'll make the previews easier to use.

preview/preview.js
```diff
@@ -54,51 +60,197 @@ const requestHandler = (request, response) => {
...
+
+  const pipeline = child.stdout.pipe(handle);
+  rig_handlers(child, response, pipeline, response => {
+    response.write(handleChunk());
+    if (chunk !== '') {
+      console.error('unprocessed results from git', chunk);
+      response.write(`  <li>Unprocessed results from git: <pre>${chunk}</pre>`);
+      response.status = 500;
+    } else if (!completeSuccess) {
+      response.write(`  <li>Error processing some entries from git. See logs.`);
```
haywoood: I might just be missing where these `li` tags are being closed.

nik9000:
Thanks for the review @haywoood!

PR #1007, opened 2019-07-19
Docs preview: fix two silly mistakes

nik9000:
Fixes two silly mistakes with the docs preview:
1. Removes a debugging output that I left in the path that handles
renames.
2. Handles empty diffs. Those should be rare, but we should produce
valid html when we have them.

### DIFF
--- a/preview/preview.js
+++ b/preview/preview.js
@@ -29,7 +29,7 @@ const requestHandler = (request, response) => {
   const parsedUrl = url.parse(request.url);
   const branch = gitBranch(request.headers['host']);
   if (parsedUrl.pathname === '/diff') {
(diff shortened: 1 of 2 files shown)


preview/preview.js
```diff
@@ -178,7 +177,7 @@ function serve_diff(branch, response) {
   });
 
   const pipeline = child.stdout.pipe(handle);
-  rig_handlers(child, response, pipeline, response => {
+  rigHandlers(child, response, pipeline, response => {
```
haywoood: yea camelCase is more idiomatic in JS as a side note.

nik9000:

PR #1012, opened 2019-07-22
Docs preview: Skip repos we don't own

nik9000:
Our pattern for pull request branches is a little broad so sometimes
normal looking branches will match it and we'll *think* they are for a
pull request when they really aren't. This change makes it so that when
we can't find the pull request information for a branch we log a warning
and keep it open rather than crash. That seems sensible.


nik9000:
Thanks @haywoood!

PR #1013, opened 2019-07-23
Docs preview: Add a "go away" robots.txt

nik9000:
The docs-preview service is never going to be the canonical place to
host docs and we're not going to set up canonical tags on it or
anything. This kindly asks search engines not to index it.


nik9000:

PR #1014, opened 2019-07-23
Drop guide_root from redirects

nik9000:
We no longer need it and it was difficult to configure.


brianjolly:
This looks good to me 👍

nik9000:

PR #1017, opened 2019-07-25
Clean up last of Makefile based integ tests

nik9000:
A few months ago all of our integration tests were in
`integtest/Makefile` and it was hard to read so we slowly moved to rspec
for those tests. Now that we're finished moving we no longer need a
large list of files. This drops those files and updates the `README` in
`integtest` to remove mention of the `Makefile` based tests.


nik9000:

PR #1024, opened 2019-07-30
Find "alternative" examples from other repos

nik9000:
This adds support for looking up "alternative examples" for examples.
Those alternative examples are included into the html of each doc but
hidden. We will build *something* to make them visible later.

You can do this when building a single book with the `--alternatives`
flag like so:

```
~/Code/Elastic/docs/build_docs \
    --doc ~/Code/Elastic/elasticsearch/docs/reference/index.asciidoc \
    --resource ~/Code/Elastic/elasticsearch/x-pack/docs/ \
    --chunk 1 --asciidoctor \
    --out ~/Code/Elastic/built_docs/ \
    --alternatives console:php:~/Code/Elastic/elasticsearch-php/examples
```

Or you can configure a book to always be build with alternatives like so:

```
          -
            title:      Elasticsearch Reference
...
            sources:
              -
                repo:   elasticsearch
                path:   docs/reference
...
              -
                alternatives: { source_lang: console, alternative_lang: csharp }
                repo:   elasticsearch-net
                path:   examples
                exclude_branches:   [ 7.x, 7.3, 7.2, 7.1, 7.0, 6.8, 6.7, 6.6, 6.5, 6.4, 6.3, 6.2, 6.1, 6.0, 5.6, 5.5, 5.4, 5.3, 5.2, 5.1, 5.0, 2.4, 2.3, 2.2, 2.1, 2.0, 1.7, 1.6, 1.5, 1.4, 1.3, 0.90 ]
```

You can manually switch to a particular language by adding this style:

```
#guide default.has-csharp {
    display: none;
}

#guide .alternative.lang-csharp {
    display: block;
}
```


nik9000:
When I artificially add a fairly large example for *every* snippet in 9 languages and build the Elasticsearch reference it takes about 220 seconds to build. When I build it with *just* the 8 examples that I've made for php it takes about 55 seconds. Building without any examples takes about the same amount of time. So that isn't great. But this is "ok", especially because you can opt out of using the examples locally pretty easily.

nik9000:
I think this is ready for some folks to look at the code. I believe the most important reviewer will be @estolfo as the non-ruby changes are pretty straitforward. Everyone else is on here to have a look at the output and make sure it makes sense.

resources/asciidoctor/lib/alternative_language_lookup/alternative.rb
```diff
@@ -0,0 +1,144 @@
...
+        doctype: Asciidoctor::DEFAULT_DOCTYPE,
+        sourcemap: @listing.document.sourcemap,
+        base_dir: @listing.document.base_dir,
+        to_dir: @listing.document.options[:to_dir]
+      )
+      @child.parse
+    end
+
+    def validate
+      unless [1, 2].include?(@child.blocks.length)
```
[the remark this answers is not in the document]
nik9000: It totally could be. I'll switch it to either `==`.

conf.yaml
```diff
@@ -235,6 +235,27 @@ contents:
                 repo:   elasticsearch
                 path:   x-pack/plugin/sql/qa
                 exclude_branches:   [ 6.5, 6.4, 6.3, 6.2, 6.1, 6.0, 5.6, 5.5, 5.4, 5.3, 5.2, 5.1, 5.0, 2.4, 2.3, 2.2, 2.1, 2.0, 1.7, 1.6, 1.5, 1.4, 1.3, 0.90 ]
+              -
+                alternatives: { source_lang: console, alternative_lang: php }
+                repo:   elasticsearch-php
+                path:   examples
+                exclude_branches:   [ 7.x, 7.3, 7.2, 7.1, 7.0, 6.8, 6.7, 6.6, 6.5, 6.4, 6.3, 6.2, 6.1, 6.0, 5.6, 5.5, 5.4, 5.3, 5.2, 5.1, 5.0, 2.4, 2.3, 2.2, 2.1, 2.0, 1.7, 1.6, 1.5, 1.4, 1.3, 0.90 ]
+                map_branches:
+                  master: doc-examples
```
nik9000: @philkra, would you be ok merging the `doc-examples` branch to `master` in `elasticsearch-php`? I think it'd be a little cleaner to pick them up from there.

conf.yaml
```diff
@@ -235,6 +235,27 @@ contents:
...
+              -
+                alternatives: { source_lang: console, alternative_lang: csharp }
+                repo:   elasticsearch-net
+                path:   examples
+                exclude_branches:   [ 7.x, 7.3, 7.2, 7.1, 7.0, 6.8, 6.7, 6.6, 6.5, 6.4, 6.3, 6.2, 6.1, 6.0, 5.6, 5.5, 5.4, 5.3, 5.2, 5.1, 5.0, 2.4, 2.3, 2.2, 2.1, 2.0, 1.7, 1.6, 1.5, 1.4, 1.3, 0.90 ]
+                map_branches:
+                  master: doc-examples
```
nik9000: @russcam , would you be ok merging the `doc-examples` branch to `master` in `elasticsearch-net`? I think it'd be a little cleaner to pick them up from there.

nik9000:
I discussed finding the examples in arbitrary subdirectories with @russcam and we decided a good idea. I've decided to investigate it in a follow up because this works pretty well and has gone through code review.

I've reconfigured this to pull in only php because php is ready. We can pull in the other clients when they make their examples directory.

PR #1028, opened 2019-08-01
Style: Fix toc line spacing

### LINKED ISSUE #1006: Line spacing in TOC makes it hard to tell when a title has wrapped
Typically I avoid using long titles, but in this case I want to pop some troubleshooting topics up a level so they appear in the navigation.

To make the titles easier to read, there should be less space between lines when a title is wrapped. Right now, it looks like this:

image

It should look more like this:

image

nik9000:
Our table of contents and "on this page" section was using
`line-height: 26px` which looked fairly bad if the title broke to the
next line. This switches it to using `line-height: 1em` and uses
margin or padding to maintain the separation between items instead.

Closes #1006

### DIFF
--- a/resources/web/styles.css
+++ b/resources/web/styles.css
@@ -132,15 +132,9 @@ h1, h2, h3, h4, h5, h6 {
     font-size: 0.85em;
 }
 
-#this_page li > a > span {
-	color: #00a9e5;
-  display: inherit;
-  font-size: inherit;
-  margin-top: inherit;
-  margin-bottom: inherit;
-  margin-left: inherit;
-  margin-right: inherit;
-  font-weight: inherit;
+#this_page li {
+    line-height: 1em;
+    margin-bottom: .5em;
 }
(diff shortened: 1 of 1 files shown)


resources/web/styles.css
```diff
@@ -132,15 +132,9 @@ h1, h2, h3, h4, h5, h6 {
     font-size: 0.85em;
 }
 
-#this_page li > a > span {
```
nik9000: This didn't match anything any more.

nik9000:
This changes this:
2019-08-01-130218_699x487_scrot
into this:
2019-08-01-130225_674x366_scrot

and this:
2019-08-01-131146_755x854_scrot
into this:
2019-08-01-131158_783x804_scrot

brianjolly:
Hi @nik9000, this looks good to me! :)

nik9000:
Thanks @brianjolly!

PR #1031, opened 2019-08-01
Add cURL host configuration to the 'settings' modal

haywoood:
Since adding a build step with Parcel.js, we now get JSX for free, so I've added Preact and Redux for better component support going forward. The original size of the file was 22kb, and after adding Preact, Redux, react-redux, and Some Ramda we've roughly doubled the size to 43kb.

The console area and the modal have been refactored into React components that connect to the store for the data they need. The store is initialized here https://github.com/elastic/docs/compare/haywood/console_copy?expand=1#diff-911555b7aa64dc79d8b9aeb2ceff62d8R152. Centralizing this data into a container that UI elements can subscribe and react to simplifies development across the board.

We still use jQuery to iterate over the elements and grab embedded data from the source. There's `components/mount.js` that you can use to attach React components to jQuery elements with the Redux store available to child elements.

`console_widget.jsx` is used for Console, Sense, and Kibana widgets. Console and Sense are mostly the same, but we use an `isKibana` boolean value to switch the curl host and hide the view in console link.

cURL host, username, and password fields were added to the form modal, which will be used in `copy as cURL` if present. That data is also stored in the cookie to persist through page transitions.


lib/ES/Util.pm
```diff
@@ -706,7 +706,7 @@ sub timestamp {
 #===================================
 sub build_docs_js {
 #===================================
-    run '/node_modules/parcel/bin/cli.js', 'build', 'resources/web/docs_js/index.js', '/node_modules', '-d', 'resources/web', '-o', 'docs.js';
+    run '/node_modules/parcel/bin/cli.js', 'build', 'resources/web/docs_js/index.js', '/node_modules', '-d', 'resources/web', '-o', 'docs.js', '--experimental-scope-hoisting';
```
haywoood: This new flag enables tree shaking so we can cherry-pick utility functions from Ramda without adding more thank a kb or so to the final package.

package.json
```diff
@@ -11,11 +11,19 @@
...
-   },
+    "setupFiles": [
+      "./jest.js"
+    ]
+  },
   "dependencies": {
     "dedent": "^0.7.0",
     "js-cookie": "^2.1.0",
-    "ramda": "^0.26.1"
+    "linkstate": "^1.1.1",
```
haywoood: Preact helper package that makes writing form components a bit easier. It's small, they recently broke it out of the core library.

lib/ES/Util.pm
nik9000: >  Tree shaking is a term commonly used in the JavaScript context for dead-code elimination. 

Thanks [webpack.js!](https://webpack.js.org/guides/tree-shaking/). Now I know a new term!

resources/web/docs_js/actions/modal.js
```diff
@@ -0,0 +1,16 @@
...
+const initialState = {isOpen: false};
+
+export const reducer = (state = initialState, action) => {
+  switch (action.type) {
+    case SET_MODAL:
+      return {isOpen: action.value};
+    default:
+      return state;
+  }
+};
```
JoshMock: Optional, but I've been finding that [`redux-actions`](https://redux-actions.js.org/) has been helping cut down on Redux action/reducer boilerplate quite a bit.

It's also helped me, in the Docker UI work, to start organizing my store by application subdomain rather than a file hierarchy based on Redux concepts. [Here's how my store is organized.](https://github.com/elastic/infra/tree/master/docker/container-library/ui/src/store/)

Not asking you to make the change, just a suggestion if/when this starts getting complex.

resources/web/docs_js/actions/settings.js
```diff
@@ -0,0 +1,39 @@
+import Cookies from "../../../../../node_modules/js-cookie";
```
JoshMock: Why are you importing from `node_modules` explicitly instead of doing `import Cookies from 'js-cookie';` and letting Parcel's resolver find it for you?

resources/web/docs_js/actions/settings.js
```diff
@@ -0,0 +1,39 @@
...
+    language: lang,
+    langStrings: LangStrings,
+    baseUrl: base_url,
+    kibana_url,
+    console_url,
+    sense_url,
+    curl_host: Cookies.get("curl_host") || "localhost:9200",
+    curl_user: Cookies.get("curl_user"),
+    curl_password: Cookies.get("curl_password")
+  */
```
JoshMock: Deletable?

resources/web/docs_js/components/console_widget.jsx
```diff
@@ -0,0 +1,96 @@
+import * as utils from "../utils.js";
```
JoshMock: Just being picky about consistency, but sometimes you import with the `.js` extension on and sometimes you don't.

resources/web/docs_js/components/console_widget.jsx
```diff
@@ -0,0 +1,96 @@
...
+      <input id="url" type="text" value={state[props.setting]} onInput={linkState(this, props.setting)} />
+
+      <label for="curl_host">cURL Host</label>
+      <input id="curl_host" type="text" value={state.curl_host} onInput={linkState(this, "curl_host")} />
+
+      <label for="curl_username">cURL Username</label>
+      <input id="curl_username" type="text" value={state.curl_user} onInput={linkState(this, "curl_user")} />
+
+      <label for="curl_pw">cURL Password</label>
+      <input id="curl_pw" type="text" value={state.curl_password} onInput={linkState(this, "curl_password")} />
```
JoshMock: That `linkState` looks pretty handy. 👌

resources/web/docs_js/components/modal.js
```diff
@@ -0,0 +1,18 @@
+import {h} from "../../../../../node_modules/preact";
+import {connect} from "../../../../../node_modules/preact-redux";
+import {closeModal} from "../actions/modal";
+
+export const Modal = props => {
+  return props.isOpen && (
+    <div id="settings_modal_bg">
+      <div id="settings_modal">
+        <div class="settings_modal-close" onClick={props.closeModal}>&times;</div>
```
JoshMock: `class` vs `className`

resources/web/docs_js/actions/settings.js
haywoood: Yea, kind of stupid impl detail we're working with at the moment. We install js dependencies in the 'host' docker image, so node_modules are up a directory from where they are when you install it locally. Parcel doesn't have an easy way to say "here's the path to node_modules" (ironically CLJS does :P), and I couldn't get it working with native package.json solutions because Parcel implements a custom resolver. So, for now, we resort to relative paths that target the folder.

JoshMock: FWIW [Webpack also has a resolver API that's pretty well documented](https://webpack.js.org/concepts/module-resolution/) that solves for exactly that. Not sure if there's a counterpart in Parcel that's easy to configure.

resources/web/docs_js/components/modal.js
haywoood: totally unintentionally forgot to change this, but I just learned that preact supports `class`. I will update it to className just to be consistent with regular React just in case.

resources/web/docs_js/components/console_widget.jsx
haywoood: Totally, need to drop js

resources/web/docs_js/actions/settings.js
haywoood: I added the comment for documentation purposes, I can remove, but even better is adding a default data structure to the file maybe.

resources/web/docs_js/actions/modal.js
haywoood: Yea I thought about it and I do like redux-actions and how you used it.

> organizing my store by application subdomain rather than a file hierarchy based on Redux concepts

I think we basically have the same thing. Our actions files export a reducer, and the names of the files correspond to the key used in the store. I'm missing some sugar from redux-actions inside my files though.

resources/web/docs_js/actions/settings.js
haywoood: Unfortunately not and I thought about switching the entire thing to Webpack more than once, or rollup. Ultimately decided to deal with the relative path for now.

nik9000:
I wonder about that test failure.... All of those passed for me locally.

nik9000:
>  ` [Merge master fix conflicts](/elastic/docs/pull/1031/commits/94eb800b275f850bad0ec4fed3930468c275890f) `

Learning from last time, I see!

PR #1032, opened 2019-08-02
Add Azure Marketplace 7.3 to docs

russcam:
This PR adds the 7.3 version of Azure Marketplace and ARM template to the docs, and marks as the current version.

It must be committed only after 7.3 is made available on the Azure Marketplace. I'll comment when this is the case.


russcam:
Merging in as 7.3.0 is on the Azure Marketplace

PR #1037, opened 2019-08-05
Lookup alternative examples in all subdirs

nik9000:
Changes the build so that, when looking for alternative examples, it
looks in all subdirectories. Instead of performing file system
operations for all lookups it indexes all of the examples up front. This
is significantly faster when there are many subdirectories.

While working on this I discovered that we can almost certainly cache
the results of converting the documents for the alternatives. That will
*probably* save a good bit of time. But it isn't 100% simple. So I left
a TODO to investigate it.


nik9000:
@estolfo, could you have a look at this one? It is all ruby and a follow up to #1024.

nik9000:
> Self-test failed

Ah rubocop, my old friend.

resources/asciidoctor/lib/alternative_language_lookup/alternative.rb
```diff
@@ -10,43 +10,42 @@ class Alternative
...
+      @counter = @document.attr 'alternative_language_counter', 0
+      @text = nil
       load
       return unless validate
 
       munge
-      @loaded = true
-      listing.document.attributes['alternative_language_counter'] = @counter + 1
+      @document.attributes['alternative_language_counter'] = @counter + 1
+      @text = @child.convert
```
estolfo: I don't know how expensive this operation is. If it is indeed expensive, maybe it's worth lazily calling `convert` in the `block` method, as it was before.

nik9000: The `load` and `convert` calls are comparatively slow, yeah. The way we use it this is fine though. I'll leave a note.

nik9000:

PR #1038, opened 2019-08-05
Officially drop support outside of build_docs

nik9000:
This drops all support for running `build_docs.pl` outside of the docker
container managed by `build_docs`. We even get to drop one of the perl
deps!


nik9000:
@haywoood, would you like to review this? I love removing code!

nik9000:

PR #1041, opened 2019-08-06
Remove python finding executable

nik9000:
Removes the python finding executable that we had used in the past to
look up an appropriate version of python. We no longer need it now that
we only support running `build_docs.pl` inside of our standard docker
container which we can ensure has an appropriate python.


nik9000:

PR #1048, opened 2019-08-07
Drop animated scrolling

### LINKED ISSUE #950: Clicking in-page link has slow scrolling
In Safari on Mac (desktop), clicking on a link that leads to the same page (different anchor), causes an animated scroll. This animated scroll is a lot slower than what Safari would do, and the time it takes is longer than -my- threshold for being annoyed ;-)

Would be great if scroll-by-link-click would be as fast as on other pages in Safari.

Possibly source of slowness: https://github.com/elastic/docs/blob/c6cebd0d905cb02fabe7da416a09b4fcfa475c34/resources/docbook-xsl-1.78.1/webhelp/docs/common/main.js#L85

Ref: https://twitter.com/xeraa/status/1142928281626468354

nik9000:
We had animated the scroll when you click on an ID or load a page with
an id to make sure we didn't cover up a "floating" pane at the top of
the page. We don't have such a pane anymore so we can fall back to the
browser's default behavior.

Closes #950

### DIFF
--- a/resources/web/template.html
+++ b/resources/web/template.html
@@ -513,42 +513,6 @@ $(document).ready(function(){
 <script type="text/javascript" src="/static/js/slick.min.js"></script>
 <script src="/static/js/nav-v5.js"></script>
 <script src="/static/js/script.js"></script>
- 
-     
-<script type="text/javascript">
-$(document).ready(function(){
-	if (window.location.hash.length) {
-	      var target = $(window.location.hash);
-	      target = target.length ? target : $(window.location.hash);
-	      if (target.length) {
-	        setTimeout(function(){ $(window).scrollTop(target.offset().top - 60); }, 1200);        
-	      }
-	}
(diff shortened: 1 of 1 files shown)


nik9000:
@elasticmachine, run elasticsearch-ci/doc rebuild

nik9000:
This should have a preview at around [here](http://docs_1048.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/search-aggregations-metrics-cardinality-aggregation.html#_script_3) in two hours or so.

nik9000:
@elasticmachine, run elasticsearch-ci/doc rebuild

nik9000:

nik9000:
Adding @haywoood as an alternative reviewer. I talked with the docs team about this an they are all happy to see it go.

nik9000:
Thanks for reviewing, @haywoood !

PR #1053, opened 2019-08-08
Add the node name to the commit message

nik9000:
This will be useful for tracking down "rogue" builds outside of our
primary build.


nik9000:

PR #1054, opened 2019-08-08
Improve error message if port 8000 taken

### LINKED ISSUE #739: Improve error message if trying to run with `--open` twice
Right now if you run with `./build_docs --open` in one window and then do it again in another window you'll get:
```
Error response from daemon: driver failed programming external connectivity on endpoint hardcore_yonath (254029c33ac546173e4f582bbefe1e995911d107ad3e9e968aad2f7efd7100d7): Bind for 0.0.0.0:8000 failed: port is already allocated.
```
We can improve this error message. Let's look for it and mention that this is probably caused by running with `--open` twice. Or because something else is on that port.

nik9000:
Catches the docker error message if port 8000 is already in use and adds
a more user friendly one below it.

Closes #739

### DIFF
--- a/build_docs
+++ b/build_docs
@@ -259,6 +259,9 @@ def run_build_docs(args):
                 environ['BROWSER'] = 'open'
             webbrowser.open('http://localhost:8000/guide',
                             new=1, autoraise=False)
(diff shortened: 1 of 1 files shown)


nik9000:

PR #1055, opened 2019-08-09
Further modernize web resources

nik9000:
This moves from one css and js file per book to one css and js file
globally and drops support for "abs mode" templates which were required
to make demos before we had a fairly easy way to make previews. Dropping
"abs mode" was required to support the global css and js. The global css
and js is nice because it'll allow us to deploy css and js changes
without rebuilding "all" books. This means we'll be able to deploy those
changes much more quickly.

This change also pushes the css file through parcel which minifies it.
We'll follow this change up with one that better organizes the CSS and
parcel will transparently merge them.

This change also enables parcel's server when building opening a single
book locally. This is nice because you can make changes to the css and
js files and they'll be automatically synced into place. And parcel will
cause them to be reloaded on the fly. This works *perfectly* with css
and seems to cause some flickering with js, but otherwise seems really
useful. I did not do this when building all books locally because I
wanted a way to test "production" settings locally and most of the quick
iteration work I've done has been on a single book.

Finally, this has the lovely side effect of speeding up building a
single book when you don't pass `--open` by entirely skipping the
parcel step. This is a thing some folks do locally to check on the
validity of changes locally before opening a PR. And the tests do it a
whole lot.


.gitignore
```diff
@@ -21,8 +21,6 @@ vendor
 
 # Compiled docs.js
 .cache/
-resources/web/docs.js
```
nik9000: Instead of writing this here I write it directly to the destination when building all docs or to `/tmp/parcel` when building a single book.

build_docs
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+# coding=utf-8
```
nik9000: So I can check for `✨`

lib/ES/Util.pm
```diff
@@ -725,9 +739,31 @@ sub timestamp {
...
+             resources/web/docs_js/index.js resources/web/styles.pcss) );
+}
+
+#===================================
+sub build_web_resources {
 #===================================
-    run '/node_modules/parcel/bin/cli.js', 'build', 'resources/web/docs_js/index.js', '/node_modules', '-d', 'resources/web', '-o', 'docs.js', '--experimental-scope-hoisting';
+    my ( $dest ) = @_;
+
+    run '/node_modules/parcel/bin/cli.js', 'build',
```
nik9000: Can't use `qw` here because of `$dest`

resources/web/docs_js/index.js
```diff
@@ -194,4 +194,11 @@ $(function() {
   } else {
     init_toc(LangStrings);
   }
+
+  // Setup hot module replacement for css if we're in dev mode.
```
nik9000: This is fairly hacky but I don't want to change the template when building a single book just to include this. I could be convinced though.

.gitignore
```diff
@@ -21,8 +21,6 @@ vendor
...
-resources/web/docs.js.map
```
haywoood: 👋

build_docs
```diff
@@ -79,6 +80,42 @@ def build_docker_image():
...
+            if platform == "darwin" and 'BROWSER' not in environ:
+                # On mac webbrowser seeem to want to default to safari which
+                # is weird. If we tell it that the browser's name is `open`
+                # we'll get whatever the user set as their default browser.
+                environ['BROWSER'] = 'open'
+            webbrowser.open('http://localhost:8000/guide',
+                            new=1, autoraise=False)
+            self.opened = True
+
+
```
haywoood: nit: newline

build_docs
```diff
@@ -79,6 +80,42 @@ def build_docker_image():
...
+        """
+        self.need_parcel = True
+
+    def on_line(self, line):
+        if self.opened:
+            return
+        if 'start worker processes' in line:
+            # Wait for nginx or nothing will be listening
+            self.need_nginx = False
+        if '✨  Built in ' in line:
```
haywoood: lol

integtest/spec/single_book_spec.rb
```diff
@@ -440,12 +440,39 @@ def self.index
...
+    end
+
     it 'serves the book' do
       expect(index).to serve(doc_body(include(<<~HTML.strip)))
         <a href="chapter.html">Chapter
       HTML
     end
+    context 'the js' do
+      it 'is unminified' do
+        expect(js).to serve(include(<<~JS))
```
haywoood: These tests are a bit brittle, but ok

nik9000: Yeah. I thought they weren't great. Do you have any recommendations?

lib/ES/Util.pm
```diff
@@ -725,9 +740,31 @@ sub timestamp {
...
+    my $parcel_pid = fork;
+    return $parcel_pid if $parcel_pid;
+
+    close STDIN;
+    open( STDIN, "</dev/null" );
+    exec( qw(/node_modules/parcel/bin/cli.js serve
+             --public-url /guide/static/
+             --hmr-port 8001
+             -d /tmp/parcel/
+             resources/web/docs_js/index.js resources/web/styles.pcss) );
```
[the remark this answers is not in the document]
nik9000: I'll take a look. I ended up enabling it because I thought we'd start using things like auto-converting images to base 64 and stuff pretty quick.

haywoood: I think if you add `resources/web/docs_js/__tests__/**/* resources/web/docs_js/__tests__/*` here it'll make testing awesome.

nik9000:
I'm going to take a look at the pcss stuff before merging!

lib/ES/Util.pm
nik9000: Haywood and I talked about this off line and I ended up fixing it by improving the Makefile.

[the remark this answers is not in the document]
nik9000: Switching it from pcss to css turns off the minification. Not that we need the minification that badly, but I'd actually really like to guilt free add comments to the css. So I'll keep pcss.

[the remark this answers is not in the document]
nik9000: Now I've confused myself. Let me check again....

integtest/spec/single_book_spec.rb
haywoood: I don't really, other than maybe we don't need to test for it or we put some text to look for in index.js and comment it with `// used in an assertion, don't remove`

nik9000: I'll use a comment like that, yeah.

nik9000:
I'd love to merge this but I want to verify that it is working after a docs rebuild after the merge and I'm not goign to be around for that. So it'll have to wait.

PR #1057, opened 2019-08-12
Improve error message when paths don't line up

nik9000:
Improves the error message when `git revparse --show-toplevel` doesn't
match the provided path. We might be able to "heal" these bad paths
later on when we have a better sense of how they can come about and
bet


nik9000:

PR #1059, opened 2019-08-12
Break xpack tag from main CSS file

nik9000:
This breaks the xpack tag out of the main CSS file, moving it into one
that we import with the goal of making the CSS more manageable.


nik9000:
@haywoood, what do you think of this? It requires that I break out pcss but it means I can keep the actual source graphics around and it means I can break the style into individual files and organize it. I'm fairly sure parcel can do the importing without postcss but I like the less-style nesting support here.

haywoood:
Looks good. Just a comment on CSS nesting, it's super easy to do when you're writing it, but is hard to reverse engineer from looking at a template, as you're never quite sure what's affecting what. And usually, the nesting is just unnecessarily coupling the css to the dom structure. I'm personally a reformed sass nester, and flat classes `Widget-h1` have served me well. 

This article was very formative for me and basically says everything :D http://nicolasgallagher.com/about-html-semantics-front-end-architecture/ if you were every curious

resources/web/style/xpack.pcss
```diff
@@ -19,17 +19,17 @@
...
-  h3 > span.xpack,
-  h4 > span.xpack,
-  h5 > span.xpack,
-  h6 > span.xpack {
+  h1 > .xpack,
+  h2 > .xpack,
+  h3 > .xpack,
+  h4 > .xpack,
+  h5 > .xpack,
+  h6 > .xpack {
```
haywoood: Do we need the h1 specifiers as well? is `.xpack`

haywoood:
And I'm just generally a fan of regular file prefixes as a future developer concern. Having to support weird old formats when the world has moved on is a bummer.

nik9000:
> And usually, the nesting is just unnecessarily coupling the css to the dom structure.

I've tried to take that to heart after our talk last time! You can see that I stripped a bunch of extra stuff from the reset around floating titles when I did this.

The two types of nesting I have here are paranoia and convenience, I think:
* Paranoia meaning "this class will never ever ever ever leak outside of the `#guide` root element. That way it'll never infect the chrome now matter what. This is exactly the kind of thing I've asked @brianjolly to do for the chrome.
* Convenience meaning `&:hover {` inside of the tag for the thing that is hovering lets me make it super clear what that hover bit is doing.

> And I'm just generally a fan of regular file prefixes as a future developer concern. Having to support weird old formats when the world has moved on is a bummer.

Do you mean we should use postcss and keep the files `.css`? I did look into that but syntax highlighting suffers and you have to configure parcel to pick up the file.

resources/web/style/xpack.pcss
nik9000: *Probably*. It is difficult to say without digging around quite a bit to make sure the `xpack` class only exists there. We probably could clean it up if we hacked on the `xslt` some but I really try to do that as little as possible.

haywoood:
Yea I agree the convenience is great for hover and other pseudo classes.

My comment about .pcss was just a general statement. I think as long as we don't lock ourselves into the format too much, making getting off it in the future more difficult, we're good.

PR #1060, opened 2019-08-13
Fixup error handling for new port

nik9000:
Switching to two ports for `--open` broke our nice error message when
ports colide. This fixes it.


nik9000:

PR #1061, opened 2019-08-13
Enable alternative examples from C#

nik9000:
Starts pulling C# examples from the elasticsearch-net repo into the
Elasticsearch reference. They aren't visible by default but we'll be
able to switch to them eventually.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -250,21 +250,16 @@ contents:
                 repo:   elasticsearch-php
                 path:   examples/docs/src/Examples
                 exclude_branches:   [ 7.x, 7.3, 7.2, 7.1, 7.0, 6.8, 6.7, 6.6, 6.5, 6.4, 6.3, 6.2, 6.1, 6.0, 5.6, 5.5, 5.4, 5.3, 5.2, 5.1, 5.0, 2.4, 2.3, 2.2, 2.1, 2.0, 1.7, 1.6, 1.5, 1.4, 1.3, 0.90 ]
-              # Coming soon!
-              # -
-              #   alternatives: { source_lang: console, alternative_lang: csharp }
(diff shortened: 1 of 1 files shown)


conf.yaml
```diff
@@ -250,21 +250,16 @@ contents:
...
-              #     master: doc-examples
+              -
+                alternatives: { source_lang: console, alternative_lang: csharp }
+                repo:   elasticsearch-net
+                path:   examples
+                exclude_branches:   [ 7.x, 7.3, 7.2, 7.1, 7.0, 6.8, 6.7, 6.6, 6.5, 6.4, 6.3, 6.2, 6.1, 6.0, 5.6, 5.5, 5.4, 5.3, 5.2, 5.1, 5.0, 2.4, 2.3, 2.2, 2.1, 2.0, 1.7, 1.6, 1.5, 1.4, 1.3, 0.90 ]
+              -
+                # These are used by .NET's doc examples but they are not in the same subdirectory
+                repo:   elasticsearch-net
+                path:   src/Examples
```
russcam: This will find any sub directories and files from this path?

russcam:
In a future iteration, I'll update the C# doc examples generation to embed the source block directly, as opposed to using an `include` to reference from source. I'll open a PR when this change is made

nik9000:
> In a future iteration, I'll update the C# doc examples generation to embed the source block directly, as opposed to using an `include` to reference from source. I'll open a PR when this change is made

When you do so please don't remove the original directory. The docs build will fail loudly if a directory it is configured to check out doesn't exist.

conf.yaml
nik9000: Right! It'll check out those files locally before building the docs.

PR #1062, opened 2019-08-13
Automatically update version drop downs when we add or remove a version

nik9000:
I was inspired by #1055 to see if I could make it so that we don't have
to run a build with `--rebuild` when we release. I failed. But I made
progress on one front so I think it is worth merging. This change makes
it so that we rebuild the "version" drop down if we add or remove versions
from a book. This'll probably allow us to skip running `--rebuild` when we
add new branches but it isn't enough to stop us from having to run a
`--rebuild` with every release because:

1. It doesn't rebuild the version drop down if we only switch the current
version to another version.
2. It won't rebuild the headers that tell you that you are reading docs
for an old version or for an unreleased version when we switch the current
version to another version.

But it *is* a start.


lib/ES/Book.pm
```diff
@@ -209,12 +209,18 @@ sub build {
         }
     }
     $pm->wait_all_children();
-    $self->_copy_branch_to_current( $self->current ) if $rebuilding_current_branch;
-    $self->remove_old_branches;
+    $self->_copy_branch_to_current if $rebuilding_current_branch;
+    $update_version_toc |= $self->_remove_old_branches;
     if ( $self->is_multi_version ) {
-        if ( $rebuilding_any_branch ) {
-            printf(" - %40.40s: Writing versions TOC\n", $self->title);
```
nik9000: I don't think the prints here add that much value because this process is pretty quick.

lib/ES/Book.pm
```diff
@@ -333,49 +337,39 @@ sub _build_book {
 }
 
 #===================================
-sub _add_title_to_toc {
+sub _update_title_and_version_drop_downs {
 #===================================
-    my ( $self, $branch, $dir ) = @_;
-    my $title = $self->title;
-    if ( $self->is_multi_version ) {
```
nik9000: We never do this if the book *isn't* multi-version.

nik9000:
It looks like this eats part of the TOC. I'll investigate.

nik9000:
@elasticmachine, run elasticsearch-ci/docs with rebuild

PR #1064, opened 2019-08-13
Convert inline admonishments to components

nik9000:
Converts inline admonishments into "component" style html using the
guidelines laid out here:
http://nicolasgallagher.com/about-html-semantics-front-end-architecture/

### DIFF
--- a/resources/web/style/admonishment.pcss
+++ b/resources/web/style/admonishment.pcss
@@ -1,6 +1,5 @@
 #guide .Admonishment {
   position: relative;
-  font-size: 16px;
 
   &:hover {
     color: black;
@@ -15,12 +14,8 @@
   }
 
   &-title, &-version {
-    font-family:Monaco, 'DejaVu Sans Mono', 'Bitstream Vera Sans Mono', 'Lucida Console', monospace!important;
     font-size: 0.8em;
     border-bottom: 1px dashed #00a9e5;
-  }
-  &-version--deprecated {
-    text-decoration: line-through;
   }
 
   &-detail {
--- a/resources/web/style/admonishment.pcss
+++ b/resources/web/style/admonishment.pcss
@@ -1,0 +1,38 @@
+#guide .Admonishment {
+  position: relative;
+  font-size: 16px;
+
+  &:hover {
+    color: black;
+    transition: all 0.3s ease;
+    .Admonishment-detail {
+      z-index: 10000;
+      left: -5em;
+      opacity: 1;
+      color: #555;
+      transition: color 0.5s ease;
+    }
+  }
+
+  &-title, &-version {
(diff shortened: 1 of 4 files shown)


nik9000:
This'll require a full `--rebuild` rebuild because it changes the xslt as well as the CSS.

nik9000:

resources/website_common.xsl
```diff
@@ -344,9 +344,9 @@
         <xsl:when test="attribute::revisionflag='deleted'">deprecated</xsl:when>
       </xsl:choose>
     </xsl:variable>
-    <span class="{$classname}">
-      [<span class="version"><xsl:value-of select="attribute::revision" /></span>]
-      <span class="detail">
+    <span class="Admonishment Admonishment--{$classname}">
+      [<span class="Admonishment-version Admonishment-version--{$classname}"><xsl:value-of select="attribute::revision" /></span>]
+      <span class="Admonishment-detail">
```
haywoood: Nice!

resources/web/style/admonishment.pcss
```diff
@@ -0,0 +1,38 @@
...
+    .Admonishment-detail {
+      z-index: 10000;
+      left: -5em;
+      opacity: 1;
+      color: #555;
+      transition: color 0.5s ease;
+    }
+  }
+
+  &-title, &-version {
```
haywoood: Oh, that's cool!

nik9000: I like it! Nice and simple way to contain the component. It doesn't *always* work as you can see inside the style that is inside the hover.

resources/web/style/admonishment.pcss
```diff
@@ -0,0 +1,38 @@
...
+    font-family:Monaco, 'DejaVu Sans Mono', 'Bitstream Vera Sans Mono', 'Lucida Console', monospace!important;
```
haywoood: Just something to think about, generalized classes like maybe `u-monotype` or something that does the font-family and font-size may be helpful. We could end up with a bucket of general classes that can be combined for the effects we want in templates.

Nothing actionable, just something to think about when looking at CSS definitions.

nik9000: ++

resources/web/style/admonishment.pcss
```diff
@@ -0,0 +1,38 @@
...
+    font-size: 0.8em;
+    border-bottom: 1px dashed #00a9e5;
+  }
+  &-version--deprecated {
+    text-decoration: line-through;
```
haywoood: Continuing the line of thought above, what value does this have over a class like say, `u-strikethrough` or something.

nik9000: I don't think it has any beyond the "traditional" separation or style and markup. I'll see about adding those utilities before merging.

nik9000:

nik9000:
@haywoood, have a look now!

nik9000:

nik9000:

PR #1068, opened 2019-08-15
Switch callouts from images to css

nik9000:
This switches the callout numbers from images to css-managed text. It is
still copy-and-paste safe.

### DIFF
--- a/resources/website_common.xsl
+++ b/resources/website_common.xsl
@@ -329,10 +329,11 @@
     <!-- Throw away the link by not copying it. -->
   </xsl:template>
 
-  <!-- Make callouts non-selectable -->
+  <!-- Use asciidoctor style callouts -->
   <xsl:template name="callout-bug">
     <xsl:param name="conum" select="1"/>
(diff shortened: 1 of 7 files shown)


resources/website_common.xsl
```diff
@@ -329,10 +329,11 @@
     <!-- Throw away the link by not copying it. -->
   </xsl:template>
 
-  <!-- Make callouts non-selectable -->
+  <!-- Use asciidoctor style callouts -->
   <xsl:template name="callout-bug">
     <xsl:param name="conum" select="1"/>
-    <span><img src="{$callout.graphics.path}{$conum}{$callout.graphics.extension}" alt="" /></span>
+    <!-- <i> is a hack here to avoid all of the styling that comes in with a <span>....-->
+    <i class="conum" data-value="{$conum}"></i>
```
nik9000: @brianjolly, this is caused by the styles i was telling you about sneaking in from elastic-nav.js.

nik9000:
> This switches the callout numbers from images to css-managed text. It is
> still copy-and-paste safe.

It also switches the callouts to a different style as provided by @Adrian-J .

nik9000:

PR #1069, opened 2019-08-15
Style: Fix up code blocks

nik9000:
We need to be pretty specific in order for code blocks to get the right
style because of the global `*` rule.


brianjolly:
Hi @nik9000, this change looks good to me!

brianjolly:
LGTM!

nik9000:

PR #1070, opened 2019-08-15
Fix the ordering of alternative examples

nik9000:
This changes the order of the html generated by alternative examples so
it is simpler to switch to different alternatives. Instead of the html
looking like:

```
default example
console widget
default callout list
alternative 1 example
alternative 1 callout list
alternative 2 example
alternative 2 callout list
```

It now looks like:

```
alternative 1 example
alternative 2 example
default example
console widget
alternative 1 callout list
alternative 2 callout list
default callout list
```

The advantage of this order is that it works with how we want to display
these examples: we always hide all but one of the examples and all but
one of the callout list and we want the order of shown things to always
be:

```
example
console widget
callout list
```


nik9000:
@estolfo, could you have a look at this when you get a chance? It is another step in the client examples project!

nik9000:

PR #1071, opened 2019-08-15
Extract list styling into its own file

nik9000:
It'll make our life simpler to have the styles grouped. This doesn't
simplify the rules, just moves them.


nik9000:

PR #1072, opened 2019-08-15
Correct style on gear icon

nik9000:
Make sure we switch to the pointer and don't underline.


nik9000:
Thanks for finding this @haywoood !

PR #1076, opened 2019-08-16
Build initial js state

nik9000:
We're getting to the point where we have to communicate specific
information about an entire book to the js on a particular page,
specifically, we need to tell the js which "alternative languages" are
available for examples. This creates an HTML tag that looks like this:

```
<script type="text/javascript">
window.initial_state = {
   "alternatives" : {
      "console" : {
         "cat" : {
            "hasAny" : false
         },
         "csharp" : {
            "hasAny" : true
         },
         "js" : {
            "hasAny" : true
         }
      }
   }
}
</script>
```

that is populated at template application time. In particular, it is
built by consuming a json file that is built by the asciidoctor process
that contains a summary of the alternative language snippets process.
That summary contains more precise information that we intentionally are
not including on the page because it'd cause the diffs to "jump around"
too much. Here is the summary:

```
{
  "console": {
    "total": 1,
    "alternatives": {
      "csharp": {
        "found": 1
      },
      "js": {
        "found": 1
      },
      "cat": {
        "found": 0
      }
    }
  }
}
```

Note the precise counts. Those *might* be useful to have in the js but
we'd prefer not to send them to keep the diffs smaller. Maybe one day
we'll apply the template on the fly and include them.....


nik9000:
:fries:

PR #1079, opened 2019-08-17
Move toc style to its own file

nik9000:
This moves and *slightly* simplifies the style for the table of
contents. It is *super* nesty, mostly because that is how the old styles
were laid out. We expect to rework the toc at some point so it isn'
worth doing any major surgury to it now. But isolating it seems good.


resources/web/style/toc.pcss
```diff
@@ -0,0 +1,86 @@
+#guide {
+  .toc {
+    ul {
+      margin: 0 0 1em 0;
+      padding: 0;
+      border: 1px solid #ddd;
+      ul {
+        /* Hide all sub-toc elements by default. See rules with `.show` that
```
nik9000: I feel like being able to add comments here and know they won't bloat the download is a nice win.

resources/web/style/toc.pcss
```diff
@@ -0,0 +1,86 @@
...
+      }
+    }
+
+    /* Customize each level of the TOC, mostly so it looks "indented". */
+    > li {
+      > span {
+        background-color: #efefef;
+        border-bottom: 1px solid #ddd;
+        font-size: 1em;
+        background-position: 0 11px;
```
nik9000: These background positions aren't the same as they used to be. I didn't like how they lined up before and they were defined all over the place. I've grouped them here which makes me feel much better about them.

nik9000:
:fries:

PR #1081, opened 2019-08-19
Drop guide.css

### LINKED ISSUE #1067: Drop guide.css
The docs use styles from https://www.elastic.co/static/css/guide.css which is sort of against our desire to keep the style for the docs in the docs where possible.

nik9000:
Removes our use of `guide.css` which is a css file maintained by the web
team for use with the docs. This merges all of the CSS that it contained
into our css. It tweaks a few small things that I saw while working on
it, like:

1. The screen-size dependant styles for the nav header didn't also apply
to the footer. Now they do.
2. We had a rule that attempted to override the font of the headers but
it wasn't taking. This comments it out so there isn't any confusion and
calls it out for future investigation.

It also factors out some more css into files, specifically the styles
for headers, the "rtpcontainer", and the nav. The "rtpcontainer" is a
little mysterious to me so at the moment I've preserved all of the
styles that it shipped with even though they aren't all used. I bet they
are used in a less than obvious way involving afer-the-fact changes that
don't happen when you render locally. We'll have to get a better handle
of those eventually, but for now we let them stay a mystery.

Closes #1067


brianjolly:
Hi @nik9000! This looks good, but can we push to staging to review there?

nik9000:
http://docs_1081.docs-preview.app.elstc.co/guide should have a preview. Do you need it specifically on staging or is that ok?

brianjolly:
This will work good. I'll check it out there. Thanks @nik9000!

brianjolly:
Hi @nik9000 , I just noticed a few things. Let me know how you want to handle, i.e. if you want some help to address any of these. 

### Titles are using `Inter`.

With the redesign we are now using:
`aktiv-grotesk` for titles
`Inter` for body text
Screen Shot 2019-08-20 at 1 57 59 PM

---
### Footer's signup style looks like the previous design:

Screen Shot 2019-08-20 at 2 58 26 PM

---
### We could update to the new link style:
Old design:

Screen Shot 2019-08-20 at 3 02 54 PM

Redesign:

Screen Shot 2019-08-20 at 3 03 02 PM

nik9000:
> Titles are using `Inter`.

In the docs right now titles are using Inter. I left a comment about how confused I was that it looked like we tried to use `Aktiv Grotesk` but didn't succeed. I can make that change as part of this. But it is a change.

> Footer's signup style looks like the previous design:

I hadn't noticed this part! I'll check into it. I wonder what happened here....

nik9000:
> We could update to the new link style:

I'd like to do that in a separate PR. I kind of wanted to change as little as possible with this PR.

> Footer's signup style looks like the previous design:

It looks like the docs look this way now. I'd *prefer* to handle that in a follow up change.

brianjolly:
That makes sense @nik9000. We could go ahead and merge, then handle those issues separately.

PR #1085, opened 2019-08-20
Let folks chose their favorite language

nik9000:
Allows folks to pick the language they'd like to see instead of "Console"
from a drop down under every "Console" snippet. The drop down includes
every language for which we have alternatives and doesn't show up in
books that don't have alaternative languages configured for "Console".

This is really step one of exposing this feature. There are a few more
things to do:

1. Add a text "bubble" above the warning icon explaining why it shows up.
It usually shows up because the user has picked a language that doesn't
have any examples for that particular snippet.
2. Prevent the "jumping" around that happens when we switch examples.
This is caused by the examples not having the same length.
3. Clean up the "funny" case when a user has a preference for a language
that doesn't show up in the book.
4. Probably performance improvements.
5. Almost certainly style polishing.


nik9000:
This is a WIP because it super deserves tests but doesn't have any!

nik9000:
[Preview](http://docs_1085.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/docs-get.html)

resources/web/docs_js/components/console_widget.jsx
```diff
@@ -69,23 +69,80 @@ export const ConsoleForm = connect((state, props) =>
...
-    </a>
-    {props.view_in_text &&
-      <a className="view_in_link"
-         target="console"
-         title={props.langStrings(props.view_in_text)}
-         href={`${props[props.setting + "_url"]}?load_from=${props.baseUrl}${props.snippet}`}>{props.langStrings(props.view_in_text)}</a>
-    }
-    <a className="console_settings" onClick={modalAction} title={props.langStrings(props.configure_text)}>&nbsp;</a>
+  return <div className="u-space-between">
+    {alternativePicker(props) || <div/>}
```
haywoood: Since it's a component we should 'render' it `<AlternativePicker {...props} />`

resources/web/docs_js/components/console_widget.jsx
```diff
@@ -69,23 +69,80 @@ export const ConsoleForm = connect((state, props) =>
...
+  if (!consoleAlternatives) {
+    return;
+  }
+
+  const items = [];
+  let sawChoice = 'console' === props.consoleAlternative;
+  items.push(alternativeChoice('console'));
+  for (const name of Object.keys(consoleAlternatives)) {
+    sawChoice |= name === props.consoleAlternative;
+    items.push(alternativeChoice(name));
```
haywoood: `items.push(<AlternativeChoice name={name} />)` and in the function grab the name off props

resources/web/docs_js/components/console_widget.jsx
haywoood: and maybe it can return an empty div instead of the or here.

resources/web/docs_js/index.js
```diff
@@ -161,8 +162,17 @@ $(function() {
       sense_url: Cookies.get("sense_url") || default_sense_url,
       sense_curl_host: Cookies.get("sense_curl_host") || "localhost:9200",
       sense_curl_user: Cookies.get("sense_curl_user"),
-      sense_curl_password: "$ESPASS"//TODO Cookies.get("curl_password")
-    }
+      sense_curl_password: "$ESPASS",//TODO Cookies.get("curl_password")
```
haywoood: can probably nix the todo here

resources/web/docs_js/components/console_widget.jsx
haywoood: I would actually change this to `<AlternativePicker />`, remove ` {alternatives: state.alternatives}` from map state to props below, and properly connect the AlternativePicker to grab `state.alternatives` and `state.settings.consoleAlternative`. That will move the concerns to the component that cares about it, and it's a good practice to get in the habit of.

resources/web/docs_js/__tests__/components/console_widget.test.js
```diff
@@ -0,0 +1,105 @@
...
+          cshell: {},
+          emacs: {},
+          powershell: {},
+        }
+      }}/>);
+      test("is just an empty div", () => {
+        expect(picker).toStrictEqual(render(<div/>));
+      });
+    });
+    describe("when the there are console alternatives", () => {
```
russcam: slight nitpick:

```suggestion
    describe("when there are console alternatives", () => {
```

nik9000:
OK! I'm feeling much more comfortable about this now. It isn't *done*, but might be close enough given that it is pretty much disabled outside of the master branch.

@haywoood, could you give this a real review when you get a chance?

nik9000:
I say that and then testing it locally it immediately breaks....

nik9000:
And Fixed!

nik9000:
Thanks so much for reviewing @haywoood and @russcam! I'll merge this and then iterate on things like the style for mobile.

PR #1086, opened 2019-08-21
Drop source maps from production web resources

nik9000:
We don't ship the original files so the source map fails to load anyway.

### DIFF
--- a/lib/ES/Util.pm
+++ b/lib/ES/Util.pm
@@ -760,14 +760,16 @@ sub build_web_resources {
     my ( $dest ) = @_;
 
     run '/node_modules/parcel/bin/cli.js', 'build',
-        '-d', $dest, '-o', 'docs.js', '--experimental-scope-hoisting',
+        '--experimental-scope-hoisting', '--no-source-maps',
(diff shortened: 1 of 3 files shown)


lib/ES/Util.pm
```diff
@@ -760,14 +760,16 @@ sub build_web_resources {
...
     run '/node_modules/parcel/bin/cli.js', 'build',
-        '-d', $dest, '-o', 'docs.js', '--experimental-scope-hoisting',
+        '--experimental-scope-hoisting', '--no-source-maps',
+        '-d', $dest, '-o', 'docs.js',
         'resources/web/docs_js/index.js', '/node_modules';
     my $docs = $dest->file('docs.js');
     my $licenses = file('resources/web/docs.js.licenses')->slurp;
-    my $minified = $docs->slurp;
-    $docs->spew($licenses . $minified);
+    my $built = $docs->slurp;
```
nik9000: I just didn't like this variable name! I know I wrote it in the first place but parcel is doing a lot more than minimizing.

nik9000:
:fries:

PR #1087, opened 2019-08-21
Commit less

nik9000:
Adds a new mechanism to detect when it is time to commit changes.
Previously, we only used `git status` to look for changes. This first
consults the branch tracker to see if any books were rebuilt with
non-`--sub_dir` based changes. If any were then we can skip the
`git status` entirely!

This allows us to wait until *after* we've decided to commit before
writing the branch tracker and the revisions helper file. This'll save
some time, especially on PR builds that are noops. It should also allow
us to consider many more PR builds to be noops then we could before! It
used to be that we're write information about the "--sub-dir"-ness of
the build to the branch tracker file and that'd *always* cause us to
think that PR builds changed docs even if they didn't. Now PR builds
won't trigger a commit unless they actually change rendered html.
:party:

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -268,8 +268,7 @@ sub build_all {
         check_links($build_dir);
     }
     $tracker->prune_out_of_date( @$contents );
-    $tracker->write;
-    push_changes( $build_dir, $target_repo ) if $Opts->{push};
+    push_changes( $build_dir, $target_repo, $tracker ) if $Opts->{push};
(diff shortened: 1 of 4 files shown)


nik9000:
@haywoood, could you have a look at this one? It doesn't have any JS but you've read quite a bit of the perl at this point to so you are *probably* the best person to review it.

nik9000:
This one is for you, @talevy. Thanks for asking me questions!

PR #1090, opened 2019-08-22
Fix console widget style on mobile

nik9000:
This changes the console widget style so it looks better on mobile. It
1. Shrinks the link text slightly when the width shrinks. This is the
same shrinking that we have for the nav header and footer.
2. When there isn't enough space for the links they now wrap "inside"
themselves.
3. Switch the "bad language" warning from "not displayed" to "not
visible" to it takes up space all the time. This should prevent weird
stuff happening when it needs to be shown on mobile.


nik9000:
:fries:

PR #1088, opened 2019-08-22
Python Client - code snippets

philkra:
The PR adds the code snippets for the ES Python Client for the following doc pages:
* Index API
* Get API
* Delete API
* Getting Started

PR #1091, opened 2019-08-22
Add python's pretty name to alternatives drop down

nik9000:
We should probably be using the language string but we'll get there
eventually. For now, lets just make Python look a little nicer, shal we?


nik9000:
It'll make #1088 a little prettier.

nik9000:
@elasticmachine, retest this please

nik9000:
:fries:

PR #1092, opened 2019-08-22
ECS 1.1.0 docs for release

webmat:
Long time fan, first pull request to docs :-)

I'm in the process of releasing ECS 1.1.0, which has a live `1.1` branch, for docs adjustments.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -142,8 +142,8 @@ contents:
           -
             title:      Elastic Common Schema (ECS) Reference
             prefix:     en/ecs
-            current:    1.0
-            branches:   [ master, 1.0 ]
+            current:    1.1
+            branches:   [ master, 1.1, 1.0 ]
             index:      docs/index.asciidoc
             chunk:      1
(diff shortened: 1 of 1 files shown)


nik9000:

webmat:
I've looked at the docs for 1.1 and current, and both look good to me.

However at the bottom of the diffs I can see changes to the logstash docs.

@nik9000 are these just recent merges that are about to be deployed to the main site? Kind of like a race condition I shouldn't worry about?

If the logstash docs diff isn't a worry, I'm good to merge this. But since I'm not familiar with the repo & stuff, I'd love to get confirmation that this is fine :-)

nik9000:
Indeed, sometimes things will sneak in to the diff, especially in the docs
repo. This is the price we pay to test new books and new branches and new
sources. This doesn't happen in the prs to the other repos because they
don't have to test new sources and stuff.

webmat:
Great, thank you @karenzone and @nik9000

PR #1093, opened 2019-08-22
Copy images from unordeed lists

nik9000:
We had support for copying images within ordered lists but not within
*un*ordered lists. This fixes that.


nik9000:
@haywoood, do you have time to look at this one?

nik9000:
There is more to do here.....

nik9000:
I think I've got it now!

nik9000:

I want to make super sure this is ok on all the books, just in case.

nik9000:

nik9000:
The failure here is real! I'm tracking it down. Weird stuff.

PR #1095, opened 2019-08-23
Prevent jumping when you pick an alternative lang

nik9000:
When you pick an alternative language the page has to reflow because the
alternatives have different vertical size. This reflow looks like a big
"jump", especially when you pick the language lower down on the page.
This prevents the "jump" by storing the location of the picker before
the reflow, updating the styles to cause the reflow, grabbing the new
location, and then scrolling the window the difference between the new
and old location. This keeps the picker, which is usually right under
your mouse when you change the language, from jumping. This "feels"
smooth!

### DIFF
--- a/resources/web/docs_js/__tests__/components/alternative_picker.test.jsx
+++ b/resources/web/docs_js/__tests__/components/alternative_picker.test.jsx
@@ -86,13 +86,10 @@ describe(_AlternativePicker, () => {
       * have to do it ourselves. Like an animal. */
     select.dispatchEvent(new Event("change"));
     test("saves the config", () => {
+      // We can't use toMatchObject here because tagName is busted.
+      expect(updates).toHaveLength(1);
+      expect(updates[0].consoleAlternative).toBe("js");
       expect(updates[0].alternativeChangeSource.tagName).toBe("SELECT");
-      expect(updates).toMatchObject([{
-        consoleAlternative: "js",
-        alternativeChangeSource: {
-          className: "SELECT",
-        }
-      }]);
     })
   });
 });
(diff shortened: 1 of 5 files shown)


nik9000:
I've not done any work on tests because I'd like @haywoood and maybe @JoshMock to have a look at the idea. I *could* have done this all in `alternative_picker.jsx` but that feels wrong because the style changing is actually done in the switcher. It feels a little weird to stick a dom element in the redux state but it gets the job done. Is there a cleaner way to do this?

haywoood:
Definitely a picky problem to deal with. As far as the approach, I think it gets the job done in an okay way since we're tending to isolate a lot of the 'messy side effect' logic to that one watcher function. It's sort of a bummer we need to overload the generic 'saveSettings` action with this concern, but I don't think it's worth making a new action just for this.

haywoood:
Would something like `element.scrollIntoView` work instead? Then maybe we just look for a `scrollElement` value in the store, and if it's there, call `store.getState().scrollElement.scrollIntoView()` at the end of our loop.

nik9000:
The trouble with `scrollIntoView` is that I want to make sure we don't jump at all if possible. I certainly could put the alternative source into another part of the state and make a reducer for it that just saves it. I already load the alternatives state at page load. I could do it there instead of in the settings. Then we'd have to make two state changes and I'd want to make sure the listener was triggered after both of them.

haywoood:
Got it, like I said I think having it in settings is okay.

nik9000:
> Got it, like I said I think having it in settings is okay.

OK. I'll see about making the tests happy.

nik9000:
And I haven't tested that visual-only bit locally yet. Let me get on that....

nik9000:
> And I haven't tested that visual-only bit locally yet. Let me get on that....

All done. Still works smooth as silk. In Firefox at least. Probably worth me digging out other browsers.....

nik9000:
:fries:

PR #1098, opened 2019-08-23
Fix warning

nik9000:
Fix a warning emitted when including alternative snippet. It isn't
causing any failure but it is making debugging failures more difficult.


nik9000:
:fries:

PR #1099, opened 2019-08-23
Style: Move the * rule into guide

nik9000:
This should prevent us from modifying the header and footer.


nik9000:
I'm think it might be nice to remove the `*` from that rule to be honest. That way we inherit properly. But I think that is a bigger change.

nik9000:
So I figured I could make this fairly small change for now.

brianjolly:
Hi @nik9000 , this looks good to me :)

nik9000:
Thanks for reviewing @brianjolly !

PR #1100, opened 2019-08-23
Fix Copy as cURL's default port for kibana

nik9000:
It was starting with Elasticsearch's port.


nik9000:
Thanks for pointing this out and reviewing it @spinscale!

PR #1102, opened 2019-08-23
Allow clients to specify alternative responses

nik9000:
This changes the "alternative language lookup" process to look for the
special suffix `-result` on the end of a language and look in the
non-`result` spot. For example, if a snippet has the langauge
`console-result` then we'll look up the snippet for all configured
alternatives for `console`.

This is useful because there are many, many "result" snippets in the
Elasticsearch reference and it looks funny today when you select, say,
"PHP" from the alternative language drop down and the *request* is in
PHP but the result of the request is in JSON. With this change all
clients, including the PHP client, will be able to provide stand ins for
all of these results snippets.

Well, all of the results snippets that are labeled explicitly as
results. Right now no snippets in the Elasticsearch reference are
labeled that way but we'll change that soon!

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -144,6 +144,7 @@ book:
 build_docs --doc path/to/index.asciidoc --chunk 1
 ----------------------------
 
+[[alternative_languages]]
 === Alternative languages for examples
 
 The build supports finding "alternative languages" for examples that allows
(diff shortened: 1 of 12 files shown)


nik9000:
We'll need https://github.com/elastic/elasticsearch/pull/45937 are more like it to be able to use this.

PR #1110, opened 2019-08-28
Shrink diffs for sitemap.xml

### LINKED ISSUE #684: Dates in sitemap.xml
Right now every time we generate new docs we rebuild the `sitemap.xml` setting every file to last edited "now". This makes our diffs pretty big and isn't really true. I wonder if it'd be ok if didn't bump the time if we didn't rebuild the book.

nik9000:
We generate `sitemap.xml` with an entry for every page in every
"current" book. Before this change we'd set the last modified time to
"now" for every single page, every time we rebuild the books. This
changes it so that we only change the last modified time when we modify
the file which is how you are supposed to use `sitemap.xml` anyway.

Beyond that, it sorts the entries in the sitemap before building it so
it doesn't "jump arround" on every commit.

These two combine to significantly shrink the size of small diffs which
should make working with the docs just a little bit faster.

Closes #684

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -431,7 +431,6 @@ sub build_sitemap {
     }
     for ( split /\0/, $changed ) {
         next unless s|^html/||;
-        say "ASDFADF $_";
         $dates{$_} = $now;
     }
 
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -17,6 +17,7 @@ our @Old_ARGV = @ARGV;
 use Cwd;
 use FindBin;
 use Data::Dumper;
+use XML::LibXML;
 
 BEGIN {
     $Old_Pwd = Cwd::cwd();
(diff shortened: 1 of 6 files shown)


build_docs.pl
```diff
@@ -402,31 +403,41 @@ sub build_entries {
...
+            }
+            die "Dind't find <loc> in $_" unless $loc;
+            die "Dind't find <lastmod> in $_" unless $lastmod;
+            $loc =~ s|https://www.elastic.co/guide/||;
+            $dates{$loc} = $lastmod;
+        }
+    }
+    for ( split /\0/, $changed ) {
+        next unless s|^html/||;
+        say "ASDFADF $_";
```
olksdr: debug `say` ?

nik9000: yeah! let me zap it.

nik9000:
Thanks for reviewing @olksdr! I'm going to enjoy the smaller diffs we git from this one!

PR #1115, opened 2019-08-29
Merge template customization into template

nik9000:
Now that the web and docs teams are collaborating more closely we
don't really *need* to patch the template on startup. Instead we can put
all of our patches into the template file directly and everything is
just a little bit clearer.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -1541,7 +1541,3 @@ redirects:
 
 template:
     defaults:
-        POSTHEAD: |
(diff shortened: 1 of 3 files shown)


nik9000:
@brianjolly, I'd like to simplify our templating mechanism and this seemed like a good first step. How do you feel about this?

brianjolly:
Hi @nik9000 , this makes sense to me :)

nik9000:

PR #1117, opened 2019-08-29
Move the repos dir config to parameter

nik9000:
Removes the `paths.repos` config entry from the conf file because we
rarely want to change it outside of tests. Adds a command line parameter
to change the path (`--reposcache`) and uses it in the tests.

This sort of accidentally fixes a problem where running `--self-test`
will trash the contents of `.repos/target_repo.git`, forcing subsequent
`--all` builds on the same machine to fetch it.

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -452,9 +452,7 @@ ENTRY
 #===================================
 sub init_dirs {
 #===================================
-    my $repos_dir = $Conf->{paths}{repos}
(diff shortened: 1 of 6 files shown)


nik9000:
Oooops test failure.

nik9000:
That should do it!

nik9000:
> Perl part - lgtm! :shipit:

Thanks for reviewing!

PR #1118, opened 2019-08-29
Keep a copy of the raw html

nik9000:
This changes the book building process so that it writes the output to a
`raw` directory and then copies that directory to the `html` directory
that we are so used to, applying the template on the fly.

Right now this just makes more files, probably slowing down the build
process. But I have plans! It'd be really nice to be able to apply the
template without rebuilding the entire book and this would make that
possible.


nik9000:
Interesting failure! I'll look into it.

nik9000:
I *think* that will fix it!

nik9000:
Thanks for reviewing @olksdr!

PR #1125, opened 2019-08-30
Stop the tracker from lying about changes

nik9000:
We set a flag on the branch tracker whenever there are changes to one of
the branches that we use to build the docs. When we're done building the
docs we commit them if *either* the flag is set or the built resulted in
any changed files. We do this because we want to commit the branch
changes even if there aren't changed files so we know were to start from
on the next build.

This fixes a problem where the tracker would lie about there being
changes when there weren't any. In particular, the kibana link checking
would set the branch even if it hadn't changed and that triggered the
tracker to *think* that it had changes. That caused us to try and
`git commit` when there aren't actually any changes. This fixes that.

Previously this wasn't causing trouble because we would rebuild the
entire sitemap if the tracker changed, causing a commit with just the
sitemap. We fixed that in #1110. With this change we can entirely skip
the commit.

### DIFF
--- a/integtest/spec/all_books_change_detection_spec.rb
+++ b/integtest/spec/all_books_change_detection_spec.rb
@@ -293,11 +293,40 @@ RSpec.describe 'building all books' do
           build_one_book_out_of_one_repo_twice
           include_examples 'second build is noop'
         end
+        context 'even when there are no changes to the kibana repo' do
+          build_one_book_out_of_one_repo_twice(
+            before_first_build: lambda do |src, _config|
+              # The kibana link checking requires a few things:
+              # 1. A repo named kibana
+              # 2. A special links file in the repo
+              # 3. A book at `en/kibana`
+              kibana_repo = src.repo_with_index 'kibana', 'words'
+              kibana_repo.write(
+                'src/ui/public/documentation_links/documentation_links.ts',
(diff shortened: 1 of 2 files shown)


integtest/spec/all_books_change_detection_spec.rb
```diff
@@ -303,6 +322,16 @@ def chapter(index)
           )
           include_examples 'second build is noop'
         end
+        context 'even when there are noop changes source repo' do
```
nik9000: This doesn't actually reproduce the problem but it was what I tried first when I realized we didn't have it. i think it is worth having.

integtest/spec/all_books_change_detection_spec.rb
```diff
@@ -293,6 +293,25 @@ def chapter(index)
           build_one_book_out_of_one_repo_twice
           include_examples 'second build is noop'
         end
+        context 'even when there are no changes to the kibana repo' do
```
nik9000: This actually reproduces the issue.

nik9000:
@olksdr, since no good dead goes unpunished, fixing the sitemap generation started causing intermittent failures in the main build. I tracked them down to this.

nik9000:
> nice find! +1

terrifying find

PR #1127, opened 2019-08-31
Drop main.css from elastic.co in favor of docs css

brianjolly:
This drops main.css from the elastic.co requests. 

Testing to see if this is possible yet.


brianjolly:
Thanks @nik9000! From what I see it looks good too :)

PR #1128, opened 2019-09-03
Skip new --sub_dirs unless all sources are present

nik9000:
When we added a new branch to the list of books last week we broke the
docs PR tests for books that were involved in more than one repo. The
build saw `--sub_dir` for *that* repo and no other history so it tried
to build those multi-source books with just the `--sub_dir`ed repo. This
didn't work because we were missing a ton of files.

This detects when `--sub_dir` is used for a new source and skips
building the book *unless* all of the sources are new *and*
`--sub_dir`ed. This means we'll still build, say, the Elasticsearch
reference on Elasticsearch PRs even if the book is new but we won't
build, say, the stack overview. So we lose test coverage here but in
exchange we don't fail builds.

The more paranoid option would be to "fail closed" and log a more
helpful message but fail the build until the "global" docs build has
run. This doesn't seem like the right option because, frankly, we'd
prefer to keep merging a little smoother on release day at the cost of
failing the global docs build a little more. We are used to some trouble
with the global docs build on release day.

### DIFF
--- a/integtest/spec/all_books_change_detection_spec.rb
+++ b/integtest/spec/all_books_change_detection_spec.rb
@@ -111,17 +111,17 @@ RSpec.describe 'building all books' do
     end
 
     def self.build_one_book_out_of_two_repos_twice(
-        before_first_build: ->(src) {},
-        before_second_build: ->(src) {}
+        before_first_build: ->(src, config) {},
+        before_second_build: ->(src, config) {}
       )
       build_twice(
-        before_first_build: lambda do |src, _config|
+        before_first_build: lambda do |src, config|
           init_include src
           # Allow the caller to customize the source
-          before_first_build.call src
+          before_first_build.call src, config
         end,
-        before_second_build: lambda do |src, _config|
-          before_second_build.call src
+        before_second_build: lambda do |src, config|
+          before_second_build.call src, config
         end
       )
       include_context 'build one book twice'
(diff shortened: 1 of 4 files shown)


integtest/spec/all_books_change_detection_spec.rb
```diff
@@ -351,6 +351,32 @@ def chapter(index)
           )
           include_examples 'second build is noop'
         end
+        context 'when --keep_hash is specified and there is a new branch' do
```
nik9000: This was my first attempt to reproduce the problem. It didn't, but it did catch some interesting cases while I was working on this issue so I want to keep it.

integtest/spec/all_books_change_detection_spec.rb
```diff
@@ -545,6 +571,53 @@ def chapter(index)
           include_examples 'toc and version drop down'
           let(:current) { 'master' }
         end
+        context 'because we add a branch to the book and specify ' \
+                '--keep_hash and --sub_dir' do
```
nik9000: Neither of these reproduced the problem either but they were useful it making sure I didn't break anything while working on it.

integtest/spec/all_books_change_detection_spec.rb
```diff
@@ -595,26 +676,45 @@ def chapter(index)
...
+            before_second_build: lambda do |src, _config|
               repo2 = src.repo 'repo2'
               repo2.write 'garbage', 'junk'
               repo2.commit 'adding junk'
             end
           )
           include_examples 'second build is noop'
         end
+        context 'because we add a new branch and specify --keep_hash ' \
+                'and --sub_dir on the new branch with one repo' do
```
nik9000: This reproduced the problem we saw on Thursday with the elasticsearch and kibana repos. The book was built from more than one source, one of which was `--sub_dir`ed.

integtest/spec/all_books_change_detection_spec.rb
```diff
@@ -595,26 +676,45 @@ def chapter(index)
...
+              expect(out).not_to include('Test: Building master...')
+            end
+          end
+        end
       end
       context "when the second build isn't a noop" do
         context 'because the index repo changes' do
           build_one_book_out_of_two_repos_twice(
-            before_second_build: lambda do |src|
+            before_second_build: lambda do |src, _config|
```
nik9000: Ceremony for the ruby gods.

integtest/spec/all_books_change_detection_spec.rb
```diff
@@ -665,6 +765,43 @@ def chapter(index)
...
+          context 'the second build' do
+            let(:out) { outputs[1] }
+            include_examples 'commits changes'
+            it "doesn't print that it is building the original branch" do
+              expect(out).not_to include('Test: Building master...')
+            end
+          end
+        end
+        context 'because we add a new branch and specify --keep_hash ' \
+                'and --sub_dir on the new branch with both repos' do
```
nik9000: This isn't really going to come up much but I like it for logical consistency.

lib/ES/Repo.pm
```diff
@@ -124,7 +134,7 @@ sub extract {
     my ( $title, $branch, $path, $dest ) = @_;
 
     if ( exists $self->{sub_dirs}->{$branch} ) {
-        # Copies the $path from the subsitution diretory. It is tempting to
+        # Copies the $path from the subsitution directory. It is tempting to
         # just symlink the substitution directoriy into the destionation and
```
olksdr:
```suggestion
        # just symlink the substitution directory into the destination and
```

I know this is not it the changes, but it caught my eye

nik9000:
Thanks for reviewing @olksdr!

PR #1131, opened 2019-09-03
Drop remaining path customizations

nik9000:
The build wouldn't work properly if we set them to anything else.


nik9000:
Thanks for reviewing @olksdr!

PR #1133, opened 2019-09-03
Drop ability to configure template

nik9000:
Now that we have control over the template file directly we really don't
need to support any kind of fancy configuration for it. This drops that.

See #1115


nik9000:
Thanks for reviewing @olksdr!

PR #1136, opened 2019-09-04
Serve the inter font from the docs

nik9000:
Another step in our ever expanding effort to bring the docs styles into
the docs build.


nik9000:
:fries:

PR #1137, opened 2019-09-04
Drop template munging

nik9000:
Drops the template munging code and makes the changes that it made
directly in the template file. This seems sensible now that we own the
template.

### DIFF
--- a/resources/web/template.html
+++ b/resources/web/template.html
@@ -262,11 +262,11 @@
           <section id="guide" <!-- DOCS LANG -->>
             <div class="container">
               <div class="row">
-                <!-- start body -->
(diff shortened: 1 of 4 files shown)


resources/web/template.html
```diff
@@ -262,107 +257,14 @@
 
     <div class="main-container">
       <section id="content" >
-        <div class="content-wrapper">   
+        <div class="content-wrapper">
 
-          <section id="guide">
+          <section id="guide" <!-- DOCS LANG -->>
```
nik9000: This one is a little funny this is how the perl script knows where to insert things.

nik9000:

PR #1138, opened 2019-09-04
Continue to group styles

nik9000:
Moves the styles for the settings_modal, this_page, and rtpcontainer
into their own files so that we can find things a little easier.

### DIFF
--- a/resources/web/style/settings_modal.pcss
+++ b/resources/web/style/settings_modal.pcss
@@ -1,0 +1,51 @@
+#settings_modal {
+  display: inline-block;
+  position: relative;
+  margin: 15% auto;
+  width: 600px;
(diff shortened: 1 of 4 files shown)


resources/web/style/settings_modal.pcss
```diff
@@ -0,0 +1,51 @@
...
+}
+
+.settings_modal-close {
+  position: absolute;
+  top: 2px;
+  right: 10px;
+  font-size: 20px;
+  color: #ccc;
+  &:hover {
+    color: black;
```
nik9000: This used to be `opacity: black` but it wasn't doing anything. I think we just intended to turn the x black though.

nik9000:

PR #1140, opened 2019-09-04
Remove locale bouncer from template

nik9000:
I don't think it has worked for a while and no one is complaining. We
always ended up with `currentlocale` as `undefined` because `/en` was
missing a trailing `/`.

With that picker gone we can also remove `jquery.cookie` because it was
the last thing to use it.


nik9000:
Thanks @brianjolly!

PR #1141, opened 2019-09-04
Remove Template.pm's "funny" init method

nik9000:
That init method used to be super important but now it is so simple we
can just inline it with the constructor.


nik9000:
Thanks for reviewing @olksdr!

PR #1142, opened 2019-09-04
Drop static scripts being loaded from elastic.co

brianjolly:
This would remove the js linked from elastic.co.


nik9000:
Woah! That'd be great! Let's rebuild all the docs so we can play with the preview:

brianjolly:
Thanks @nik9000, One more step closer! :)

PR #1144, opened 2019-09-05
Slightly clean up raw output

nik9000:
This'd make it simpler for *anything* else to use the raw output by
normalizing docbooks quirks away.

### DIFF
--- a/lib/ES/Template.pm
+++ b/lib/ES/Template.pm
@@ -50,12 +50,6 @@ sub apply {
 
         my $contents = $source->slurp( iomode => '<:encoding(UTF-8)' );
 
-        # Strip XML guff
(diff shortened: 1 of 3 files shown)


lib/ES/Template.pm
```diff
@@ -50,12 +50,6 @@ sub apply {
 
         my $contents = $source->slurp( iomode => '<:encoding(UTF-8)' );
 
-        # Strip XML guff
```
nik9000: I'm slowly, slowly removing everything from this file. Internationally. Making the templating process as simple as possible will allow us much more flexibility about *when* we do it.

nik9000:
Thanks for reviewing @olksdr!

Issue #804, opened 2019-04-11
Should we standardize on using shared version file for stack docs?


dedemorton:
During release day, someone always needs to chase down the PRs that contain doc bumps and verify that they've been merged. We could avoid the hassle by using the shared version file for all relevant projects, but this has some problems, too. I've captured my thoughts for discussion. Please add your thoughts! Let's discuss.

Pros:

- One place to update doc paths used across stack docs. Helps release day go more smoothly.
- Doc release manager is more likely to know how the attributes are used and not introduce problems. For example, every few months, someone in dev forgets that the major version should be n.x, not n.0, and the RPM repo downloads break.

Cons:

- Version info lives in a separate repo where contributors can't find it easily unless they understand how our doc build works.
- Most projects  still need a separate version file to set versions of other software mentioned in the docs (like go-version, python version, etc). Having a pair of eyes on that file on a regular basis is a good thing.
- Some project, like Beats, use the version info to generate docs, so the version will still need to be bumped in those projects. If the doc bump happens outside of the file, people are more likely to forget to bump the version.
- Making one mistake breaks all projects (not sure if this is a pro or con).
- Each time a new branch is cut, someone will need to update the `include` to point to the new version file. This is no more work than bumping, but maybe slightly less obvious and intuitive since the shared version files live in a separate repo. The person creating the branch might not know that there are are files named versions67.asciidoc, versions70.asciidoc, and so on. We complicate this further by having a shared attributes file that uses a different versioning strategy (everyone points to attributes.asciidoc unless something changes and they need older attribute settings).

nik9000:
I'd like to have play with this some. I've had it in the back of my mind for a while but figured I'd wait until after the Asciidoctor migration. I'll think on it some more, and see if I can get it in earlier.

lcawl:
Having just been affected by this again when the creation of the 7.4 branches here are some thoughts:

Our list of books continues to grow. We have a wiki page that lists where the doc version info is for most of those books, we still inevitably miss some, which can cause broken links and possible delays releasing the new content. To me, this is a reason to try to use a shared file for as many as possible. Otherwise, I think we need to do multiple small updates to the conf.yaml file (i.e. don't block 7.4 book X release because 7.4 book Y attributes aren't up-to-date).  Even then, you'd end up with broken links if 7.4 book X links to (non-existent) 7.4 book Y.

lcawl:
> Each time a new branch is cut, someone will need to update the include to point to the new version file.

[...]

PR #1147, opened 2019-09-05
Add source_branch attribute

nik9000:
The `source_branch` attribute should be useful in resolving inter-book
links. We can get this *perfect* when we're running with `--all` because
every book already *has* a branch attribute. When we build with `--doc`
we have to guess. I've added some heuristics that I feal are likely to
guess right most of the time for most folks. I'm not too worried about
guessing wrong *sometimes* because, for the most part, that'll only
break interbook links and lots of folks don't check those locally
anyway. And because we have the pull request tests which run with
`--all` so they'll serve as an accurate backstop.

Relates to #804

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -188,8 +188,32 @@ sub _guess_edit_url {
             $remote = 'unknown';
         }
     }
-    my $branch = eval {run qw(git rev-parse --abbrev-ref HEAD) } || 'master';
+    my $branch = eval { run qw(git rev-parse --abbrev-ref HEAD) } || 'master';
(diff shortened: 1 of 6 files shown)


nik9000:
@olksdr I went "full perl" on this one. The only difficult part of this is the regular expression. And the testing.

nik9000:
Oooh good catch past-me!

nik9000:
Wrote a test for that weird case.

lcawl:
When I remove the source_branch attribute from https://github.com/elastic/stack-docs/pull/493,  the Stack Overview still builds successfully.  Therefore, this LGTM

nik9000:
Thanks for reviewing @olksdr!

PR #1150, opened 2019-09-05
Move alternatives_summary.json

nik9000:
It is an input to the templating process so it should be in the `raw`
directory.


nik9000:
Thanks for reviewing @olksdr!

PR #1153, opened 2019-09-06
Add APM .NET Agent docs

bmorelli25:
Adds the APM .NET Agent `1.0.0` documentation and copies it to `current`. This documentation is built from a [tag](https://github.com/elastic/apm-agent-dotnet/tags) in the repo (not a branch).

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -1070,8 +1070,8 @@ contents:
                   -
                     title:      APM .NET Agent
                     prefix:     dotnet
-                    current:    master
-                    branches:   [ master ]
+                    current:    1.0.0
+                    branches:   [ master, 1.0.0 ]
                     index:      docs/index.asciidoc
                     tags:       APM .NET Agent/Reference
                     subject:    APM


bmorelli25:
There's a problem. The removal of the `beta` tag for the docs didn't happen until after the tagged commit for `1.0.0`. This means the `beta` label still displays over the `1.0.0` documentation.

nik9000:
> There's a problem. The removal of the `beta` tag for the docs didn't happen until after the tagged commit for `1.0.0`. This means the `beta` label still displays over the `1.0.0` documentation.

Tags are a real pain for building docs from. I'd prefer to avoid it if possible. Does the .net apm agent have a 1.x branch we could build instead? Or a 1.0 branch?

bmorelli25:
Unfortunately, no. They only have a `master` branch.

nik9000:
> Unfortunately, no. They only have a `master` branch.

Maybe it is worth them forking a 1.0 branch so they can be compatible with the way the rest of Elastic's repos work. Pretty much all of them fork a branch on release so they can backport bugfixes and docs changes.

bmorelli25:
cc @gregkalapos / @SergeyKleyman

gregkalapos:
So.. my offer from previous discussions still holds: in case having versioned docs is very hard based on tags without branches I'm totally open to think about how we should branch in the .NET Agent repo. 

Downside is that it'll probably take a little bit until I find the time to do that. Ideally this should happen before the `1.1` release (at the same time with the `7.4.0` stack release), but for that we are already packed and probably a little bit behind.

nik9000:
We can use tags as a temporary measure if you are sure they are right. The
trouble is that docs are always "living" in that folks always want to
change them after release. You can move tags but the docs build doesn't
know that.

If the docs as published now are fine I'm fine waiting on this. I bet,
though, that you'll want to bump the version of the docs in the master
branch before you release 1.1 and doing that'd be confusing.

It might be simpler to fork a 1.0 branch off of whatever commit the docs
should come from and revisit it later. It isn't perfect, but it'll make the
release process simpler and you can review it later.

bmorelli25:
Closing this as it looks like they decided to branch in https://github.com/elastic/docs/pull/1184

gregkalapos:
> Closing this as it looks like they decided to branch in #1184

Confirming.. yes, we go with branch.

PR #1154, opened 2019-09-06
Better Parcel watch support in Docker for Mac

JoshMock:
Fixes broken Parcel watcher when running it in a mounted volume in a Docker container using Docker for Mac.

See:

- https://github.com/docker/for-mac/issues/2216
- https://github.com/parcel-bundler/parcel/issues/2539#issuecomment-462946468

### DIFF
--- a/build_docs
+++ b/build_docs
@@ -113,6 +113,15 @@ class OpenBrowserState:
                 # is weird. If we tell it that the browser's name is `open`
                 # we'll get whatever the user set as their default browser.
                 environ['BROWSER'] = 'open'
+
+                # Docker for Mac has trouble watching filesystem events on
+                # a mounted volume, so Parcel's watcher was not triggering
+                # rebuilds when saving files locally that are mapped to the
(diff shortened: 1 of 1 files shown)


build_docs
```diff
@@ -237,6 +246,11 @@ def run_build_docs(args):
             docker_args.extend(['--publish', '8000:8000/tcp'])
             docker_args.extend(['--publish', '8001:8001/tcp'])
             open_browser = OpenBrowserState()
+
+            # ulimit has to get bumped to allow the container to do the
+            # more-expensive polling-style Parcel watcher on Docker for Mac;
+            if environ.get("CHOKIDAR_USEPOLLING"):
```
nik9000: I feel like it'd be a little cleaner to check for mac here and add the environment variable as a `-e` to docker, also right here. *Technically* you only need this if you have both `--doc` and `--open` but it won't hurt anything to add it if you just have `--open`.

JoshMock: good call!

PR #1155, opened 2019-09-06
Replace perl template with js template

nik9000:
This replaces the perl templating code with javascript that does almost
exactly the same thing. The implementations are almost word for word the
same. This is ever so slightly slower because the template has to be
initalized once per book instead of once and because of the fork
overhead of starting a new node process. It has the distinct advantage
of being much simpler to embed in something like the preview service.


template/template.js
```diff
@@ -0,0 +1,128 @@
...
+const readdir = promisify(fs.readdir);
+const readFile = promisify(fs.readFile);
+const stat = promisify(fs.stat);
+const writeFile = promisify(fs.writeFile);
+
+module.exports = async templatePath => {
+  const contents = await readFile(templatePath, {encoding: 'UTF-8'});
+
+  const map = {};
+  const parts = contents.split(/<!-- (DOCS \w+) -->/).map((part, index) => {
```
JoshMock: Just calling out that this would only split on the first instance of this pattern. `/g` needed if you intend to split on all occurrences.

template/template.js
```diff
@@ -0,0 +1,128 @@
...
+    const matcher = /DOCS (\w+)/.exec(part);
+    if (matcher) {
+      part = '';
```
JoshMock: While legal, treating arguments as mutable can be an unsafe practice; makes it easy for future maintainers to accidentally assume an arg is constant when it isn't.

template/template.js
```diff
@@ -0,0 +1,128 @@
...
+      map[matcher[1]] = index;
+    }
+    return part;
+  });
+  const apply = (raw, lang, initialJsState) => {
+    const head = /<head>(.+?)<\/head>/s.exec(raw);
+    if (!head) {
+      throw `Couldn't find head in ${raw}`;
```
JoshMock: While JS is (again) extremely permissive, it's generally recommended to throw something that extends `Exception` or just `throw new Exception("message")` for consistency.

template/template.js
```diff
@@ -0,0 +1,128 @@
...
+      return "{}";
+    }
+    const readAltSummary = JSON.parse(await readFile(alternativesSummary));
+    return JSON.stringify(module.exports.buildInitialJsState(readAltSummary));
+  };
+  const applyToDir = async (sourcePath, destPath, lang, alternativesSummary, tocMode) => {
+    const initialJsState = await buildInitialJsStateFromFile(alternativesSummary);
+    const entries = await readdir(sourcePath);
+    await mkdir(destPath, {recursive: true});
+    for (var e = 0; e < entries.length; e++) {
```
JoshMock: nit: [`Array.prototype.forEach`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) would work here.

template/template.js
nik9000: It seems to fully split the string.

template/template.js
nik9000: Sure! I copied this exactly from the perl which super doesn't care what you do with parameters. I'll change it.

template/template.js
nik9000: Got it.

template/template.js
nik9000: How does that work with async/await? I think it was upset at me when I tried it.

nik9000:
@JoshMock, could you have another look at this when you get a chance?

nik9000:
> @nik9000 Josh is out, I took another look, hamburger !

:fries:

PR #1156, opened 2019-09-06
Drop styles that leak into the header and footer

nik9000:
These are managed by the header and footer itself. For the most part
this doesn't actually change anything visually because the header and
footer *already* manage it.


nik9000:
@brianjolly, could you have a look at this one?

nik9000:

PR #1157, opened 2019-09-06
Move table styles into their own file

nik9000:
Moves the styles for tables into their own file so it is a bit easier to
manage. Drops the stype for paragraphs inside tables because it isn't
being used.


nik9000:
Thanks for reviewing @JoshMock!

PR #1158, opened 2019-09-06
Style: Move console widget sytles into file

nik9000:
Moves the console widget styles into their own file so they are a little
easier to keep track of. Pulls the gear icon into its own file and lets
parcel inline it so we can look at the file rather than a blob of base64
stuff.


nik9000:
indeed

PR #1159, opened 2019-09-06
Drop template fetch script

nik9000:
We don't fetch the template from anywhere anymore. We maintain it in the
docs repo.


nik9000:
@brianjolly, what do you think of this?

nik9000:
Thanks for reviewing @brianjolly !

PR #1163, opened 2019-09-06
Drop m-shortcut styles

nik9000:
I *think* these aren't used any more. I can't find them, at least.


nik9000:
Here is another one for you @brianjolly. I haven't seen these icons in as long as I can remember.

nik9000:
Amazing.

PR #1164, opened 2019-09-09
Add Elasticsearch repo to Stack Glossary for content reuse

jrodewig:
Updates the Stack Glossary build to include the Elasticsearch repo as a dependency.

This allows the Elasticsearch docs team to reuse definitions from the [Elasticsearch Reference Guide's glossary](https://www.elastic.co/guide/en/elasticsearch/reference/master/glossary.html).

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -113,6 +113,9 @@ contents:
               -
                 repo:   stack-docs
                 path:   docs/en
+              -
+                repo:   elasticsearch
+                path:   docs
           -
             title:      Stack Overview
             prefix:     en/elastic-stack-overview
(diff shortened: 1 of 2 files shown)


jrodewig:
elastic/stack-docs#500 would take advantage of this PR once merged.

@lcawl  @nik9000 Let me know your overall thoughts on this approach. I'm in favor of content reuse, but I see the tradeoffs with adding dependencies, especially if other books use this same approach.

conf.yaml
```diff
@@ -113,6 +113,9 @@ contents:
               -
                 repo:   stack-docs
                 path:   docs/en
+              -
+                repo:   elasticsearch
+                path:   docs
```
nik9000: Do you think you'd be ok with `docs/reference` or something like that? Making these as precise as possible will cut down on noop builds.

jrodewig: Makes sense. Done with a04d63a.

nik9000:
@jrodewig the PR checks will be useful for this PR, paricularly the elasticsearch-ci/docs one. It will make sure that it can resolve the new path.

PR #1167, opened 2019-09-09
We have the autoprefixer

nik9000:
Enabling pcss enabled the autoprefixer plugin so we really don't need to
be defining any of the funny prefixed styles. This drops all of them.

### DIFF
--- a/resources/web/style/conum.pcss
+++ b/resources/web/style/conum.pcss
@@ -6,7 +6,6 @@
     line-height: 16px;
     color: #f5f7fa;
     background-color: #0080d5;
-    -webkit-border-radius: 100px;
     border-radius: 100px;
     text-align: center;
     font-size: 10pt;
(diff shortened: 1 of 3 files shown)


resources/web/style/link.pcss
```diff
@@ -2,12 +2,10 @@
   a {
     color: #00a9e5;
     font-weight: normal;
-    -webkit-text-decoration: none;
     text-decoration: none;
     outline: none;
     outline-offset: 0;
     &:hover, &:focus {
-      -webkit-text-decoration: underline !important;
```
nik9000: Autoprefixer doesn't handle these at all because they haven't been needed in a long time.

resources/web/style/conum.pcss
```diff
@@ -6,7 +6,6 @@
     line-height: 16px;
     color: #f5f7fa;
     background-color: #0080d5;
-    -webkit-border-radius: 100px;
```
nik9000: Autoprefixer doesn't handle this one because it hasn't been needed in a long time.

PR #1168, opened 2019-09-09
Screenshot tests, examples, and style

nik9000:
* Adds an example of a screenshot to README.asciidoc
* Adds a test that it renders how we like it
* Moves `cat.jpg` to a more appropriate spot
* Updates a lot of tests after moving `cat.jpg`
* Moves the styles for the screenshot into their own file


resources/web/styles.pcss
```diff
@@ -198,19 +199,6 @@ p.remark {
     clear: right;
 }
 
-/* Images */
```
nik9000: This is what I started out doing. But there wasn't an example of a screenshot in the readme and on thing led to another and now we have this.

PR #1171, opened 2019-09-09
Upgrade some dependencies

nik9000:
Upgrades us from a rather old version of debian and all that brings
along which includes at least ruby 2.5. Upgrades nodejs to 12.x which'll
be nicer to work with on the server side. Switches to installing bundler
with apt to save a little time.


nik9000:
@haywoood, could you have a look at this one? I started working on something and realized I should work on some other thing first. Then I started work on that and I realized I'd really like node 12.x to do it in. So I started work on this.

PR #1173, opened 2019-09-09
Warn if empty diff

nik9000:
Empty diffs aren't as common as we'd hoped! This changes the diff page
generation to add a line explaining that the diff is empty if it is
indeed empty. Without this the diff page just looks like a totally blank
page which is confusing.


nik9000:
:fries:

PR #1174, opened 2019-09-10
Allow selecting headings

### LINKED ISSUE #277: Allow selection of headlines
Today it's not possible to select headlines with your mouse so that you can copy it. This is important for the APM Node.js agent as the headlines in many cases are command that the user needs to type, so not being able to select and copy them is really annoying.

Here's an example of one such headline:

https://www.elastic.co/guide/en/apm/agent/nodejs/current/agent-api.html#error-on-aborted-requests

nik9000:
Previously our headings were "covered up" by a funny anchor tag that we
used to make it possible to link to portions of the page. This made them
impossible to select. This changes headings so they are *part of* the
anchor tag. Text inside an anchor tag isn't selectable on its own but
you *can* select it by dragging from outside of the tag inwards. This is
"how browsers work" so folks should be used to this. Hopefully making
the text part of the anchor tag will make it much more obvious what is
going on.

Closes #277

### DIFF
--- a/integtest/spec/helper/dsl/file_contexts.rb
+++ b/integtest/spec/helper/dsl/file_contexts.rb
@@ -62,7 +62,9 @@ module Dsl
       let(:title) do
         return unless body
 
-        m = body.match %r{<h1 class="title"><a id=".+"></a>([^<]+)(<a.+?)?</h1>}
+        m = body.match(
+          %r{<h1 class="title"><a id="[^\"]+">([^<]+)</a>(<a.+?)?</h1>}
+        )
         raise "Can't find title in #{body}" unless m
 
         m[1]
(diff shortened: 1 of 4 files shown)


nik9000:
@elasticmachine run elasticsearch-ci/docs rebuild. We have to pick up the xslt change in the preview.

nik9000:
@tylerjl this is one of those things where we'd *really* like to have cache busting. As we stand now merging this will "look funny" for a day because we're changing the shape of the html. This one is almost so bad I think we should wait on merging it until we have cache busting.

nik9000:
Woops. This needs little more work.

nik9000:
Ok. This is getting super complex, sadly. docbook really gets in the way here. This might be something that we should wait until after we've dumped docbook before we can have it.

tylerjl:
> this is one of those things where we'd really like to have cache busting. As we stand now merging this will "look funny" for a day because we're changing the shape of the html. This one is almost so bad I think we should wait on merging it until we have cache busting.

Thank you for raising the topic @nik9000 - agreed that this is something that, logically, we should be able to perform.

I'm going to get a conversation going somewhere so we can track this.

PR #1175, opened 2019-09-10
Streaming templates

nik9000:
Replaces the template implementation copied from perl with a work-alike
that uses async iterators and async generators. It has *slightly* higher
overhead but streams the template and raw html rather than pulling it
all into memory which protects it against very large html which should
make it embeddable into the preview application so we can template on
the fly.

Note: to get this I had to update Node to 12.x to support turning an
async generator into a readable stream. This is important because these
streams properly handle backpressure. I don't really expect any
backpressure when streaming to local files but I *do* expect it in the
preview app.


nik9000:

nik9000:
Neat! The last test failure totally found something!

PR #1176, opened 2019-09-10
Hide nav until styles are loaded

brianjolly:
This hides the nav until JS and styles are loaded to prevent the flash of unstyled content.

### DIFF
--- a/resources/web/template.html
+++ b/resources/web/template.html
@@ -132,7 +132,8 @@
 
     <!-- Header Section -->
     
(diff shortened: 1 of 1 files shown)


nik9000:
@brianjolly we'll have to wait on the docs build to quiet down before we can get a preview.

nik9000:
@brianjolly, now that the docs build is calmer I think it is worth kicking off a preview build of this. If you merge master into the PR branch it should build some books. If not we can force it to rebuild them all.

PR #1181, opened 2019-09-11
Update icons for Docs RTP, sync rtp styles

brianjolly:
This adds new icons for docs RTP and syncs the #rtpcontainer styles in docs with the styles from elastic.co's RTP styles.


resources/web/style/rtpcontainer.pcss
```diff
@@ -4,95 +4,179 @@
...
-    }
-  }
-
-  .mktg-promo {
-    border: 1px solid #ccc;
-    border-radius: 5px;
-    margin-bottom: 40px;
-    padding: 20px;
-  }
+	.mktg-promo {
```
nik9000: Could you change the change the indentation to be two spaces like the old file?

brianjolly:
Hey @nik9000, it looks like the link checker doesn't like this style:
`background-image: url(/static-res/images/logo-elasticsearch-32-white.svg);`

It would resolve to this URL: https://www.elastic.co/static-res/images/logo-elasticsearch-32-white.svg

Is there a way to set the style in a way that the link checker will like?

Do we need to set an absolute URL, or is there a better way?
`background-image: url(https://www.elastic.co/static-res/images/logo-elasticsearch-32-white.svg);`

nik9000:
> Is there a way to set the style in a way that the link checker will like?

I'd put the logo into the docs repo and link to it like the other images in the styles. That'll be important for it to work offline anyway.

resources/web/style/rtpcontainer.pcss
```diff
@@ -4,95 +4,179 @@
...
-            width: 0.7em;
-            background-repeat: no-repeat;
-            background-size: 70%;
-            background-position: center center;
-          }
-        }
-
-        /* Product icons */
-        &.icon-elasticsearch-white:before {
-          background-image: inline("img/rtpcontainer/icon-elasticsearch-bb-white.svg");
```
nik9000: If we're not using any of these any more I think we should remove them from the docs repo.

brianjolly: These background images will be used sometimes by rtp sometimes. They won't be used every time, but gives the people creating the rtp some options.

brianjolly:
Good point @nik9000 , I added a little right margin to give the icons some space.

PR #1184, opened 2019-09-12
APM .NET Agent: Add 1.x to branches list and set current to 1.x

SergeyKleyman:


bmorelli25:
Using this approach instead of the one outlined in https://github.com/elastic/docs/pull/1153.

PR #1189, opened 2019-09-14
Break out styles

JoshMock:
Quick and easy. 🧹


resources/web/style/comments.pcss
```diff
@@ -0,0 +1,10 @@
+p.remark {
```
nik9000: I don't *think* we actually use these:

```
$ pwd
/home/manybubbles/Downloads/built-docs/html
$ find . -name '*.html' -exec grep --color '[ "]remark[ "]' {} \+
<nothing>
```

resources/web/style/guide_base_styles.pcss
```diff
@@ -0,0 +1,25 @@
+#guide {
```
nik9000: maybe just `base`? everything in the docs is "guide".

resources/web/style/guide_base_styles.pcss
```diff
@@ -0,0 +1,25 @@
...
+    code {
+      font-size: 16px;
+    }
+  }
+
+  dd {
+    margin: 0 0 0.5em 2em;
+  }
+
+  ul.lists li {
```
nik9000: I think we have `list.pcss` for this one. It'd probably be worth tracking down if we actually use it though.

resources/web/style/js_api.pcss
```diff
@@ -0,0 +1,33 @@
+/* JS API TOC */
```
nik9000: I don't think we use any of these:
```
$ find . -name '*.html' -exec grep --color 'js-api-method' {} \+
<nothing>
$ find . -name '*.html' -exec grep --color 'js-client-docs' {} \+
<nothing>
````

resources/web/style/comments.pcss
JoshMock: removed

resources/web/style/guide_base_styles.pcss
JoshMock: renamed!

resources/web/style/guide_base_styles.pcss
JoshMock: I scanned and couldn't find any references, so it's dropped now.

resources/web/style/js_api.pcss
JoshMock: Removed this whole file!

PR #1190, opened 2019-09-17
Add Noto Sans Japanese font to support air-gapped docs

JoshMock:
As part of the work of to air-gap the docs, and following @nik9000's lead in #1136.


integtest/spec/all_books_spec.rb
```diff
@@ -55,8 +55,10 @@ def self.has_license(name, heading)
     end
     file_context 'html/static/styles.css' do
       has_license 'Inter', 'SIL OPEN FONT LICENSE'
+      has_license 'Noto Sans Japanese', 'SIL OPEN FONT LICENSE'
     end
     file_context 'html/static/Inter-Medium.5d08e0ba.woff2'
+    file_context 'html/static/NotoSansJP-Black.woff2'
```
JoshMock: I wasn't certain what the purpose of running `file_context` without a block was, so I just added this to, presumably, verify that some Noto Sans file is present at build time.

nik9000: It just checks if the file exists.

nik9000:
> LGTM

Except the tests failed!

JoshMock:
Cache-busting filenames: very helpful for making production easy to cache, not very helpful for guessing filenames in integration tests against built artifacts.

PR #1191, opened 2019-09-17
Template on the fly in preview

nik9000:
This makes the preview service template html pages on the fly.

A few things had to happen to make this go:
1. Start writing everything that we need to render the tempelate on the
fly into built-docs, specifically the template and the language.
2. Only use the template if it is in the built-docs branch we're trying
to render. Branches without the template serve from the "html" directory
like we do in production.
3. Rework error handling for the git operations.
4. Move the git operations into their own file and unit test them.
5. Rework some other error handling after I found out the `.pipe`
doesn't propagate errors in node.

The performance cost of apply the template on the fly is something like
20 milliseconds which doesn't feel to bad because we're running six git
operations instead of two.

### DIFF
--- a/integtest/spec/helper/book.rb
+++ b/integtest/spec/helper/book.rb
@@ -24,6 +24,10 @@ class Book
   # Should this book allow overriding :edit_url:? Defaults to false.
   attr_accessor :respect_edit_url_overrides
 
+  ##
+  # The language of the book. Defaults to `en`.
+  attr_accessor :lang
+
   def initialize(title, prefix)
     @title = title
     @prefix = prefix
@@ -33,6 +37,7 @@ class Book
     @branches = ['master']
     @current_branch = 'master'
     @respect_edit_url_overrides = false
+    @lang = 'en'
   end
 
   ##
@@ -81,6 +86,7 @@ class Book
       tags:       test tag
       subject:    Test
       asciidoctor: #{@asciidoctor}
+      lang:       #{@lang}
     YAML
   end
 
--- a/preview/git.js
+++ b/preview/git.js
@@ -1,0 +1,252 @@
+/**
+ * @license
+ * Licensed to Elasticsearch under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Elasticsearch licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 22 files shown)


nik9000:
Hello integration tests

my old friend

preview/git.js
```diff
@@ -0,0 +1,252 @@
...
+
+const parseDiffTreeZ = async function* (itr) {
+  const loadFirstChunk = await itr.next();
+  if (loadFirstChunk.done) {
+    // Empty diff!
+    return;
+  }
+  let chunk = loadFirstChunk.value;
+  const sliceOffNul = async from => {
+    while (true) {
```
JoshMock: Did the option of using a [Duplex stream](https://nodejs.org/api/stream.html#stream_class_stream_duplex) come up instead of an infinite loop? It does add a little complexity to a simpler use case like this, but it'd be taking better advantage of a stdlib tool that's designed for handling this exact kind of need.

nik9000: I had a Transform stream implementation of this and it required a whole bunch of accounting that was difficult to get right. To the point where I'm pretty sure I didn't. The async-generators feel a little less tricky to write. Though how they execute is a bit of a trip.

JoshMock: Fair point. As I was reading I wondered if it was something like that. They're awesome when you get them dialed in just right, but it takes some time to do the dialing-in well. :+1:

preview/test.sh
```diff
@@ -8,11 +8,16 @@
 set -e
 
 cd $(git rev-parse --show-toplevel)
+../infra/ansible/roles/git_fetch_reference/files/git-fetch-reference.sh built-docs.git
```
nik9000: These changes *vastly* speed up testing outside of unit and integration tests.

integtest/spec/helper/book.rb
```diff
@@ -24,6 +24,10 @@ class Book
   # Should this book allow overriding :edit_url:? Defaults to false.
   attr_accessor :respect_edit_url_overrides
 
+  ##
+  # The language of the book. Defaults to `en`.
```
estolfo: There's no functionality to set/override the `@lang`, is that intentional?
This comment says the lang default is 'en' but you can't set it otherwise...

nik9000: Ah! I certainly meant to....

nik9000: I set it with `book.lang` [here](https://github.com/elastic/docs/pull/1191/files#diff-88db19d5cf245e6c1d6566dc005be3abR427).  It seems to work...

PR #1192, opened 2019-09-18
Adding non stack versions to the docs repo

bmorelli25:
Will it work 🤷‍♂. Mostly for discussion purposes.


bmorelli25:
This would work but is way too complicated.

PR #1194, opened 2019-09-18
Drop some meta

nik9000:
Drops some meta tags from the header. One is hard coded which seems
wrong. The other one isn't all that helpful.


nik9000:
@brianjolly, do you know if these meta tags *do* anything? They don't feel super helpful.

nik9000:
:beer:

PR #1195, opened 2019-09-18
Drop comments and empty meta from template

nik9000:
Drops two misleading comments from the template and an empty meta tag.


nik9000:

PR #1198, opened 2019-09-19
Create "current" stack version file

nik9000:
This file will be useful for books that *generally* want to point the
current release of the stack. We'll update it when we release a new
version of the stack.

### DIFF
--- a/shared/versions/stack/README
+++ b/shared/versions/stack/README
@@ -11,14 +11,10 @@ branch.
 
 NOTE: `current.asciidoc` is never automatically used.
 
-Books outside of the stack have three choices:
+Books outside of the stack have two choices:
 
-1. Don't use these versions files at all.
-2. Use the `current.asciidoc` file.
-3. "Pin" theselves to some version of the stack.
-
-Not using these files is perfectly appopriate for books that don't link to the
-stack.
+1. Use the `current.asciidoc` file.
+2. "Pin" theselves to some version of the stack.
 
 Using `current.asciidoc` is useful because we'll update on whenever we release
 a new version of the stack so the doc will automatically update on release.
--- a/shared/versions/stack/README
+++ b/shared/versions/stack/README
@@ -24,6 +24,6 @@ Using `current.asciidoc` is useful because we'll update on whenever we release
 a new version of the stack so the doc will automatically update on release.
 
 "Pinning" the version of the stack that a book refers to is quite useful for
-old versions of books for which consistently see broken links when we release
+old versions of books for which we consistently see broken links when we release
 new versions of the docs. It could also be useful for books that need special
 care when bumping the stack version.
--- a/shared/versions/stack/README
+++ b/shared/versions/stack/README
@@ -16,7 +16,7 @@ Books outside of the stack have two choices:
 1. Use the `current.asciidoc` file.
 2. "Pin" theselves to some version of the stack.
 
-Using `current.asciidoc` is useful because we'll update on whenever we release
+Using `current.asciidoc` is useful because we'll update it whenever we release
 a new version of the stack so the doc will automatically update on release.
 
 "Pinning" the version of the stack that a book refers to is quite useful for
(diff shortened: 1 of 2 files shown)


nik9000:
@bmorelli25 I've tried to summarize why I think we want this file in the README that I created. 

@lcawl, we talked a while back about maybe wanting this file. It turns out that @bmorelli25 wants it for APM. If we merge this we really should update the docs release instructions. Could you have a look at this and approve or reject it?

shared/versions/stack/README
```diff
@@ -0,0 +1,29 @@
...
+
+Which will resolve to the `6.3.asciidoc` file for he 6.3 branch. Or the
+`7.2.asciidoc` file for the 7.2 branch. Or `master.asciidoc` for the master
+branch.
+
+NOTE: `current.asciidoc` is never automatically used.
+
+Books outside of the stack have three choices:
+
+1. Don't use these versions files at all.
```
bmorelli25: Hmmm, I don't think this is really an option any more. With the shared version change, if I want to use any of the shared attributes in the APM Agents, they also **have** to use the shared version files. An example of why is in the recent ECS PR: https://github.com/elastic/docs/pull/1170.

bmorelli25: So, ok, _technically_ it is an option still. But I would have to start setting attribute values in each of the Agent's `index.asciidoc` files.

nik9000: Maybe I should just drop it from the list then.

bmorelli25: I think so.

shared/versions/stack/README
```diff
@@ -0,0 +1,29 @@
...
+2. Use the `current.asciidoc` file.
+3. "Pin" theselves to some version of the stack.
+
+Not using these files is perfectly appopriate for books that don't link to the
+stack.
+
+Using `current.asciidoc` is useful because we'll update on whenever we release
```
lcawl:
```suggestion
Using `current.asciidoc` is useful because we'll update it whenever we release
```

shared/versions/stack/README
```diff
@@ -0,0 +1,29 @@
...
+a new version of the stack so the doc will automatically update on release.
+
+"Pinning" the version of the stack that a book refers to is quite useful for
+old versions of books for which consistently see broken links when we release
```
lcawl:
```suggestion
old versions of books for which we consistently see broken links when we release
```

lcawl:
I don't have a problem with this "current.asciidoc" solution. I think it will still incur some of the usual problems where updating that file will result in broken links, since we invariably have pages that existing in Version X but don't exist anymore in Version X+1.  If we can freeze the docs and stage that change to current.asciidoc, it buys us a bit more time to clean up links like that.

nik9000:
> If we can freeze the docs and stage that change to current.asciidoc, it buys us a bit more time to clean up links like that.

We can delay bumping current.asciidoc if we have to. That'd mean that things outside the stack don't point to the most current version of the stack for a little while. That seems "ok" while we're resolving broken links.

lcawl:
> We can delay bumping current.asciidoc if we have to. That'd mean that things outside the stack don't point to the most current version of the stack for a little while. That seems "ok" while we're resolving broken links.

That sounds reasonable to me.  I think we'll also be able to see the full list of broken links in the PR that bumps current.asciidoc, as long as we remember to do a full rebuild test there.

charith-elastic:
I was also interested in this solution to keep ECK code samples pointing to the current _released_ version of the stack. However, `current.asciidoc` is pointing to 7.4 -- which is yet to be released. Is that because the release is imminent?

nik9000:
> I was also interested in this solution to keep ECK code samples pointing to the current _released_ version of the stack. However, `current.asciidoc` is pointing to 7.4 -- which is yet to be released. Is that because the release is imminent?

I think I just made a mistake there.

PR #1200, opened 2019-09-19
Fix serving SVG images

### LINKED ISSUE #896: Make .png images work
`.png` images in our documentation do not render. It'd be nice if they did. I tested this with a fresh install of asciidoctor and they work just fine. This problem might have something to do with docbook.

nik9000:
We have to set the right mime type.

Closes #896


lib/ES/Util.pm
```diff
@@ -794,11 +798,6 @@ http {
         add_header 'Access-Control-Allow-Headers' 'kbn-xsrf-token';
       }
     }
-    types {
-      text/html  html;
-      application/javascript  js;
-      text/css   css;
-    }
```
nik9000: These didn't do anything because nginx expects the proxied application to set the mime type.

PR #1201, opened 2019-09-20
Switch the alternatives report to json

### LINKED ISSUE #1135: Emit client docs examples reference as JSON/XML
The reference file for `Console` examples that can be implemented as client examples is currently an asciidoc file. Whilst parsing the format of this file is not too tricky, it would likely be less error prone to use more structured format like JSON or XML, which would allow clients to consume it more easily.

nik9000:
So it'll be easier to parse.

Closes #1135


nik9000:

nik9000:
Thanks for reviewing @estolfo ! @russcam, are you ok with this? It moves the file but not *too* far.

nik9000:

PR #1202, opened 2019-09-20
Fix up tests for git

nik9000:
The tests relied on `HEAD` which isn't really *there* for intake builds.


nik9000:
Right now the intake builds are failing due to differences in the git layout between the PR builds and local build and intake build. This should fix it so I'm going to self-merge this once the PR tests pass.

nik9000:
> once

if

PR #1205, opened 2019-09-23
Serve air gapped template from preview

nik9000:
This allows you to request the "air gapped" template from the preview
application by prefixing the branch name with `gapped_`. So you'd do
`http://gapped_master.docs-preview.app.elstc.co/guide` to see the master
branch with the air gapped template. It won't look good at all because
I've made no real effort to make it look right yet. That'll come in a
bit. For now, just seeing it is nice.

This also adds an air gapped template that drops some lines that
obviously aren't going to work properly when the docs are air gapped.
I've kept things like the favicon around because I believe we *should*
be able to make those work.

If you request the air gapped template from preview for a branch that
doesn't have a copy of the air gapped template then the request will
just 404.

Finally, I removed a few comments from the template because they don't
look like they really belong. I saw then when I was building the air
gapped template.

### DIFF
--- a/integtest/spec/preview_spec.rb
+++ b/integtest/spec/preview_spec.rb
@@ -41,30 +41,29 @@ RSpec.describe 'previewing built docs', order: :defined do
   let(:preview) { @preview }
   let(:logs) { preview.logs }
 
+  # The preview server reads the branch from the `Host` header. It throws out
+  # everything after and including the first `.` so you can hit a branch
+  # at urls like `http://master.docs-preview.app.elstc.co/`. Also, if it
+  # can't find the `.` then it just assumes master.
+  let(:host) { "#{branch}.localhost" }
+
   def wait_for_logs(regexp, timeout: 10)
     preview.wait_for_logs(regexp, timeout)
   rescue Timeout::Error
     expect(preview.logs).to match(regexp)
   end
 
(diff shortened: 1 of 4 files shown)


resources/web/template.html
```diff
@@ -54,7 +50,6 @@
     <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start': new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0], j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src= '//www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f); })(window,document,'script','dataLayer','GTM-58RLH5');</script>
     <!-- End Google Tag Manager -->
     
-    <!-- new ga for staging server -->
```
nik9000: @brianjolly Does it make sense to remove these comments? Are they accurate?

nik9000:
> @JoshMock I'm doing this in baby steps!

Failing tests! Also, I didn't add a test. Oops!

nik9000:
That should do it!

nik9000:
Thanks for reviewing @JoshMock! I'd like to do a couple more things before I deploy this....

PR #1208, opened 2019-09-24
Apply template on the fly with --doc

nik9000:
This applies the template on the fly when building docs locally. This is
useful for quickly iterating on the template. It also allows you to get
the air gapped template for the doc by requesting
`gapped.localhost:8000/guide`.

I implemented this by extracting the git logic for the preview service
into a "git core" and adding a new "fs core" which can serve documents
from the local filesystem. I like this particular implementation because
it gives us fewer things to test because almost everything is going
through the preview application.

It doesn't apply the template on the fly to `--all --open` because
*that* is supposed to be a more "production like" test. And, at least
for now, production uses nginx to serve static files locally.

### DIFF
--- a/build_docs
+++ b/build_docs
@@ -84,6 +84,7 @@ class OpenBrowserState:
     def __init__(self):
         self.need_nginx = True
         self.need_parcel = False
+        self.need_preview = False
         self.opened = False
 
     def building_single_book(self):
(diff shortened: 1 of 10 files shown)


nik9000:
Yeah.... We have tests. But that isn't great.

nik9000:
> Yeah.... We have tests. But that isn't great.

Actually, we really don't have tests for this particular thing either!

PR #1213, opened 2019-09-24
Add bootstrap css to our CSS

nik9000:
Adds bootstrap's css to our CSS which:
1. Prevents some of the "jumping" that happens on page load.
2. Allows the air-gapped template to work *much* better.

Follow-up work includes:
1. Step sending bootstrap's css to the docs as part of the header.
2. Figure out exactly what part of bootstrap we actually use.

### DIFF
--- a/integtest/spec/all_books_spec.rb
+++ b/integtest/spec/all_books_spec.rb
@@ -54,6 +54,7 @@ RSpec.describe 'building all books' do
       has_license 'symbol-observable', 'The MIT License (MIT)'
     end
     file_context 'html/static/styles.css' do
+      has_license 'Bootstrap', 'The MIT License (MIT)'
       has_license 'Inter', 'SIL OPEN FONT LICENSE'
       has_license 'Noto Sans Japanese', 'SIL OPEN FONT LICENSE'
     end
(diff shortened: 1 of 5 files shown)


nik9000:
Wow that is huge:
```
-rw-rw-r--. 1 manybubbles manybubbles 169K Sep 24 20:08 styles_new.css
-rw-rw-r--. 1 manybubbles manybubbles  50K Sep 19 16:44 styles_old.css
```

nik9000:
> Wow that is huge:

So maybe we don't merge this and instead figure out what bits we need. Or maybe we continue to include it as a separate file. It is just so big and so unlikely to change that bundling it with the other CSS doesn't make too much sense.

nik9000:
So I think it'd be *fairly* reasonable to merge this if we promise ourselves we'll stop sending the file to the docs right after and then we reduce the styles to just what we need. The trouble with iterating on those styles before merging this is that the public site will continue to send us bootstrap.css until we shut that off. We could try to do both changes at the same time but such things always run into problems.

nik9000:
Ok. Let's get this in an we can refactor it back out once we stop the public site from sending it to us.

PR #1216, opened 2019-09-25
Limit git gc's memory usage

nik9000:
This limits `git gc`'s memory usage. I chose a single thread because all
of the memory is per thread. I chose a small window size because I
[heard][so] that it isn't particularly useful. Even if it were, it'd make
the compressed result smaller which isn't currently super important to
us. I went with 500mb as the limit for the window because this seemed
more useful.

[so]: https://stackoverflow.com/a/27839473


nik9000:

PR #1217, opened 2019-09-25
Fix toc in preview

nik9000:
Fix the table of contents in the preview and `--doc --open` modes.
Without this the table of contents just *doesn't* show up. With this we
continue to perform client side merging of the toc into the page.

Now that we're templating on the fly it is tempting to merge the toc at
template time, but that won't work in production because we're not
templating on the fly in production. Yet.


nik9000:
It looks like this'll require another release of the preview app.  Let me bump the version as part of this PR.

nik9000:

nik9000:
Thanks for reviewing @JoshMock!

PR #1218, opened 2019-09-25
Template: remove slick.js

nik9000:
We aren't using it.


nik9000:
@brianjolly I think we talked this one over a while ago but I never actually removed the reference to the js file.

nik9000:
:zap:

PR #1219, opened 2019-09-25
Drop ie8 and social stuff from air gapped

nik9000:
Drops our "best effort" for IE8 from the air gapped network template
because it requires access to the external web. We *could* bundle it but
it doesn't seem worth it for IE8.

Drops our `og:image` because that is used by facebook when linking to
the page. The air gapped docs aren't going to be linked to facebook
anyway. And that link is the last reference to www.elastic.co on the
template.


nik9000:

PR #1220, opened 2019-09-25
Adds transform redirects

lcawl:
This PR adds redirects for transform content that was moved from the Stack Overview to the Elasticsearch Guide.  It also adds redirects for machine learning content that was moved from the Kibana Guide to the Stack Overview.

These redirects enable us to remove the matching "deleted pages" from the end of the books.


lcawl:

PR #1221, opened 2019-09-25
Drop bootstrap.js

nik9000:
It isn't used.


nik9000:
> It isn't used.

This statement is more hopeful than sure. I'm using the preview that this generates to check.

nik9000:

PR #1223, opened 2019-09-25
Air gapped docs in a docker image

nik9000:
Adds scripts to build a docker image with the air gapped docs. This
required a ton of changes that came before this, but this commits adds
the following features:

1. Adds a `--default-template` option to `preview/cli.js` which is used
when you don't specify a branch or the `gapped_` prefix.
2. Adds the `--gapped` option to the perl code handling previews which,
when specified, skips the initial fetch of the built docs, specifies the
`--default-template` to the preview, and skips the periodic fetches.
3. Adds a Dockerfile and scripts to build the docker image.

This project still isn't done because:
1. We need to add the infrastructure to publish the docker image.
2. We need to remove all external dependencies from the air gapped
template.

Probably more stuff I haven't thought of yet.

### DIFF
--- a/.ci/publish_air_gapped_docs.sh
+++ b/.ci/publish_air_gapped_docs.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+set -eo pipefail
+
+./air_gapped/build.sh
+# TODO publish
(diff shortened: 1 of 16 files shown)


nik9000:
Is a special thing you can do if you have more than one dockeriginore file.
I don't want to suck up the git clone when building the other imeages.

PR #1224, opened 2019-09-26
Include jquery in built-docs

nik9000:
Switches the template from serving the jquery from the public site to
serving it from the `/guide/static` which is served from the
`built-docs` repo. Also copies jquery.min.js into the `built-docs` repo.

I tried bundling `jquery` into `docs.js` but it made it much, much
bigger. That didn't seem worth it.


nik9000:
@JoshMock I think it'd be nice to drop our dependency on jquery altogether but that is a problem for another day, I think.

PR #1227, opened 2019-09-27
Include remaining static resources in preview

nik9000:
Includes the last static resources used by the air_gapped_template. Now
air gapped docs shouldn't redirect to production at all.


integtest/spec/helper/dsl.rb
```diff
@@ -41,6 +43,19 @@ def dest_file(file)
     end
   end
 
+  RSpec.shared_examples 'the favicon' do
+    context 'the fivicon' do
```
JoshMock: `s/fivicon/favicon`

resources/web/__test__/air_gapped_template.test.js
```diff
@@ -0,0 +1,64 @@
...
+  beforeAll(async () => {
+    template = await readFile("air_gapped_template.html", {encoding: "utf8"});
+  });
+  describe("links", () => {
+    /**
+     * Outgoing links from the template.
+     */
+    let links;
+    beforeAll(() => {
+      links = Array.from(new Set([...template.matchAll(/(?:href|src)="([^"]+)"/g)].map(m => m[1]))).sort();
```
JoshMock: What's the value of `...` spreading into a fresh array? Does `template.matchAll` return a pseudo-array?

resources/web/__test__/air_gapped_template.test.js
```diff
@@ -0,0 +1,64 @@
...
+     * Resources that are statically served from the preview app.
+     */
+    let staticResources;
+    beforeAll(async () => {
+      staticResources = (await readdir("static")).map(e => `/${e}`);
+    });
+
+    it("to expected resources", () => {
+      const allResources = ["/guide/", ...compiledWebResources, ...staticResources].sort();
+      expect(links).toEqual(allResources);
```
JoshMock: Does `toEqual` do a deep equality check here?

resources/web/__test__/air_gapped_template.test.js
nik9000: An iterator. Let me check if that works actually. I had the spread before the set.

resources/web/__test__/air_gapped_template.test.js
nik9000: It does, yeah.

resources/web/__test__/air_gapped_template.test.js
nik9000: I played with it a bit and the `[...]` is required because you can't call `map` on the iterator that `matchAll` returns.

PR #1228, opened 2019-09-30
Update 'current' to '3.x' for APM Node.js Agent

watson:


watson:
Merging without a review - I hope that's ok 😅 

The new version is already published, so I wanted to make sure our users go the correct documentation by default.

PR #1230, opened 2019-09-30
Redirect from the root the `/guide`

nik9000:
This should make it ever so slightly simpler to generate links to "the
docs".


nik9000:
:rewind:

PR #1231, opened 2019-09-30
Footer for air gapped docs

nik9000:
Adds a basic footer for air gapped docs. It links to our public site's
conditions of use at this point. It *probably* should link to some page
about how you are allowed to use the air gapped docs but we don't have a
page for that yet.


nik9000:
Failure is legit. fixing.

nik9000:
Thanks for looking @brianjolly and @JoshMock !

PR #1232, opened 2019-09-30
Add ci script to run tests

nik9000:
Adds a ci script to run the tests so we have the flexibility of changing
how we run the tests if we want to.


nik9000:
Thanks @JoshMock !

PR #1235, opened 2019-10-01
Air gapped docs always serve master

nik9000:
I'd configured the air gapped docs so if you access them as
`localhost/blah` they'd always give you the master branch. But I still
allowed you to grab other branches with `foo.localhost/blah`. The
trouble with that is that folks will tend to want to have `.`s in the
hostname when they deploy it.

This locks the branch to always be `master` regardless of the host
header.

### DIFF
--- a/integtest/spec/all_books_broken_link_detection_spec.rb
+++ b/integtest/spec/all_books_broken_link_detection_spec.rb
@@ -106,7 +106,6 @@ RSpec.describe 'building all books' do
       end
       include_examples 'all links are ok'
     end
-    shared_examples ''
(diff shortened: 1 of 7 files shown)


integtest/spec/all_books_broken_link_detection_spec.rb
```diff
@@ -106,7 +106,6 @@
       end
       include_examples 'all links are ok'
     end
-    shared_examples ''
```
nik9000: This line was making silly warnings.

nik9000:
:heart:

PR #1239, opened 2019-10-01
Switch the edit_me link implementation

nik9000:
Switches the edit_me link implementation from a TreeProcessor to a
DelegatingConverter. This is nice because, eventually, we'll want to
generate the edit_me links differently if we're converting to html
instead of docbook. Converters are the place where we can make that
call.

Also I can remove some rubocop suppressions which is nice.

### DIFF
--- a/resources/asciidoctor/.rubocop.yml
+++ b/resources/asciidoctor/.rubocop.yml
@@ -4,13 +4,11 @@ inherit_from: ../../.rubocop.yml
 
 Metrics/AbcSize:
   Exclude:
-    - lib/edit_me/extension.rb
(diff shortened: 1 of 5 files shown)


nik9000:
@JoshMock, this one is all ruby. Would you be comfortable looking at it or should I reach out to other folks?

nik9000:
Well that is quite broken!

nik9000:
@estolfo, could you have a look at this? I'm low key preparing for the day when we can generate html directly instead of docbook at switching these to a converter based implementation feels like a good start.

nik9000:
Thanks for reviewing @estolfo and @JoshMock !

PR #1246, opened 2019-10-03
ECS 1.2.0 release.

webmat:
1.2 branch should now be current

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -139,8 +139,8 @@ contents:
           -
             title:      Elastic Common Schema (ECS) Reference
             prefix:     en/ecs
-            current:    1.1
-            branches:   [ master, 1.1, 1.0 ]
+            current:    1.2
+            branches:   [ master, 1.2, 1.1, 1.0 ]
             index:      docs/index.asciidoc
             chunk:      1
             tags:       Elastic Common Schema (ECS)/Reference


benskelker:
run elasticsearch-ci/doc rebuild

Randy-312:
Argh... I was just about to rollout 7.4.0 to use 1.1.0.
I'll have to look at the changes.

Will this be in 7.4.1 ?

webmat:
No, not for 7.4.1. Patch releases are just for bugfixes :-) Likely 7.5.0.

Although for custom pipelines, you can use any version of ECS you want, no matter the stack version. You can also use ECS in Elasticsearch 6.

Randy-312:
Since the templates in the modules are patch specific (7.4.0), for the beats and elastic, we're evaluating the best way to handle the patches.  This is because we have to bring these grok patterns over to a separate logstash server.  Currently, we were going to do that based on patch version (ie 7.4.0), which would lead to a lot of extra work, or limiting the ability to accept patched versions.

If we CAN count on the ECS schema NOT changing except for a Minor version, then that we we have to update our Logstash pipelines every time a new version comes out.  I guess I'm asking if the Semantic policy around patches would apply to other Modules in the same version.

Can you point me to a plan around ECS Patch and the main product stream Version/ Patch plan?  @gose is going to work with us more on this too.

PR #1250, opened 2019-10-03
Improvements for book sources

nik9000:
This adds two features that are sort of interlinked in their
implementation but not so much in their use. We want them both though, I
think:

1. Allows books to explictly depend on files in the docs repo so they
are rebuilt when those files are changed.
2. Create attributes for the "root" of each repository. This should make
a lot of the `../../../../` paths unneeded.

So the second feature allows you to write includes like:
```
include::{elasticsearch-root}/docs/reference/glossary.asciidoc[]
```

instead of:
```
include::{docdir}/../../../../elasticsearch/docs/reference/glossary.asciidoc[]
```

The first feature combined with the second feature lets you write this
to include the shared files:
```
include::{docs-root}/shared/versions/stack/{source_branch}.asciidoc[]
include::{docs-root}/shared/attributes.asciidoc[]
```

instead of:
```
include::{asciidoc-dir}/../../shared/versions/stack/{source_branch}.asciidoc[]
include::{asciidoc-dir}/../../shared/attributes.asciidoc[]
```

Which is nice. In addition, the first feature lets you write:
```
              -
                repo:   docs
                path:   shared/attributes.asciidoc
              -
                repo:   docs
                path:   shared/versions/stack/{version}.asciidoc
```

in `conf.yaml` to cause a book to be rebuilt if either:
* `attributes.asciidoc` changes
* `shared/versions/stack/master.asciidoc` changes. Assuming we're
building the master branch of the book. It'll be `7.x.asciidoc` if we're
building `7.x`, etc.

You can also declare a dependency on
`shared/versions/stack/current.asciidoc` and we'll cause a rebuild if
either `current.asciidoc` changes to point to a new file *or* the file
that it points to changes.


nik9000:
@olksdr, I'm going to try and work through the last NOCOMMITs this afternoon. But I figure because this one is all OO-perl you're the best reviewer we have for it. There is a fair bit of ruby too but I think it is pretty uncontroversial stuff.

nik9000:
The error is real. I'll get to it in the morning. Instead of checking the NOCOMMIT I did #1252. It almost cuts the time to run the tests in half.

nik9000:
@lcawl, I think you should also have a look at this one, at last the README changes and the description.

nik9000:
OK @olksdr, I think this is ready for you when you have time.

lib/ES/DocsRepo.pm
```diff
@@ -0,0 +1,94 @@
...
+    $dir = Path::Class::dir( $dir );
+    my $self = $class->SUPER::new(
+        name      => 'docs',
+        git_dir   => $dir->subdir( '.git' ),
+        tracker   => $tracker,
+        url       => 'git@github.com:elastic/docs.git',
+        keep_hash => 0,
+    );
+    $self->{dir} = $dir;
+    return $self;
```
olksdr: I think you still need to `bless` the instance before returning it. Something like 
```perl
return bless $self, $class;
```

But can be that it gets blessed from the parent

lib/ES/Repo.pm
```diff
@@ -74,12 +77,13 @@ sub has_changed {
...
         return 'changed' unless $old_info;
 
-        $new = sha_for($branch) or die(
+        $new = $self->sha_for_branch( $branch ) or die(
                 "Remote branch <origin/$branch> doesn't exist in repo "
                 . $self->name);
     }
     my $new_info = $new;
     $new_info .= '|asciidoctor' if $asciidoctor;
+    say "ASDFADF " . $self->{name} . " $branch $path $old_info -> $new_info";
```
olksdr: debug `say`?

build_docs.pl
```diff
@@ -531,7 +555,7 @@ sub init_target_repo {
     my ( $repos_dir, $temp_dir, $reference_dir ) = @_;
 
     my $target_repo = ES::TargetRepo->new(
-        dir         => $repos_dir,
+        git_dir     => $repos_dir->subdir('target_repo.git'),
```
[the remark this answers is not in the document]
nik9000: I this case we really prefer to keep this in a stable place because it prevents us from having to pull as much.

Long before I got to it `build_docs --all` sacrificed being run in parallel for less git traffic. For the most part we don't want to run it in parallel anyway because its job is to commit to a particular repository.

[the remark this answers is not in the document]
olksdr: I see. Thanks for the explanation!  Then it doesn't really matter 😄

lib/ES/DocsRepo.pm
nik9000: The parent blesses it, yeah.

nik9000:
Let's get this in!

Thanks @olksdr!

PR #1252, opened 2019-10-04
Speed up integration tests

nik9000:
```
Before: 10m47.920s
After:   6m1.201s
```

The integration tests have felt slower to me so I did a little digging.
`top` told me that git was pretty busy during the integration tests and
it dawned on me that we put the fonts in the built-docs repo now and
they are pretty big.

This speeds up the tests by creating a "shell" repository before most
tests that already has the fonts in it and then clonging the "target"
repository from that shell before every test. That way we only have to
pay the cost of `git add`ing the font files one time.

### DIFF
--- a/integtest/.rubocop.yml
+++ b/integtest/.rubocop.yml
@@ -1,0 +1,6 @@
+inherit_from: ../.rubocop.yml
+
+
+Metrics/ClassLength:
+  Exclude:
+    - spec/helper/dest.rb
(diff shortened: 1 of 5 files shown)


integtest/.rubocop.yml
```diff
@@ -0,0 +1,6 @@
+inherit_from: ../.rubocop.yml
+
+
+Metrics/ClassLength:
+  Exclude:
+    - spec/helper/dest.rb
```
nik9000: There are a couple of TODOs that I haven't gotten around to for a few months that will make this pass once I do them. I didn't want to do them as part of this change. I'll get them in a follow up and remove this.

nik9000:
:rocket:

PR #1254, opened 2019-10-04
Check warnings from asciidoctor before next steps

nik9000:
Before this commit we bundled executing `asciidoctor`, `xmllint`, and
`xsltproc` together, checking warnings after the whole process
succeedes. If any step in the process fails we'd log the errors for
*that* *step* and exit non-0. The trouble with this behavior is that
we'd throw out the warnings from the previous steps. This is especially
bad because the warnings from `asciidoctor` are *much* easier to debug
that errors from `xmllint`.

This commit causes us to check the warnings after each step. If there
are any warnings we log them and exit non-0. This is will effectively
prefer warnings from asciidoctor if there are any which should make
debugging problems like bad links *much* simpler.

### DIFF
--- a/integtest/spec/single_book_spec.rb
+++ b/integtest/spec/single_book_spec.rb
@@ -738,4 +738,22 @@ RSpec.describe 'building a single book' do
       LOG
     end
   end
+  context 'when a referenced id is missing' do
+    convert_before do |src, dest|
+      repo = src.repo_with_index 'src', <<~ASCIIDOC
+        <<missing-ref>>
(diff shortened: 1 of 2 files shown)


nik9000:
@olksdr this is another one for you!

nik9000:
I don't think it'll matter, but just for extra paranoia, @elasticmachine please run elasticsearch-ci/docs  rebuild.

nik9000:

nik9000:
Thanks for reviewing @olksdr!

PR #1255, opened 2019-10-04
Automatically relativize links to elastic.co

### LINKED ISSUE #1029: Move to root-relative URLs for links between books
Most of the Elastic documentation now uses attributes to fill in the URL for links between books. For example, in https://github.com/elastic/stack-docs/blob/7.3/docs/en/install-upgrade/installing-stack.asciidoc:

> Install the Elastic Stack products you want to use in the following order:
> . Elasticsearch ({ref}/install-elasticsearch.html[install instructions])
> . Kibana ({kibana-ref}/install.html[install])
> . Logstash ({logstash-ref}/installing-logstash.html[install])
> . Beats ({beats-ref}/getting-started.html[install instructions])
> . APM Server ({apm-server-ref-70}/installing.html[install instructions])
> . Elasticsearch Hadoop ({hadoop-ref}/install.html[install instructions])

These attributes ( {ref}, {kibana-ref}, etc) are defined in https://github.com/elastic/docs/blob/master/shared/attributes.asciidoc

We should consider changing the values in those attributes such that they're root-relative instead of specifying the full URL. This would make the links work better in scenarios where we're providing a full set of offline documentation.

It would mean you wouldn't be able to check your links if you're building a single book, however. Since we do link checks when we build *all* documentation, however, and that's the build we're running in our PR checks now, I think that is okay.

nik9000:
This adds an asciidoctor plugin to automatically relativize links with a
particular prefix and configures it with `https://www.elastic.co/

Closes #1029

### DIFF
--- a/integtest/Makefile
+++ b/integtest/Makefile
@@ -24,4 +24,4 @@ rubocop:
 
 .PHONY: rspec
 rspec:
-	rspec -e 'README'
+	rspec
--- a/integtest/Makefile
+++ b/integtest/Makefile
@@ -24,4 +24,4 @@ rubocop:
 
 .PHONY: rspec
 rspec:
-	rspec
+	rspec -e 'README'
--- a/integtest/html_diff
+++ b/integtest/html_diff
@@ -47,6 +47,8 @@ def normalize_html(html):
                         '"programlisting prettyprint lang-console"')
     html = html.replace('"pre_wrapper lang-js"',
                         '"pre_wrapper lang-console"')
+    # Asciidoctor relativizes https://www.elatic.co/ into /.
+    html = html.replace('https://www.elastic.co/', '/')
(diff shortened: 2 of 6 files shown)


nik9000:
@elasticmachine, run elasticsearch-ci/docs  rebuild

nik9000:
The failure is because I've not implemented the change for Asciidoc because, well, I can't. I'll see about normalizing it away.

nik9000:
I think that does it.

nik9000:
@JoshMock or @estolfo, would you mind having a look at this? It *should* help quite a bit with the air-gapped docs deploy. Not as much "sneaking" out of the site into the public site.

nik9000:

nik9000:

PR #1256, opened 2019-10-04
Add explicit dependency on docs repo for some books

nik9000:
This adds an explicit dependency on the docs repo for all books build
from `stack-docs`. This will expose the `docs-root` attribute to those
books and automatically rebuild the books if any of those file change.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -79,6 +79,12 @@ contents:
                 repo:   logstash
                 path:   docs/
                 exclude_branches:   [ 6.8, 6.7, 6.6, 6.5, 6.4, 6.3, 6.2, 6.1, 6.0, 5.6, 5.5, 5.4, 5.3, 5.2, 5.1, 5.0 ]
+              -
+                repo:   docs
+                path:   shared/versions/stack/{version}.asciidoc
+              -
+                repo:   docs
+                path:   shared/attributes.asciidoc
           -
             title:      Getting Started
             prefix:     en/elastic-stack-get-started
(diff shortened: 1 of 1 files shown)


nik9000:
I'd like to merge #1257 before merging this one.

conf.yaml
```diff
@@ -79,6 +79,12 @@ contents:
                 repo:   logstash
                 path:   docs/
                 exclude_branches:   [ 6.8, 6.7, 6.6, 6.5, 6.4, 6.3, 6.2, 6.1, 6.0, 5.6, 5.5, 5.4, 5.3, 5.2, 5.1, 5.0 ]
+              -
+                repo:   docs
+                path:   shared/versions/stack/{version}.asciidoc
+              -
```
lcawl:
```suggestion
                 exclude_branches:   [ 6.2, 6.1, 6.0, 5.6, 5.5, 5.4, 5.3, 5.2, 5.1, 5.0 ]
              -
```

conf.yaml
```diff
@@ -79,6 +79,12 @@ contents:
...
+                repo:   docs
+                path:   shared/attributes.asciidoc
           -
```
lcawl:
```suggestion
             exclude_branches:   [ 6.2, 6.1, 6.0, 5.6, 5.5, 5.4, 5.3, 5.2, 5.1, 5.0 ]
          -
```

nik9000:
It turns out we didn't need the explicit `exclude`s but they probably do make it a little faster.

nik9000:
@lcawl, what do you think of this as a start?

PR #1257, opened 2019-10-04
Make sure we don't rebuild at a bad time

nik9000:
It is important that changes to the docs repo don't trigger a rebuild
when `--keep_hash` is specified or else every time we change this
file we'll rebuild all books in every PR build. That is a waste of
time and a potential source of spurious errors.

This PR does exactly that: it adds tests that make sure we don't rebuild
and it makes a small change to the initialization fo the docs repo to
make the test pass.

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -643,7 +643,11 @@ sub init_repos {
 
     # Setup the docs repo
     # We support configuring the remote for the docs repo for testing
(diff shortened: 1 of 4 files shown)


nik9000:
@olksdr this is another one for you!

nik9000:
Thanks!

PR #1258, opened 2019-10-04
Drop the user option

nik9000:
We haven't supported it for a while. We rely on git over ssh for this.


nik9000:
Here is a small one, @olksdr and @JoshMock .

nik9000:
Thanks for reviewing @olksdr!

PR #1259, opened 2019-10-06
Clean integtest/Makefile

nik9000:
It has some variables that are left over from when we were writing tests
in the Makefile directly. We don't need them any more.


nik9000:
:zap:

PR #1262, opened 2019-10-07
Fancy --sub_dir and --keep_hash

nik9000:
This adds a fancy interaction between `--sub-dir` and `--keep_hash` that
is meant to make sure we build *exactly* the right docs for pull
requests.

When we first deploy pull request builds we targeted them at the tip of
the pull request branch. This caused trouble because the tip of the pull
request branch was often behind the last successful build. Building the
docs against that would often cause spurious failure from out of date
links. Merging `master` into the PR branch usually fixed it but it was
annoying.

We fixed this problem by switching our target to the result of merging
the pull request branch into the target branch. This was *better*
because it had fewer spurious failures, but often caused us to think
that the docs changes were larger than they actually were because the
target branch often had changes in it that hadn't yet been picked up by
the docs build. This was annoying because the `diff`s generated by the
docs builds had changes that weren't part of the PR. Which made it
difficult to track down what actually changed.

This solves the problem by first merging the `--sub_dir`ed branch into
the last successful docs build. Thus we only build *exactly* what is in
the PR. We trigger this behavior by checking for both the `--keep_hash`
and `--sub_dir` flags *and* checking for outstanding changes in the
`--sub_dir`ed branch.

This behavior might end up being slower sometimes because merging can be
expensive, but it is usually less expensive than building a book we
don't have to build. So maybe it'll all come out in the wash.

### DIFF
--- a/lib/ES/Repo.pm
+++ b/lib/ES/Repo.pm
@@ -4,9 +4,10 @@ use strict;
 use warnings;
 use v5.10;
 
-use Path::Class();
+use Cwd;
 use Encode qw(decode_utf8);
 use ES::Util qw(run);
+use Path::Class();
 
 use parent qw( ES::BaseRepo );
 
@@ -139,8 +140,7 @@ sub prepare {
     my $dest = $dest_root->subdir( $prefix );
 
     if ( exists $self->{sub_dirs}->{$branch} ) {
-        my $source_root = $self->{sub_dirs}->{$branch};
-        $self->_extract_from_dir( $source_root, $dest, $path );
+        $self->_prepare_sub_dir( $title, $branch, $path, $dest );
         return $dest;
     }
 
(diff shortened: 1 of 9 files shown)


nik9000:
@olksdr This one kind of terrifies me a little, but I think it'll be ok. I did a fair bit of playing with it locally and everything *seemed* right. If we merge this then all of the PR builds will immediately start doing noop merges because we build against github's pre-merged hash. When we're good and ready we can switch them.

lib/ES/Repo.pm
```diff
@@ -151,17 +151,94 @@ sub prepare {
...
+    }
 
-    return $dest;
+    my $has_uncommitted_changes = eval {
+        local $ENV{GIT_WORK_TREE} = $source_root;
+        local $ENV{GIT_DIR} = $ENV{GIT_WORK_TREE} . '/.git';
+        run qw(git diff-index --quiet HEAD --);
+        1;
+    };
+    unless ( $has_uncommitted_changes ) {
```
olksdr: shouldn't it be `if` here ? 
```perl
if ($has_uncommitted_changes ) {
 # execute the code inside ?
}
```

olksdr:
@nik9000  before I finished this review you'd already fixed few things 😄

nik9000:
> @nik9000 before I finished this review you'd already fixed few things smile

Sorry!

lib/ES/Repo.pm
nik9000: Weird! I've got something backwards....

nik9000: OK - here is what happened `--exit-code` causes `run` to throw an exception if there *are* changes. Which causes `eval` to return 0. So, you could say that I named the variable backwards. Fixing.

nik9000:
I've reverted this because it looked like it was causing trouble with the PR builds that it was trying to help. I'll investigate further.

lib/ES/Repo.pm
olksdr: yeah 😄 the name was confusing a little bit.

PR #1264, opened 2019-10-08
GitHub user via code

JoshMock:
Putting optional env vars in the Dockerfile `CMD` was breaking the Helm chart deployment. Moving the logic into the code.

My first time committing production Perl!

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -18,6 +18,7 @@ use Cwd;
 use FindBin;
 use Data::Dumper;
 use XML::LibXML;
+use URI;
 
 BEGIN {
     $Old_Pwd = Cwd::cwd();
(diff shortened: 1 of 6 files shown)


JoshMock:
Hmm. CI error says `INFO:build_docs:Can't locate URI.pm`. @nik9000 maybe you can school me on Perl module imports in the morning?

build_docs.pl
```diff
@@ -555,9 +556,14 @@ sub init_target_repo {
 #===================================
     my ( $repos_dir, $temp_dir, $reference_dir ) = @_;
 
+    my $git_uri = URI->new($Opts->{target_repo});
+    if ( $ENV{GITHUB_USER} ){
+        $git_uri->userinfo( $ENV{GITHUB_USER} . ":" . $ENV{GITHUB_PASS} );
+    }
+
```
nik9000: I think we should probably only try and do this if it *look* like an http uri. The regular docs build uses git-over-ssh and probably should get this decoration. Same for tests that reference the filesystem.

JoshMock: Added a check in cccabfdc3ab2f26f094309ce54701a1f1da2ad2b to ensure it's using https before attaching creds to the URI.

Also, as of e3063d6126ab1faab103c778f145f6408e767371, it falls back to the originally-passed argument if the string's scheme is not https.

JoshMock:
Woo hoo, all checks have passed!

I just republished Docker image version 12 now, overwriting it since it's not used by anything yet. @nik9000 this is ready for review again!

JoshMock:
thanks for the helpful review @nik9000!

PR #1265, opened 2019-10-08
Add 3.x branch to Ruby APM Agent

mikker:
We've released a new major version 🎉


mikker:
@bmorelli25 Merging this now so we can get the docs on the site. Let me know if things need to be different 😊

bmorelli25:
Nice!

PR #1268, opened 2019-10-08
Adjust ECS version mapping for unreleased Elastic docs

webmat:
ECS 1.2 came out in advance of 7.5, and should be associated with ECS 1.2
when it comes out.

Until 7.5 is released however, 7.x and 'master' should be associated with ECS 1.2.

cc @karenzone @benskelker


benskelker:
LGTM
Let me know if you want me to merge it

PR #1270, opened 2019-10-08
Support redirects in preview

nik9000:
This add support for parsing the redirects.conf file that we've been
including in every do build for a while. It is *technically* an nginx
config file but it is safe to treat it as a list of regex substitutions.


nik9000:
@JoshMock, I think we can be more efficient here but I think this is a good start.

PR #1272, opened 2019-10-08
[docs] Update current for APM Node.js Agent

bmorelli25:
Updates `current` links for the APM Node.js Agent.


bmorelli25:
@Qard -- Tagging you here in case I'm not around when you do the next Node.js release. This PR should be merged immediately after release to prevent docs from 💥

Qard:
Immediately after master merges to 3.x, right?

bmorelli25:
Correct

PR #1277, opened 2019-10-09
Obs docs box

bmorelli25:
### DO NOT MERGE THIS PR

TLDR: **[CLICK HERE FOR A PREVIEW](http://docs_1277.docs-preview.app.elstc.co/guide/en/apm/get-started/current/index.html)** that mostly works (ish)

--- 
Over in https://github.com/elastic/observability-dev/issues/14, we've been trying to figure out how best to structure our Observability documentation. A general plan was eventually formed: Each solution's docs would live in its own documentation book, and we would explore ways to connect this content via the doc build. This could mean a [related docs widget](https://github.com/elastic/observability-dev/issues/14#issuecomment-534637225), an [aware navigation](https://github.com/elastic/observability-dev/issues/14#issuecomment-535991212),  [dynamic documentation](https://docs.google.com/document/d/1Ty0m-N4FTWN6zCODGxVk90KLj4RNODY48v0yPwm880U/edit#heading=h.4ob4rpg4dw57) based on a user's needs, or a host of other options discussed in weekly doc huddles.

The problem with most of these solutions is that they're not quick fixes. And although effort is already underway for a long-term solution in https://github.com/elastic/Design/issues/1650#issuecomment-540981241, we could still do with an  easier, short-term solution.

I decided to run with @dedemorton's mockup in [this comment](https://github.com/elastic/observability-dev/issues/14#issuecomment-534637225), also shown below.
dedesmockup

For this PR, I tried to simplify the idea even more. It is my understanding that we can't remove the "Getting Started Videos" widget that displays in the documentation, but what about adding to it? So, I came up with this:
Screen Shot 2019-10-09 at 4 01 50 PM

---

But Brandon, won't that just push the TOC even _further_ down the page? A little, yes. But here's a side-by-side comparison -- with a few CSS adjustments, it's not as bad as you'd think.

Screen Shot 2019-10-09 at 4 00 16 PM

---

But Brandon, doesn't that just add more links to our documentation? Yes it does, but it's so painfully hard to navigate between these solutions right now. I think this could be a good short-term fix.

---

But Brandon, isn't this only specific to Observability? Yes, but it wouldn't necessarily have to be. Each documentation book could define their own "related docs" content as needed.

---

At the end of the day, editing our doc build is not my wheelhouse. In fact, this PR won't work as-is. I'm not sure:
1. If it's possible to only apply a template to certain documentation books -- as this would only need to be applied to Observability content.
2. Just how flexible my CSS changes are across different devices 🤷‍♂
3. How to version links in this widget to point to the correct version of the documentation.

The idea here was mostly to provide the simplest thing I could think of as a **short-term** solution. 

---

If anyone has thoughts, or more importantly, criticism, I'd love to hear it. cc @dedemorton, @debadair, @lcawl, @nik9000, @Titch990 @tbragin

### DIFF
--- a/resources/web/style/rtpcontainer.pcss
+++ b/resources/web/style/rtpcontainer.pcss
@@ -6,8 +6,8 @@
 #rtpcontainer {
   .mktg-promo {
     border-radius: 5px;
-    margin-bottom: 40px;
-    padding: 20px;
+    margin-bottom: 20px;
+    padding: 0px 20px;
     border: 1px solid #ccc;
   }
   ul.icons {
@@ -15,7 +15,7 @@
     li {
       align-items: center;
       list-style: none;
-      min-height: 60px;
+      min-height: 35px;
       padding-left: 0.6em;
       display: flex;
       :global(a.link) {
@@ -40,8 +40,8 @@
         background-size: 55%;
         background-position: center center;
         content: '';
-        height: 2.5em;
-        width: 2.5em;
+        height: 1.5em;
+        width: 1.5em;
         display: block;
         float: left;
         margin-left: -2.5em;
@@ -56,6 +56,36 @@
       &.icon-elasticsearch-white {
         &:before {
           background-image: inline(img/rtpcontainer/logo-elasticsearch-32-white.svg);
+        }
+      }
+      &.icon-apm-white {
+        &:before {
+          background-image: inline(img/rtpcontainer/logo-apm-32-white.svg);
+        }
+      }
+      &.icon-logging-white {
+        &:before {
+          background-image: inline(img/rtpcontainer/logo-logging-32-white.svg);
+        }
+      }
+      &.icon-uptime-white {
+        &:before {
+          background-image: inline(img/rtpcontainer/logo-uptime-32-white.svg);
+        }
+      }
+      &.icon-metrics-white {
+        &:before {
+          background-image: inline(img/rtpcontainer/logo-metrics-32-white.svg);
+        }
+      }
+      &.icon-code-search-white {
+        &:before {
+          background-image: inline(img/rtpcontainer/logo-code-search-32-white.svg);
+        }
+      }
+      &.icon-stack-white {
+        &:before {
+          background-image: inline(img/rtpcontainer/logo-stack-32-white.svg);
         }
       }
       &.logo-kibana-white {
--- a/resources/web/style/toc.pcss
+++ b/resources/web/style/toc.pcss
@@ -55,7 +55,108 @@
       > span {
         background-color: #efefef;
         border-bottom: 1px solid #ddd;
+        border-left: 1px solid #ddd;
         font-size: 1em;
+        background-position: 0 11px;
+        padding-left: 20px;
+      }
+      > ul > li {
+        > span {
+          padding-left: 40px;
+          background-position: 20px 8px;
+        }
+        > ul > li {
+          > span {
+            padding-left: 60px;
+            background-position: 40px 8px;
+          }
+          > ul > li {
+            > span {
+              padding-left: 80px;
+              background-position: 60px 8px;
+            }
+            ul > li > span {
+              padding-left: 80px;
+              background-image: none;
+            }
+          }
+        }
+      }
+    }
+
+    .experimental,
+    .beta,
+    .coming,
+    .added,
+    .deprecated {
+        display: none;
+    }
+  }
+  .toc2 {
+    ul {
+      margin: 0 0 1em 0;
+      padding: 0;
+      border: 1px solid #ddd;
+      ul {
+        /* Hide all sub-toc elements by default. See rules with `.show` that
+         * show specific sub-toc elements. */
+        display: none;
+      }
+    }
+    li {
+      margin: 0;
+      padding: 0;
+      list-style: none;
+      ul {
+        border: none;
+      }
+    }
+    span {
+      display: block;
+      padding: 0.5em 0;
+      font-size: 0.85em;
+      line-height: 1.3em;
+    }
+
+    .current_page {
+      font-weight: bold;
+    }
+
+    .collapsible.show {
+      > ul {
+        display: block;
+      }
+      > span {
+        background-image: inline("img/minus.png");
+        background-color: white;
+        border-bottom: 1px solid white;
+      }
+    }
+    .collapsible {
+      > span {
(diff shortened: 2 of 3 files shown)


nik9000:
> It is my understanding that we can't remove the "Getting Started Videos" widget that displays in the documentation,

Really it is that we aren't the only owner of the `rtpcontainer`. That is sort of jointly owned between us and the web team. The web team can manipulate it on page load for a/b testing and things. 

I do love the idea though! Would we put things in that corner based on which book you are in or sort of hard code them like you proposed?

I wonder if we should actually use the a/b testing capabilities of the rtpcontainer to try this out! @brianjolly, what do you think of this?

bmorelli25:
> I do love the idea though! Would we put things in that corner based on which book you are in or sort of hard code them like you proposed?

I think initially we would just use this in the Observability documentation. I really like the idea of a/b testing it to see how well it works. If it works well, other books could eventually adapt it to their needs. I think we have to be careful though, in its current form, I don't see this as a good long term solution. But, if we could eventually iterate on this, perhaps into some sort of [aware nav](https://github.com/elastic/observability-dev/issues/14#issuecomment-535991212), that'd be cool.

brianjolly:
The `rtpcontainer` is managed by Demand Gen. I'll check to see who would be best to talk to about this one.

sfingerhut:
Just so we don't end up in too much of a muck - why not have the Observability content be more native directly in the docs, vs. on the right side with this RTP content (that's actually a place where the contact can be dynamically changed out). It could be that you separate them so that changes in the dynamic RTP content don't mess up the intention of the observability items.

Titch990:
>Just so we don't end up in too much of a muck - why not have the Observability content be more native directly in the docs . . . 

I'm certainly in favour of separating the **Observability docs** section from the **Getting started videos**.  I assume that's what you mean @sfingerhut?

The existing box is a convenient place to put the Observability nav info, I agree, and it's better than nothing if there is no other option. But by coupling these two largely unrelated things, I think we may be storing up problems later. I'd much prefer to decouple them from the outset of we can. They are quite different: they contain different information for different kinds of readers, solve different user problems, and probably need to be presented under different conditions. 

If it is a separate box, then presumably it will be easier to implement something similar for other areas of documentation that are split over several books, for example, Beats? Observability is a particular problem because I can think of many user journeys through the Observability documentation which will cross and re-cross inter-book boundaries, perhaps more so than in other areas. But it's not the only area with a problem. I suspect that decoupling the two boxes will make the solution more generally applicable rather than being an Observability special case?

>  I really like the idea of a/b testing it to see how well it works

Perhaps I'm missing something here, but what would we be testing and what would it tell us?

Do we have the tools to collect the information we need to determine how "well it works"? I'm not sure what the current a/b testing is looking at (conversion rates, perhaps?), but I suspect the info we'd be interested in would be different. In our case we'd be interested in finding out whether users spot the box and work out that it's a navigation aid for the Observability docs, and whether using it stops them getting less lost.

I'm not a web analytics expert, but I'd assume we'd need a measure of how lost/frustrated they get without the nav box, so we would compare it with how much more successfully they find all the information they need with the nav box. Are we going to be able to measure this easily without explicitly asking a question like "Did you find what you were looking for today?". If users only view a page or two of the Observability docs, then click away, how can we distinguish between those that found all they wanted, and those who failed to find all they wanted, and end up going to the forums, or support or giving up?

Even if we do have the tools, would they tell us anything we don't already know? With a user journey that requires multiple inter-doc jumps, the need for an overall nav box is a bit of a no-brainer. Now, if we were testing two different kinds of inter-doc navigation, then I'd understand the use of a/b testing.

Nevertheless, I'm happy with any solution that can be implemented quickly and relatively painlessly. If this is it, then let's go! Thanks @bmorelli25 for suggesting it.

bmorelli25:
> Just so we don't end up in too much of a muck - why not have the Observability content be more native directly in the docs, vs. on the right side with this RTP content (that's actually a place where the contact can be dynamically changed out). It could be that you separate them so that changes in the dynamic RTP content don't mess up the intention of the observability items.

@sfingerhut -- That's the long term plan. You can see some effort is underway in https://github.com/elastic/Design/issues/1650#issuecomment-540981241. The problem is that we're still a long way from that plan coming to fruition. 

I agree, it's probably a messy fix to lump this content together -- a separate box from the RTP content makes more sense. But this was the simplest solution I could get working _by myself_. Again, really just trying to propose something easy that we can do in the short-term.

sfingerhut:
sorry for delay in responding. Again, let's make things far less complex by separating the boxes and having the Observability box on the top perhaps (it's more directly relevant). It shouldn't be too hard. Then render the Top Videos (or whatever we dynamically place) below it.

bmorelli25:
That sounds good to me

debadair:
TBH, the idea of bolting on yet another nav widget because our existing nav/structure doesn't work well makes me cringe. We already have too many different nav elements on the page--adding another one addresses one need, while making the overall nav experience more confounding. 

That said, I understand the motivation and desire for a quick fix. If we do this:

- It definitely needs to be separate from the RTP box.
- It needs to be styled so it looks like "doc nav". 

I don't love how the "On this page" block is styled, but I'd opt for using the same look and feel and avoid having the links look like the video links or introducing yet another block style. (Perhaps we could steal the revised styling from Adrian's mock-ups for both?)

I understand the desire to get "Observability" into the heading, but I'd lean toward using the more generic "Related docs" heading.  To me, that's a better fit with the existing nav.

bmorelli25:
Updated to look more like the doc nav per @debadair's request:
Screen Shot 2019-11-19 at 10 36 08 AM

dedemorton:
Just my opinion, but having the same look/feel actually makes this sea of links harder for my eyes to process. It might help to add some kind of heading to the TOC, but I dunno.

IMO this would work a lot better with the proper book navigation on the left. I wonder if that's something we can change sooner rather than later?

bmorelli25:
> Just my opinion, but having the same look/feel actually makes this sea of links harder for my eyes to process.

Yeah, I think I like the original better as well.

> IMO this would work a lot better with the proper book navigation on the left. I wonder if that's something we can change sooner rather than later?

I agree. That'd be awesome if we could!

gtback:
@bmorelli25 Are there important things we need to keep from this PR? I'm just trying to clean up old PRs.

bmorelli25:
No. We can close this.

PR #1278, opened 2019-10-09
Update Azure Marketplace docs to 7.4

russcam:
This PR adds version 7.4  to the Azure Marketplace docs and sets 7.4 as the current version


russcam:
This should be merged once 7.4.0 is listed in the Azure Marketplace. I'll update when this happens

russcam:
thanks @nik9000. Merging this in now as 7.4.0 is live on the Azure Marketplace

PR #1280, opened 2019-10-10
Fix some duplication problems in buffer itr

nik9000:
The code that emits the `html` for our `diff` endpoint if pretty chatty
and we wrap it in a "buffering iterator" to save up chunks until they
are worth sending. It had a bug where it'd duplicate stuff. This fixes
that bug and adds tests.

### DIFF
--- a/preview/__test__/buffer_itr.test.js
+++ b/preview/__test__/buffer_itr.test.js
@@ -24,6 +24,7 @@ const bufferItr = require("../buffer_itr.js");
 
 const bufferedCollect = async itrGen => {
   const all = [];
+  // Use 20 chunking point to make it easy to write test strings.
   for await (const c of bufferItr(itrGen(), 20)) {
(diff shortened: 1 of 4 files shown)


preview/__test__/buffer_itr.test.js
```diff
@@ -0,0 +1,68 @@
...
+ * under the License.
+ */
+
+'use strict';
+
+const bufferItr = require("../buffer_itr.js");
+
+const bufferedCollect = async itrGen => {
+  const all = [];
+  for await (const c of bufferItr(itrGen(), 20)) {
```
JoshMock: Is there any rhyme or reason to "long" being defined as >20? Might be worth dropping a comment here to be clear.

nik9000: Just because it is long enough to trigger the grouping. I'll add a comment.

PR #1283, opened 2019-10-10
Allow explicit suppress of MIGRATION warnings

nik9000:
MIGRATION warnings are problems that we should fix in newer versions of
a book but aren't worth fixing in old versions of a book. But some books
are old and aren't going to be updated, probably ever. Those books
should just suppress their MIGRATION warnings all together. This allows
you to configure that.


nik9000:
Thanks @olksdr!

PR #1290, opened 2019-10-11
Don't choke on utf8 in logs

nik9000:
We helpfully tried to decode utf8 when we to a `git log`. The trouble
with this is that we've *already* decoded those characters so this blows
up.


nik9000:
More fun with git. Sort of. This is more like fun with people who's names aren't in low Ascii.

nik9000:

PR #1292, opened 2019-10-11
Fix error reporting when book depends on docs

nik9000:
When a book fails to build we try to list the latest commits from each
of the sources of the book. When one of the sources of that book is the
docs repo *and* the docs repo was cloned with `--reference` then we were
failing to list the source and crashing. This fixes that by mounting the
`--reference`d repo.

### DIFF
--- a/build_docs
+++ b/build_docs
@@ -182,7 +182,7 @@ def run_build_docs(args):
         return repo_mount + path[len(repo_root):]
 
     def mount_alternates(path_in_repo):
-        # We're prefer to ues --absolute-git-dir here but not all users have it
+        # We'd prefer to use --absolute-git-dir here but not all users have it
         git_dir = subprocess.check_output(
                 ['git', 'rev-parse', '--git-dir'],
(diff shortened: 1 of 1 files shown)


nik9000:
This should fix the error thrown by https://github.com/elastic/elasticsearch/pull/47888

nik9000:
This one is yet more fun with git. Docker makes things interesting!

build_docs
```diff
@@ -182,10 +182,12 @@ def run_build_docs(args):
         return repo_mount + path[len(repo_root):]
 
     def mount_alternates(path_in_repo):
+        # We're prefer to ues --absolute-git-dir here but not all users have it
```
olksdr:
```suggestion
        # We're prefer to use --absolute-git-dir here but not all users have it
```

JoshMock:
```suggestion
        # We'd prefer to use --absolute-git-dir here but not all users have it
```

PR #1293, opened 2019-10-11
Add an attribute for the root of the ES reference

nik9000:
We include files from the Elasticsearch reference in a bunch of
different books and it is big to type
`{elasticsearch-root}/docs/reference`. That takes up half of a line!
This adds an attribute, `{elasticsearch-reference}` to the root of the
reference. It is slightly shorter and hopefully a little less breakable.


nik9000:
Thanks @lcawl !

PR #1295, opened 2019-10-11
Explicitly depend on shared docs in logstash

nik9000:
Add explicit dependencies on the files in the docs repo that logstash
uses so that when they change we automatically rebuild the docs.


nik9000:
@lcawl, this one is waiting on you!

lcawl:
I've added the references to attributes62.asciidoc too.  If the checks succeed, LGTM

PR #1297, opened 2019-10-11
Explicitly depend on shared docs in cloud-on-k8s

nik9000:
The cloud-on-k8s docs use some shared files in the docs repo. This adds
an explicit dependency on those files so that when they change we can
automatically rebuild the docs.


lcawl:
In [0.9](https://raw.githubusercontent.com/elastic/cloud-on-k8s/0.9/docs/index.asciidoc ) and [0.8](https://raw.githubusercontent.com/elastic/cloud-on-k8s/0.8/docs/index.asciidoc), there's a dependency on shared/versions/stack/7.x.asciidoc. Should that be added here as well?

nik9000:
> In [0.9](https://raw.githubusercontent.com/elastic/cloud-on-k8s/0.9/docs/index.asciidoc) and [0.8](https://raw.githubusercontent.com/elastic/cloud-on-k8s/0.8/docs/index.asciidoc), there's a dependency on shared/versions/stack/7.x.asciidoc. Should that be added here as well?

I didn't see that! Probably.

nik9000:
@lcawl I updated this one.

Issue #1300, opened 2019-10-15
Local docs builds are setting all links, including links external to the doc being built, to localhost


Titch990:
This issue has been observed in the stack-docs repo, while building the Infrastructure Monitoring Guide (command `docbldinf`). So far it doesn't seem to affect the Kibana docs.

Previously, in a local docs build, external links were used exactly as coded, so for example, in the built docs, an external link from the Infrastructure Monitoring Guide to a page in the Kibana doc may look like this: https://www.elastic.co/guide/en/kibana/master/infra-ui.html. A link from the same document to the Elastic cloud sign-up page would be https://elastic.co/cloud/elasticsearch-service/signup.

Now, all these addresses are being resolved to localhost addresses, so the first becomes http://localhost:8000/guide/en/kibana/master/infra-ui.html and the second becomes http://localhost:8000/cloud/elasticsearch-service/signup.

Neither of these work satisfactorily for local builds. 

In the first case, it isn't possible any longer to check external links to other docs unless you build all the docs locally. I write a bit, check a bit, write a bit more, check a bit more, and my docs are going to have *lots* of external links to several other Elastic docs. So this effectively means every docs build will need to be a full docs build, and I'm going to get through an awful lot of coffee waiting from them to complete.

In the second case, I can't see how it's ever going to be possible to check external links until the docs are published. I'm never going to build the entire website locally.

Please can we revert to including external links as coded? @nik9000

dedemorton:
This is a relatively recent problem. I had an older version of the docs repo on my system (from about a week ago), and did not see this behavior, even with asciidoctor builds.

If this is expected behavior, would it be possible to add a flag to the build command so that we can have a local build with working external links? I actually use these links quite a lot.

Titch990:
In the short term, when testing links, I can continue to work almost as normal by replacing `localhost/8000` in the broken link with `https://elastic.co`.

Titch990:
I think that if the link is into a document that has been included in the build, it should be a localhost link, so we follow the link into the doc we built. Whereas if the link is to a document that isn't included in the local build, it should be to elastic.co, so we can be sure we haven't broken a link that used to work.

So for a build of all the docs, all links into docs pages should be local, as we've built them all, but links to the website for example, should be too elastic.co.

nik9000:
[...]

PR #1307, opened 2019-10-17
Redirect slightly more

nik9000:
When you make a request to a location outside of the `guide` we should
redirect to `www.elastic.co` *unless*:
1. We have the file (then we should just send it to you).
2. You made the request for air gapped docs (we should 404).

This implements that by adding a `try_files` block to the `/` arm of our
nginx config which forwards requests for missing files right into the
preview application. The preview application checks if you are asking
for the air gapped deploy or not and performs the redirect or not.

Relates to #1300


nik9000:

PR #1308, opened 2019-10-17
Only relativize links build with --all

nik9000:
I don't particularly like building different books with `--all` vs
`--doc`, but in this case it seems like it is the best solution....

We want to relativize links to `www.elastic.co` because it prevent
linking off into "nowhere" when you use the docs in a air gapped
network. And it makes it less likely that you'll accidentally get linked
to production when browsing the docs preview. On the other hand, when
you are building docs with `--doc` the URL space looks nothing like when
you build all the docs so links into the docs that aren't already part
of the book aren't going to resolve properly if we relativize them.

So, this does what I don't really want to do but think is probably the
right thing to do anyway: it only relativizes the links if you build
all of the books and skips it if you build with `--doc`.

Relates to #1300

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -115,12 +115,14 @@ sub build_local {
         build_single( $index, $raw_dir, $dir, %$Opts,
                 latest       => $latest,
                 alternatives => \@alternatives,
+                relativize => 0,
         );
     }
     else {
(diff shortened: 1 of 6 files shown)


nik9000:

PR #1312, opened 2019-10-18
[DOCS] Split Infrastructure Monitoring guide into separate Logs and Metrics docs.

Titch990:
Changes required as a result of splitting Infrastructure Monitoring Guide into separate Logs and Metrics guides.

This is part of a multi-repo change (see elastic/observability-dev#98). DO NOT PUSH UNTIL EVERYTHING IS READY.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -1214,7 +1214,7 @@ contents:
                 repo:   docs
                 path:   shared/attributes.asciidoc
     -
-        title:      Observability: APM, Infrastructure, Logs, and Uptime
+        title:      Observability: APM, Logs, Metrics, and Uptime
         sections:
           -
             title:      Application Performance Monitoring (APM)
--- a/conf.yaml
+++ b/conf.yaml
@@ -1678,7 +1678,7 @@ contents:
             title:      Infrastructure Monitoring Guide for 6.5-7.4
             prefix:     en/infrastructure/guide
             current:    7.4
-            branches:   [ 7.x, 7.5, 7.4, 7.3, 7.2, 7.1, 7.0, 6.8, 6.7, 6.6, 6.5 ]
+            branches:   [ 7.4, 7.3, 7.2, 7.1, 7.0, 6.8, 6.7, 6.6, 6.5 ]
             index:      docs/en/infraops/index.asciidoc
             chunk:      1
             tags:       Infrastructure/Guide
--- a/conf.yaml
+++ b/conf.yaml
@@ -1369,7 +1369,27 @@ contents:
                         repo:   apm-agent-rum-js
                         path:   docs
           -
-            title:      Infrastructure Monitoring Guide
+            title:      Logs Monitoring Guide
+            prefix:     en/logs/guide
+            current:    7.4
+            branches:   [ master, 7.x, 7.4, 7.3, 7.2, 7.1, 7.0, 6.8, 6.7, 6.6, 6.5 ]
+            index:      docs/en/logs/index.asciidoc
+            chunk:      1
+            tags:       Logs/Guide
+            subject:    Logs
+            asciidoctor: true
+            sources:
+              -
+                repo:   stack-docs
+                path:   docs/en
+              -
+                repo:   docs
+                path:   shared/versions/stack/{version}.asciidoc
+              -
+                repo:   docs
+                path:   shared/attributes.asciidoc
+          -
+            title:      Metrics Monitoring Guide
             prefix:     en/infrastructure/guide
             current:    7.4
             branches:   [ master, 7.x, 7.5, 7.4, 7.3, 7.2, 7.1, 7.0, 6.8, 6.7, 6.6, 6.5 ]
--- a/conf.yaml
+++ b/conf.yaml
@@ -1383,7 +1383,7 @@ contents:
             title:      Logs Monitoring Guide
             prefix:     en/logs/guide
             current:    master
-            branches:   [ master ]
+            branches:   [ master, 7.x, 7.5 ]
             index:      docs/en/logs/index.asciidoc
             chunk:      1
             tags:       Logs/Guide
@@ -1403,7 +1403,7 @@ contents:
             title:      Metrics Monitoring Guide
             prefix:     en/metrics/guide
             current:    master
-            branches:   [ master ]
+            branches:   [ master, 7.x, 7.5 ]
             index:      docs/en/metrics/index.asciidoc
             chunk:      1
             tags:       Metrics/Guide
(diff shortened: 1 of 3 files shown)


Titch990:

Titch990:
Note that this is one of four linked PRs, which need to be merged together to avoid build failures.
elastic/stack-docs#581
elastic/docs#1312 (this one)
elastic/kibana#48633
https://github.com/elastic/beats/pull/14158

DO NOT MERGE UNTIL EVERYTHING IS READY

dedemorton:
Can you please assign @zuketo as the reviewer here since it pertains to info architecture changes?

shared/attributes.asciidoc
```diff
@@ -60,6 +59,10 @@
 :curator-ref:          https://www.elastic.co/guide/en/elasticsearch/client/curator/{branch}
 :curator-ref-current:  https://www.elastic.co/guide/en/elasticsearch/client/curator/current
 :infra-guide:          https://www.elastic.co/guide/en/infrastructure/guide/{branch}
+:metrics-ref:          https://www.elastic.co/guide/en/metrics/{branch}
+:metrics-guide:        https://www.elastic.co/guide/en/metrics/guide/{branch}
+:logs-ref:             https://www.elastic.co/guide/en/logs/{branch}
+:logs-guide:           https://www.elastic.co/guide/en/logs/guide/{branch}
```
bmorelli25: Any reason why you want to include these changes in this PR? If you break these changes into a separate PR it would allow you to merge the Kibana and Beats changes prior to splitting the book. That would make the process much simpler than trying to merge 4 PRs at the same time.

conf.yaml
```diff
@@ -1600,6 +1620,26 @@ contents:
     -
         title:      Legacy Documentation
         sections:
+          -
+            title:      Infrastructure Monitoring Guide for 6.5-7.4
+            prefix:     en/infrastructure/guide
+            current:    7.4
+            branches:   [ 7.x, 7.5, 7.4, 7.3, 7.2, 7.1, 7.0, 6.8, 6.7, 6.6, 6.5 ]
```
bmorelli25: Do we still want to build `7.x` and `7.5` here? `7.4` will be the last current so might make sense not to build those other branches.

If we do want to build them, you should update the title in L1624

conf.yaml
```diff
@@ -1363,14 +1363,34 @@ contents:
...
           -
-            title:      Infrastructure Monitoring Guide
-            prefix:     en/infrastructure/guide
-            current:    7.4
-            branches:   [ master, 7.x, 7.5, 7.4, 7.3, 7.2, 7.1, 7.0, 6.8, 6.7, 6.6, 6.5 ]
-            index:      docs/en/infraops/index.asciidoc
+            title:      Logs Monitoring Guide
+            prefix:     en/logs/guide
+            current:    master
+            branches:   [ master ]
```
bmorelli25: Should we also build `7.5` here? Then update that to `current` when `7.5` releases?

conf.yaml
```diff
@@ -1363,14 +1363,34 @@ contents:
...
+                repo:   docs
+                path:   shared/versions/stack/{version}.asciidoc
+              -
+                repo:   docs
+                path:   shared/attributes.asciidoc
+          -
+            title:      Metrics Monitoring Guide
+            prefix:     en/metrics/guide
+            current:    master
+            branches:   [ master ]
```
bmorelli25: Should we also build `7.5` here? Then update that to `current` when `7.5` releases?

shared/attributes.asciidoc
```diff
@@ -60,6 +59,10 @@
 :curator-ref:          https://www.elastic.co/guide/en/elasticsearch/client/curator/{branch}
 :curator-ref-current:  https://www.elastic.co/guide/en/elasticsearch/client/curator/current
 :infra-guide:          https://www.elastic.co/guide/en/infrastructure/guide/{branch}
```
bmorelli25: I wonder if we should create a new "LEGACY" section in this file and move this attribute to it. That would help prevent someone from accidentally trying to use this attribute in the future.

conf.yaml
debadair: Continuing to build 7.x and 7.5 would be really confusing, I think. 

When we've refactored docs in the past, folks have had a hard time figuring out how to get to the latest docs. I think you should customize the banner message to point to the new docs. (Not everyone will read it, but it will help some folks.)

conf.yaml
Titch990: Yes, but that will break the build if I do that before the changes have been pushed to the 7.5 branch, because there are no 7.5 docs to build yet. Unless you know differently, I think I have to update them one at a time.

shared/attributes.asciidoc
Titch990: Because that's the way it came out. These definitions are needed to resolve broken links else where in the docs. I'm really reluctant to spend any more time unpicking and reworking this any further this unless there is a clear benefit. As far as I can tell, as implemented, this builds across all the affected repos. Any more fiddling is probably going to break things.

conf.yaml
Titch990: I've had to leave 7.5 and 7.x for now, because otherwise the 7.5 and 7.x builds break. I can't remove these until I've merged the related changes in Kibana, beats and elsewhere in stack-docs into the 7.5 and 7.x branches to refer to the new docs instead of this one. I'll move the 7.5 and 7.x builds to the new docs as soon as there is nothing in the 75 and 7.x branches that refers to the old docs.

conf.yaml
bmorelli25: Cool beans

shared/attributes.asciidoc
bmorelli25: Sounds good

shared/attributes.asciidoc
Titch990: That's not a bad idea, although if someone tries to use this attribute, the build will break anyway, as the file doesn't exist any more. But I'm all for clarity up front.

conf.yaml
Titch990: Comment as below

bmorelli25: 👍

conf.yaml
```diff
@@ -1363,14 +1363,34 @@ contents:
                         repo:   apm-agent-rum-js
                         path:   docs
           -
```
bmorelli25: @Titch990 How about changing L1215:
```
        title:      Observability: APM, Infrastructure, Logs, and Uptime
```

To this:
```
        title:      Observability: APM, Logs, Metrics, and Uptime
```

Based on this image:
01-top-level

Titch990: Fair comment. I have a separate task to change Infrastructure to Metrics throughout, so I wasn't too bothered about finding them all yet, but it does make sense to pick this one off now.

conf.yaml
Titch990: @debadair Thanks for the clarification (via Slack). I've created https://github.com/elastic/stack-docs/issues/645 to address this separately so that I can progress this mega-merge without any more hold-ups.

nik9000:
I've dropped the change to `attributes.asciidoc` and we can do it in a follow up. It was causing the PR build to take a long, long time and we'd like to get a clean build of this ASAP.

PR #1313, opened 2019-10-18
Switch all attribute links to https

nik9000:
We should prefer https for the security and because our link relativizer
only does the job for https.


nik9000:
Thanks for reviewing @Titch990 !

PR #1321, opened 2019-10-18
Switch the Korean ES reference to Asciidoctor

nik9000:
AsciiDoc is unmaintained and we need to drop support for it.


nik9000:
The diff is *tiny* and quite ok. Merging.

PR #1322, opened 2019-10-18
Switch uptime docs to Asciidoctor

nik9000:
AsciiDoc is unmaintained and we should drop support for it.


nik9000:
@bmorelli25 it looks like the alias for the uptime docs uses asciidoctor but `conf.yaml` doesn't.

bmorelli25:
doh! Thanks for catching this.

nik9000:
The diff looks pretty empty too.

PR #1323, opened 2019-10-18
Create dedicated livenss endpoint

nik9000:
Adds a `/liveness` endpoint which returns a 200 strait from nginx so it
will be considered "alive" very, very soon.


nik9000:
It's Alive!

PR #1325, opened 2019-10-18
Make doc repo dependencies explicit in Beats documentation

dedemorton:
Adds explicit paths for shared attributes and versions used by the Beats documentation.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -804,6 +804,14 @@ contents:
               -
                 repo:   beats
                 path:   CHANGELOG.asciidoc
+              -
+                repo:   docs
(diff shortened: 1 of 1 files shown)


dedemorton:
>  There is also an attributes62.asciidoc file that you may or may not be using in the older versions to look out for.

Does that matter though? I wouldn't expect anyone to be updating that file at this point.

nik9000:
> Does that matter though? I wouldn't expect anyone to be updating that file at this point.

I changed it earlier today to switch all the links to https. Honestly though, I don't think folks will change it much. It is kind of nice to have it form a "central place to see all the sources" standpoint but it will rarely trigger builds.

PR #1332, opened 2019-10-22
Move the air gapped docs

nik9000:
It turns out we do want them publicly accessible.


nik9000:
In fact we want to keep the doc where they are!

PR #1335, opened 2019-10-22
Switch cloud's heroku docs to asciidoctor

nik9000:
AsciiDoc is unmaintained and we need to drop support for it.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -645,6 +645,7 @@ contents:
             chunk:      1
             noindex:    1
             private:    1
+            asciidoctor: true
             sources:
               -
                 repo:   cloud


nik9000:
@nrichers I'll use this PR to generate a list of errors, and eventually a preview and a diff for the coming migration.

nrichers:
jenkins test this please

nik9000:
The diff is fairly tame, but here it is:

* [ ] [Kerbeose](https://www.elastic.co/guide/en/cloud-heroku/release-ms-28/ech-secure-clusters-kerberos.html) has  `{s} supports Kerberos with Elastic Stack version 7.3 and later.`. It wasn't in the output before because `{s}` wasn't defined. But Asciidoctor renders the names of attributes that aren't defined.
* [ ] A note in [OpenID](ech-secure-clusters-oidc.html) is rendering as `+ NOTE:`.
* [ ] The numbering on a list in [OpenID](ech-secure-clusters-oidc.html) is reset around `Update Kibana in the`
* [ ] There is an errant `+` in in [OpenID](ech-secure-clusters-oidc.html) before `This configuration disables all other` and the list has a few more `+`es below that.
* [ ] [Saml](ech-securing-clusters-SAML.html) has `{s} supports SAML with Elastic Stack version 6.4 and later.`
* [ ] There is an extra `+` before "Custom bundles are unzipped under the path" in [Saml](ech-securing-clusters-SAML.html).
* [ ] It looks like the list resets itself a few times in [Saml](ech-securing-clusters-SAML.html) as well.
* [ ] The [index](https://www.elastic.co/guide/en/cloud-heroku/release-ms-28/index.html) contains

PR #1337, opened 2019-10-23
Drop support for building with AsciiDoc

nik9000:
We are now building entirely with Asciidoctor. This doesn't remove the
AsciiDoc dependency entirely, but we'll get there soon. It also doensn't
remove all of the asciidoctor vs asciidoc plumbing, but we'll get there
in a follow up.

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -62,6 +62,13 @@ our $Conf = LoadFile(pick_conf());
 # passing this argument.
 die 'build_docs.pl is unsupported. Use build_docs instead' unless $Opts->{in_standard_docker};
 
+if ( $Opts->{asciidoctor} ) {
+    say <<MSG
+The Asciidoctor migration is complete! --asciidoctor will emit this message
+forever in honor of our success but otherwises doesn't do anything.
+MSG
+}
+
 init_env();
 
 $Opts->{doc}           ? build_local()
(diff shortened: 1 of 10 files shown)


nik9000:
We're not quite ready for this one yet but I figured I'd start the process now.

build_docs.pl
```diff
@@ -62,6 +62,13 @@ BEGIN
 # passing this argument.
 die 'build_docs.pl is unsupported. Use build_docs instead' unless $Opts->{in_standard_docker};
 
+if ( $Opts->{asciidoctor} ) {
+    say <<MSG
+The Asciidoctor migration is complete! --asciidoctor will emit this message
+forever in honor of our success but otherwises doesn't do anything.
```
olksdr:
```suggestion
forever in honor of our success but otherwise doesn't do anything.
```

integtest/spec/helper/source.rb
```diff
@@ -71,7 +71,7 @@ def repo_with_index(name, index_content)
   end
 
   ##
-  # Create a repo with an index.asciidoctor file and a book that uses it as
+  # Create a repo with an index.asciidoct file and a book that uses it as
```
olksdr:
```suggestion
  # Create a repo with an index.asciidoc file and a book that uses it as
```

nik9000:

PR #1347, opened 2019-10-23
Drop unused python program

nik9000:
Drops a small python program that we used to use for serving the docs
"locally" before we had docker and could use nginx.


nik9000:
:zap:

PR #1348, opened 2019-10-23
Finish a ruby migration

nik9000:
A while back I started migrating my calls to "convert" documents in the
integration tests from a single method into a fluent builder style
because we've grown so many parameters for the conversion. This finishes
that migration.


nik9000:
I've added @estolfo as a reviewer because she is the best at ruby! But I've also added @olksdr and @JoshMock because this is *fairly* straightforward.

nik9000:
Thanks @JoshMock!

PR #1349, opened 2019-10-23
Simplify open_in_widget

nik9000:
I've been wanting to remove our fairly complex `TreeProcessorScaffold`
for a while now because it mirror's Asciidoctor's `Conveter` except it
doesn't properly handle attributes. This migrates `open_in_widget` off
of the scaffold to a converter. This also has the advantage of giving a
clear path to implement the converter if we target html instead of
docbook, which we'd very much like to do.

### DIFF
--- a/resources/asciidoctor/.rubocop.yml
+++ b/resources/asciidoctor/.rubocop.yml
@@ -4,11 +4,9 @@ inherit_from: ../../.rubocop.yml
 
 Metrics/AbcSize:
   Exclude:
-    - lib/open_in_widget/extension.rb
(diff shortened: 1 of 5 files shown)


nik9000:
I'm hoping this change is also *fairly* simple to review. I think most of the change is actually adding a level of indentation.

nik9000:

I'd like to know if anything actually changes. I don't *think* anything will.

nik9000:
> I'd love to talk this functionality out in real time with you sometime soon, @nik9000, so I can understand it better.

Happy to do it!

PR #1355, opened 2019-10-24
Add new attributes for the upcoming guides

nik9000:
We're splitting the logging and metrics guides. Let's get the attributes
in first so we can build everything else a little more cleanly.


nik9000:

PR #1358, opened 2019-10-25
Upgrade asciidoctor

nik9000:
We need to pick up the fix for
https://github.com/asciidoctor/asciidoctor/issues/3089 so we need to
upgrade. These are the changes required to upgrade:

* Add our own docbook 4.5 converter because Asciidoctor 2 dropped
support for docbook 4.5. We'll drop support for docbook entirely in a
follow up so there is no point in upgrading to docbook 5.
* Rework converter methods to be `convert_{name}` instead of `{name}`.
* Replace `with_format` with `format`.
* `node.attr` no longer delegates missing attributes to the document.
You now be explict if you want an attribute from the document and do
`node.document.attr`.
* Asciidoctor no longer `WARN`s when it sees an invalid reference. It
logs as INFO. We convert those back to warnings
* Asciidoctor 2 seems to log some bad links incorrectly. We patch it to
flip a conditional and have sent the patch upstream.


resources/asciidoctor/lib/docbook45/converter.rb
```diff
@@ -0,0 +1,171 @@
+# encoding: UTF-8
```
nik9000: This file comes from asciidoctor's source. Well, mostly. I had to do quite a bit of funky things to it to make it work properly in Asciidoctor 2, but I don't believe it'll want to stick to our style or other guidelines. Mostly because we plan to remove it in favor of converting directly to html soon enough anyway.

nik9000:
This is probably worth a look from @estolfo, though the big file is mostly copied from Asciidoctor and the rest is fairly forced.

nik9000:
If CI passes then I'd like to rebuild all the docs and compare the result. *hopefully* this is a noop.

nik9000:

nik9000:
> @elasticmachine, run elasticsearch-ci/docs rebuild

Woah! The build worked! Wonderful!

nik9000:
@estolfo, could you look at this one?

.docker/asciidoctor_2_0_10.patch
```diff
@@ -0,0 +1,14 @@
+Tracked at https://github.com/asciidoctor/asciidoctor/issues/3469
```
nik9000: I'm sure this sort of thing isn't normal but it seems nicer than trying to work around the bug in another way. Hopefully they'll just take the patch.

nik9000:
Thanks for the reviewing @olksdr and @estolfo!

PR #1361, opened 2019-10-27
Add go-elasticsearch to the list of alternative languages

karmi:
This patch adds the Go to the list of alternative languages to choose from in the documentation.


karmi:

PR #1367, opened 2019-10-29
[DOCS] Update custom macro examples for Asciidoctor

jrodewig:
Now that most of our docs are produced using Asciidoctor, we can update the custom macro examples included in the elastic/docs README.

This changes the `macro[]` convention to `macro:[]` for inline macros and `macro::[]` for block-style macros.

As the old convention still works, this may not be worth auditing in our existing docs. The intent is to adopt the new conventions going forward.


nik9000:
This seems like the right thing to do but there is a test that might fail on this because AsciiDoc doesn't produce the same output as Asciidoctor. We'll drop the test once we switch the last document using AsciiDoc (ECE). That *should* come in a week or two, if all things go well.

jrodewig:
Thanks @nik9000. Makes sense. I'll throw the WIP label on this PR until after the ECE migration. I'll re-ping y'all for reviews then too.

jrodewig:
@elasticmachine run self-test

nik9000:
@jrodewig I think we're ready for this now, but you'll have to merge master into it first or the self-test will be upset.

jrodewig:
Thanks @nik9000. That did the trick. Ready for review by you or @lcawl at your convenience.

PR #1368, opened 2019-10-29
Don't allow tables to escape

### LINKED ISSUE #1237: Tables overlap table of contents
Table 85 wraps poorly when I'm not viewing the following page in a full-sized window:
https://www.elastic.co/guide/en/elasticsearch/reference/master/transform-search.html

image

Is this something we can address?

nik9000:
This prevent super-wide tables from escaping their containers and ending
up under the nav.

Closes #1237


nik9000:

PR #1369, opened 2019-10-29
Fix bug in preprocessor

nik9000:
Something about upgrading to asciidoctor 2 broke our custom preprocessor
when it sees `include::foo.adoc[leveloffset=+1]`, specifically the
`leveloffset` part. We didn't have a test for it because we didn't think
it was a big deal, but it came up.

Rubocop made this change larger than it had to be by insisting that I
break up `postprocesses`. I think the result is more readable, but the
diff could have been one line without the required refactoring.


nik9000:
Could you take a look at this one, @estolfo?

nik9000:
:rocket:

Issue #1196, opened 2019-09-18
Document best practices for content reuse


dedemorton:
Content reuse is great, but builds dependencies between projects and makes it more difficult for external contributors to update docs.

Before we go hog wild with tagged regions, let's come up with some best practices around content reuse and identify strategies for creating shared content that's easier to consume.

dedemorton:
I'm going to dump some raw info here because I'm seeing stuff during the asciidoctor migration that we should avoid.

- [ ] Avoid using tagged regions to include content that contains ifdefs and ifevals.

benskelker:
Don't know if this make any sense at all, and it would be loads of work for Nik even if it does, so just throwing out an idea for using unique IDs:

1. Tag content you want to reuse in the source file and open a PR
2. The elasticsearch-ci/docs build creates a unique ID for the tagged content and displays it somewhere in the console output 
3. Use the ID in the target file
4. Resolve content reuse links during doc builds

What happens when the target file is merged but the PR for the source content is not? Only allow target file merges after the source file is merged?

Loads of disadvantages (file walk to resolve links and the need to dynamically add the link somewhere in the transform process) and may not even be doable, but it would give us robust content reuse and the ability to easily see and track all sources and targets.

debadair:
It would be really helpful to have some authoring tooling that enabled you to introspect the source files and see all of the places where topics and tagged regions are included. And, conversely, easily open included topics to make modifications. It wouldn't really help with the contributor problem, but it would  make it a little easier for Elastic folks to understand the implications of their changes.

debadair:
WRT unique IDs, I think the way Antora handled this was that everything was scoped based on IDs set in the playbooks. I'd worry about having to author using generated IDs instead of being able to discover them from just looking at the source.

dedemorton:
A group of us have been discussing approaches to content reuse. @lcawl created a PR that shows what it might look like to house shared content about monitoring in the docs repo: 
https://github.com/elastic/docs/pull/1370

After reviewing this PR, I have some concerns about chopping the content into a lot of small pieces and housing it outside of the main content repos. It just doesn't feel right to me, especially given the limitations of our current authoring environment. I worry that reuse at this level of granularity:

- Makes the content difficult to update, especially for external contributors, because it removes the content from the main flow. 
- Makes the content prone to getting stale over time. Because the content would live outside the project repo, developers are likely to overlook it when they make changes.
[...]

PR #1370, opened 2019-10-29
Adds shared monitoring content

lcawl:
This PR adds files that contain tagged regions related to configuring monitoring in Kibana and Beats.

These regions can then be used in multiple books from diverse repos. 

For example, this information currently appears in  https://www.elastic.co/guide/en/kibana/master/monitoring-metricbeat.html, https://www.elastic.co/guide/en/beats/metricbeat/master/monitoring-metricbeat-collection.html, and https://www.elastic.co/guide/en/elastic-stack-overview/master/esms.html

This file format is copied from how common API reference parameters are re-used in https://github.com/elastic/elasticsearch/blob/master/docs/reference/rest-api/common-parms.asciidoc

Ideally we can switch to using these shared files, as in https://github.com/elastic/elasticsearch/pull/48594
NOTE: The inclusion of the branch in the path for these files means that we can include the appropriate version of the files using attributes. For example:

> include::{docs-root}/shared/monitoring/{source_branch}/kibana-config.asciidoc[tag=disable-kibana-collection]

### DIFF
--- a/shared/monitoring/master/beats-config.asciidoc
+++ b/shared/monitoring/master/beats-config.asciidoc
@@ -1,0 +1,117 @@
+//The following tagged sections pertain to configuring monitoring in Beats
+
+// Enable the HTTP endpoint
+tag::enable-http-endpoint[]
+Add the following setting in the {beatname_uc} configuration file
+(+{beatname_lc}.yml+):
+
+[source,yaml]
+----------------------------------
+http.enabled: true
+----------------------------------
+
+By default, metrics are exposed on port 5066. If you need to monitor multiple
+{beats} shippers running on the same server, set `http.port` to expose metrics
+for each shipper on a different port number:
+
+[source,yaml]
+----------------------------------
+http.port: 5067
+----------------------------------
+end::enable-http-endpoint[]
+
+// Enable the `beat-xpack` module in Metricbeat
+tag::enable-beat-module[]
+For example, to enable the default configuration in the `modules.d` directory, 
+run the following command, using the correct command syntax for your OS:
+
+["source","sh",subs="attributes,callouts"]
+----------------------------------------------------------------------
+metricbeat modules enable beat-xpack
+----------------------------------------------------------------------
+
+For more information, see 
+{metricbeat-ref}/configuration-metricbeat.html[Specify which modules to run] and 
+{metricbeat-ref}/metricbeat-module-beat.html[beat module]. 
+end::enable-beat-module[]
+
+// Disable the system module in Metricbeat:
+tag::disable-system-module[]
+By default, the {metricbeat-ref}/metricbeat-module-system.html[system module] is
+enabled. The information it collects, however, is not shown on the
+*Stack Monitoring* page in {kib}. Unless you want to use that information for
+other purposes, run the following command:
+
+["source","sh",subs="attributes,callouts"]
+----------------------------------------------------------------------
+metricbeat modules disable system
+----------------------------------------------------------------------
+end::disable-system-module[]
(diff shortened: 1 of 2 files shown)


shared/monitoring/master/beats-config.asciidoc
```diff
@@ -0,0 +1,117 @@
...
+Set the `hosts`, `username`, and `password` settings as required by your
+environment. For other module settings, it's recommended that you accept the
+defaults.
+
+By default, the module collects {beatname_uc} monitoring data from
+`localhost:5066`. If you exposed the metrics on a different host or port when
+you enabled the HTTP endpoint, update the `hosts` setting.
+
+To monitor multiple 
+ifndef::apm-server[]
```
nik9000: I think we might be better off standardizing on an attribute for the name of the thing we're documenting.

bmorelli25: Unfortunately, "agent" means two completely different things in the APM and Beats world. But instead of changing this ifndef, it's probably less messy to just get rid of it. We could either add another attribute for `[Beats agents] | [APM Server instances]`, or simplify to something like:
```
To monitor multiple instances, specify a list of hosts. For example:
```
or
```
To monitor multiple {beats} agents or APM Server instances, specify a list of hosts. For example:
```

dedemorton:
@lcawl Thank you for putting this together. It's a good test case to get us talking about specific examples. After looking at this example, though, I already have some concerns about chopping the content into a lot of small pieces and housing it outside of the main content repos. This level of reuse could get messy very quickly. I've added my comments to the discussion in [#1196](https://github.com/elastic/docs/issues/1196#issuecomment-547669879).

It might help for me to get some draft guidelines down on paper before we start moving content around.

debadair:
I feel pretty strongly that we should not put content in the docs infra repo. Having shared resources like the attribute definitions makes sense because they aren't tied to particular products or versions. 
(Though in the past we had to do some pretty hacky things to get around the fact that even those shared files aren't versioned using the standard version branch scheme.)

I would rather see us make the stack-docs repo a built-in dependency for all books, put shared content there, and rely on the standard versioning scheme. With the CI builds, I'm less uncomfortable about adding a dependency for people building locally. (Because I think fewer people bother.)

That doesn't address @dedemorton's concerns about the complexity that granular re-use introduces. Truthfully, even the shared ES reference descriptions make me a bit twitchy. There are obvious benefits to consistency across the ref pages, but without having any way to see where the descriptions are used, I think there's a risk of introducing subtle inaccuracies. (In ES, just because a request or response element has the same name, it doesn't necessarily mean that it's identical--context matters.)  And that's with content that is well-suited for reuse--definitions that don't change much over time.

I'm concerned about reusing parts of procedures, because that indicates to me that we're telling people how to do things in possibly slightly different ways in different places. To a large degree, this is a reflection of the complexity involved in turning all the right knobs, but it feels like it increases the risk of someone landing on what *looks* like the right set of steps, and then getting confused (or doing the wrong thing) because they aren't actually in the right place. 

I think it's no surprise that monitoring is the guinea pig for this, because the monitoring docs have been problematic for a long time, there's been a *lot* of change, and lots of "band-aids" have been applied. 

I don't have a better solution to offer, but I think we need to take a hard look at whether or not ensuring consistency across these topics is improving the user or authoring experience in a meaningful way.

benskelker:
@lcawl 
Thanks for adding me as a reviewer. Sorry, don't have anything to add as we're still stuck on the same issues. I'll add general thoughts to https://github.com/elastic/docs/issues/1196.

PR #1373, opened 2019-10-31
Fix preview's root redirect

nik9000:
The preview was redirecting to port `8000` because that is the port
nginx runs on. But k8s serves the actual result out of port `80`! While
we *could* turn off setting the port in nginx, it feels simpler to send
a relative redirect. Every modern browser supports this and the preview
app does it all the time. So this delegates the redirect for `/` to the
preview application.


nik9000:
Thanks for reviews!

PR #1374, opened 2019-10-31
Clean up leftover

nik9000:
Drops sending the Dockerfile into the docker build process. It isn't
required.


nik9000:
Thanks for reviewing, @olksdr!

PR #1375, opened 2019-10-31
Switch to docker's buildkit everywhere

nik9000:
I hadn't used it when I first migrated the docs build to docker because
it was slower when the image was already cached. But it is now a little
faster for me. And it is the future of the docker project. And it gives
more features which we will use eventually.


nik9000:
:fist_raised: :robot: :policeman:

PR #1376, opened 2019-10-31
More docker cleanup

nik9000:
* Don't version the docs:build image. It is really just a dependency
resolution mechanism. Let's just always call it `latest`.
* Reduce the number of places we have to copy and paste the tags for the
preview image.
* Bump the version of the preview image so we can release the next
version.
* Move the `publish_docker.sh` file to `.ci/publish_preview.sh` and make
it only publish the docs preview image.


nik9000:

PR #1378, opened 2019-10-31
Shrink docker images

nik9000:
This *massively* shrinks our docker images, most importantly, the air
gapped docs image:
```
-rw-------. 1 manybubbles manybubbles 1.1G Oct 31 17:16 /home/manybubbles/Downloads/air_gapped_after.tar
-rw-------. 1 manybubbles manybubbles 1.6G Oct 31 17:09 /home/manybubbles/Downloads/air_gapped_before.tar
```

It does so by using the "multi-stage builds" feature from docker's
buildkit. It allows you to create multiple docker images in the same
Dockerfile and copy parts each image into another. We use that to build
images that *just* install dependencies and then we copy those
dependencies from those images into a "final" image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,98 +1,99 @@
+# This builds a "base" image that contains dependencies that are required at
+# runtime *and* to install dependencies like `ruby`. Then it builds images that
+# intall those dependencies. Then it builds a "final" image that copies the
+# dependencies and installs package that it needs at runtime. Compared to
+# installing everything in one go this shrinks the "final" docker image by
+# about 50%, mostly because we don't need things like `gcc` and `ruby-dev` in
+# the "final" image.
+
 # Debian builds the docs about 20% faster than alpine. The image is larger
 # and takes longer to build but that is worth it.
-FROM bitnami/minideb:buster
+FROM bitnami/minideb:buster AS base
 
-LABEL MAINTAINERS="Nik Everett <nik@elastic.co>"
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,98 +1,99 @@
...
+COPY --from=py_deps /usr/local/lib/python3.7/dist-packages /usr/local/lib/python3.7/dist-packages
+COPY --from=py_deps /usr/local/bin/pycodestyle /usr/local/bin/pycodestyle
+COPY --from=ruby_deps /var/lib/gems /var/lib/gems
+COPY --from=ruby_deps /usr/local/bin/asciidoctor /usr/local/bin/asciidoctor
+COPY --from=ruby_deps /usr/local/bin/rspec /usr/local/bin/rspec
+COPY --from=ruby_deps /usr/local/bin/rubocop /usr/local/bin/rubocop
+
+# We mount these directories with tmpfs so we can write to them so they
+# have to be empty. So we delete them.
+RUN rm -rf /var/log/nginx && rm -rf /run/nginx
```
olksdr: you could shave few more MB off by running `apt-get clean` to remove downloaded packages from cache.

nik9000: I believe `install_package` does that. Which is actually a problem because we run it a few times so it runs `apt-get update` and clears that cache over and over again.

nik9000:

PR #1380, opened 2019-11-01
Fix a NO SUCH FILE bug

nik9000:
One that I don't see locally, but other folks do. I'm not sure that this
fixes it but it is worth a shot!


nik9000:
Oh cool! I had a good guess, I guess....

PR #1381, opened 2019-11-01
Drop vendored modules

nik9000:
Moves resolution of perl dependencies from files stored in the repo to
packages installed in the docker container. It doens't feel worth it to
keep our own copy when we have a perfectly good dependency resolution
mechanism just waiting for us.


nik9000:
This entirely drops the `URL` perl library because we don't actually use it any more. Same for `File::Which`.

nik9000:
Yay smaller checkout!

PR #1382, opened 2019-11-05
Prevent informal table from escaping

nik9000:
It turn out that tables that don't have a title don't have a
`table-contents` class. They have an `informaltable` class intead.
:shrug: This prevents those sorts of table from escaping under the nav
as well.


nik9000:
:heart:

PR #1384, opened 2019-11-05
Drop vendored copy of AsciiDoc

nik9000:
We don't use it any more.


nik9000:
Looks like I tried to drop too much!

nik9000:
Looks like we couldn't drop *all* of asciidoc. We still need bits of it for docbook to work!

nik9000:
Zap!

PR #1385, opened 2019-11-05
Clean up some AsciiDoc leftovers

nik9000:
Cleans up some bits of the perl that are "funny" now that we don't have
to support AsciiDoc.


nik9000:
Zoom!

PR #1387, opened 2019-11-05
Track direct-to-html conversions

nik9000:
Replaces the mechanism that we used track the Asciidoctor migration with
a new mechanism that we will use to track the "direct to html"
migration.


nik9000:
I'll add some test changes to this one soon.

nik9000:
> I'll add some test changes to this one soon.

Done!

PR #1388, opened 2019-11-06
Switch ECS ref from version to current

DonNateR:
Switch ECS ref attribute from version to current. When using the `{ecs_version)` variable, every ECS link in the SIEM and Kibana SIEM sections default to v1.1.


nik9000:
We try to avoid linking to `current` just because urls for that can jump around some. So I think the version attribute is normally right.

nik9000:
The idea is to update `ecs_version` on release, in the files that have it and are live. He idea is that old versions of books can keep pointing to the same version of ecs so all the links all make sense.

DonNateR:
Got it. Closing.

PR #1390, opened 2019-11-07
Emulate docbook chunking

nik9000:
We'd very much like to avoid having to use docbook to generate our HTML
because it is slow and difficult to customize. This creates the
beginnings of an Asciidoctor plugin that emulates docbook's chunking.
The plan is to enable it on the perl side with a command line flag and a
flag in conf.yaml in a follow up change.

Relates to #1390.

### DIFF
--- a/resources/asciidoctor/spec/chunker_spec.rb
+++ b/resources/asciidoctor/spec/chunker_spec.rb
@@ -1,0 +1,131 @@
+# frozen_string_literal: true
+
+require 'chunker/extension'
+require 'fileutils'
+require 'tmpdir'
+
+RSpec.describe Chunker do
+  before(:each) do
+    Asciidoctor::Extensions.register Chunker
+  end
+
+  after(:each) do
+    Asciidoctor::Extensions.unregister_all
+  end
+
+  include_context 'convert without logs'
+  let(:backend) { :html5 }
+
+  context 'when outdir is configured' do
+    let(:outdir) { Dir.mktmpdir }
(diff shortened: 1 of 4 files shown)


nik9000:

nik9000:
@JoshMock and @olksdr, given the amount of ruby that'll probably be involved in the "generate html directly" project it'd probably be good to get a bit more familiar with it. It won't be a *ton*, but it'll be a fair bit.

resources/asciidoctor/spec/chunker_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+  after(:each) do
+    Asciidoctor::Extensions.unregister_all
+  end
+
+  include_context 'convert without logs'
+  let(:backend) { :html5 }
+
+  context 'when outdir is configured' do
+    let(:outdir) { Dir.mktmpdir }
+    after(:example) { FileUtils.remove_entry outdir }
```
tylerjl: This looks totally fine, but just to share something I've used in the past: the `around` hook can bring up/teardown temp files in one block to consolidate where that's handled to avoid any weirdness if lazy `let` variables don't get evaluated or something else in between the `let` and `after`. [Example](https://gist.github.com/sue445/d840d6f68771a14a48fc)

nik9000: I do that sort of thing in the integration tests because I want to be super careful about building things one time. But I try to go traditional in the unit tests because it is still quite fast and makes ruby folks feel a little more at home.

Issue #743, opened 2019-03-21
Build documentation directly to HTML; eliminate DocBook step


lcawl:
After our migration to Asciidoctor is complete, there are features we cannot use due to the fact that we still convert to DocBook then HTML. The features don't seem to generate errors or warnings, they just don't work. For example:

- https://github.com/elastic/docs/issues/726
- ~https://github.com/elastic/docs/issues/896~
- https://github.com/elastic/docs/issues/1279

Per https://github.com/elastic/docs/issues/249#issuecomment-343424353 "Docbook output is required as the asciidoctor html output doesn't support html chunking, plus docbook gives us strict link checking. It would still be a big speedup. That said, I don't know about the difference in output and error checking behaviour. Plus the customizations need to be ported over. This is a huge undertaking..."

This issue is intended to collect a list of the features that don't work in the current implementation and ideas about how we could eliminate the need for Docbook.

nik9000:
I'm going to pick this up soon!

nik9000:
I think we've actually finish this. I removed docbook from the docs repo entirely in January.

PR #1391, opened 2019-11-08
Add the option to skip docbook for single page

nik9000:
This adds the option to generate html directly from asciidoctor for
single page books with `--direct_html`. Right now no books use this but
we'd like to migrate all of them too it once we're sure the output looks
good for them!

Relates to #743

### DIFF
--- a/lib/ES/Util.pm
+++ b/lib/ES/Util.pm
@@ -54,6 +54,7 @@ sub build_chunked {
     my $branch = $opts{branch};
     my $roots = $opts{roots};
     my $relativize = $opts{relativize};
+    my $direct_html = $opts{direct_html};
 
     die "Can't find index [$index]" unless -f $index;
 
@@ -191,6 +192,7 @@ sub build_single {
     my $branch = $opts{branch};
     my $roots = $opts{roots};
     my $relativize = $opts{relativize};
+    my $direct_html = $opts{direct_html};
 
     die "Can't find index [$index]" unless -f $index;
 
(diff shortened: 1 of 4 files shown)


lib/ES/Util.pm
```diff
@@ -206,27 +208,7 @@ sub build_single {
         $raw_dest->mkpath;
     }
 
-    my %xsltopts = (
```
nik9000: Moved next to the xslt invocation.

lib/ES/Util.pm
```diff
@@ -206,27 +208,7 @@ sub build_single {
...
-            'navig.graphics'           => 1,
-            'admon.textlabel'          => 0,
-            'admon.graphics'           => 1,
-    );
-    if ( $type eq 'book' ) {
-        $xsltopts{'chunk.section.depth'} = 0;
-    }
-
     my ( $output, $died );
-    my $dest_xml = $index->basename;
```
nik9000: Moved into the branch that deals with the xml.

lib/ES/Util.pm
```diff
@@ -267,25 +254,47 @@ sub build_single {
     } or do { $output = $@; $died = 1; };
     _check_build_error( $output, $died, $lenient );
 
-    if ( !$lenient ) {
+    unless ( $direct_html ) {
```
nik9000: Everything in here is either moved from above or the same as it was before, just indented one more time.

template/template.js
```diff
@@ -52,7 +52,6 @@ module.exports = templateSource => {
            * at the end of the chunk in case the marker is on the edge. */
           const slice = Math.max(0, chunk.length - preserve);
           chunk = chunk.slice(slice) + result.value;
-          // TODO check if slice + keeps a copy of internal memory. This implementation assumes it *doesn't*
```
nik9000: I checked this while debugging the failure below.

template/template.js
```diff
@@ -189,7 +193,8 @@ module.exports = templateSource => {
         const out = apply(raw[Symbol.asyncIterator](), lang, initialJsState);
         write.on("close", resolve);
         write.on("error", reject);
-        out.on("error", write.destroy);
+        // out.on("error", write.destroy) doesn't properly forward the error!
+        out.on("error", err => write.destroy(err));
```
nik9000: I have no idea why this is required. Maybe `destroy` is altered once you start the pipe or something.

PR #1393, opened 2019-11-09
Fix a test bug in template

nik9000:
The template tests were missing a `return` which meant that the
assertion didn't trigger. That assertion started failing and we didn't
know! This fixes the `return` and the failing assertion.


nik9000:
> Promises are weird. +1

So weird!

PR #1396, opened 2019-11-12
Default to built-docs repo

nik9000:
`--target_repo` has been required with `--all` for a while now, but we
only ever really call it with
`--target_repo git@github.com:elastic/built-docs.git`. We typically use
`--target_branch` for experimentation. So let's just assume `built-docs`
unless someone passes `--target_repo`.


nik9000:
:+1:

PR #1397, opened 2019-11-12
Invert Makefile control

nik9000:
Previously we ran make inside of our docker container. Long ago this was
important because we did complex things in make. We don't any more
though because that was a stygian nightmare. This flips the `Makefile`s
so *they* run docker. This is nice because:
1. It is now possible to `cd` into any directory you like and run `make`
to run the tests.
2. We can use different docker images for different tests. We *don't* do
that right now, but we certainly *can* do that with this change in the
future.

This routes all execution of docker through the `build_docs` script,
which is nice because docker requires a bunch of command line options to
properly run the scripts, particularly in the integration tests. It'd
be nice if we could limit those options to just the tests that need
them, but that is a problem for another change.

Routing docker execution through `build_docs` also makes sure the docker
containers are always up to date, which is nice too.

### DIFF
--- a/resources/web/Makefile
+++ b/resources/web/Makefile
@@ -1,22 +1,13 @@
-# We expect this to be run in the a docker container managed by
-#   build_docs --self-test
-
-SHELL = /bin/bash -eux -o pipefail
-MAKEFLAGS += --silent
+include ../../common.mk
 
 .PHONY: check
 check: jest
 
-.PHONY: test_watch
-test_watch: tests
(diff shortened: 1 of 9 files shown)


resources/web/Makefile
```diff
@@ -1,22 +1,13 @@
...
-MAKEFLAGS += --silent
+include ../../common.mk
 
 .PHONY: check
 check: jest
 
-.PHONY: test_watch
-test_watch: tests
-	/node_modules/parcel/bin/cli.js watch ./docs_js/__tests__/**/* ./docs_js/__tests__/* -d ./tests/ --target node --no-source-maps &
-	/node_modules/jest/bin/jest.js ./tests/** --watch
```
nik9000: I don't believe this has worked in a while to be honest.

PR #1398, opened 2019-11-12
Speed up building "normal" docker image

nik9000:
Our "normal" docker image was including context files needed for the
`preview` docker image because I didn't know how to make separate
`.dockerignore` files for each `Dockerfile`. I know now. So this fixes
up the main `.dockerignore` file.


nik9000:
:zap:

PR #1400, opened 2019-11-12
Clean old preview branches

nik9000:
Cleans branches that were committed more than 31 days ago. This should
keep the `built-docs` repo from growing out of control somewhat.


nik9000:
clean

nik9000:
Grumble grumble, markdown no like me.

PR #1401, opened 2019-11-12
Add processor doc folders to the conf.yaml

dedemorton:
Adds processor doc files so we can move them to a directory within the code modules.

~Currently only covers Packetbeat. I will copy to other beats after @nik9000 confirms that this approach will work.~ done

~This PR must be merged after https://github.com/elastic/beats/pull/14488~ done

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -871,6 +871,10 @@ contents:
                 repo:   beats
                 path:   libbeat/docs
               -
+                repo:   beats
+                path:   libbeat/processors/*/docs/*.asciidoc
+                exclude_branches:   [ 7.x, 7.5, 7.4, 7.3, 7.2, 7.1, 7.0, 6.8, 6.7, 6.6, 6.5, 6.4, 6.3, 6.2, 6.1, 6.0, 5.6, 5.5, 5.4, 5.3, 5.2, 5.1, 5.0, 1.3, 1.2, 1.1, 1.0.1 ]
+              -
                 repo:   docs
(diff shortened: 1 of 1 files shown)


nik9000:
I'd avoid *.asciidoc and just let the directory do it. I *think* it should work though.

nik9000:
The build failed because the files don't exist yet.

dedemorton:
@nik9000 I thought the build needed the full path to trigger rebuilds when a file changes?

https://github.com/elastic/beats/pull/14488 should fix the broken CI.

dedemorton:

dedemorton:
jenkins test this again

dedemorton:
@nik9000 It looks like the CI passed even with the splats. If it looks OK to you, let's merge this puppy so I can get the processor doc refactoring merged soonish.

conf.yaml
```diff
@@ -870,6 +870,10 @@ contents:
               -
                 repo:   beats
                 path:   libbeat/docs
+              -
+                repo:   beats
+                path:   libbeat/processors/*/docs/*.asciidoc
```
nik9000: Could you drop the `*.asciidoc` bits here?

Nothing else has them and they'll be surprising if we ever want to include an image in these directories or something like that.

dedemorton:
@nik9000 I've removed *.asciidoc. I thought a full path was required to trigger doc rebuilds when source files change, but it sounds like the full path to the dir is enough? I'll merge as soon as the CI is green. Thanks!

PR #1403, opened 2019-11-13
Shrink production images

nik9000:
This shrinks the build, preview, and air gapped images by moving the
test dependencies out of them. This is also nice because changing test
dependencies won't cause us to have to rebuild the production images.
The cost is that it'll be a little slower to run the tests the first
time. This isn't a problem locally, but will slow down CI. We can
mostly mitigate that in a follow up change though.

Before:
```
-rw------- 1 manybubbles manybubbles 1.1G Nov 12 19:06 air_gapped_with_test.tar
```

After:
```
-rw------- 1 manybubbles manybubbles 930M Nov 12 19:01 air_gapped_no_test.tar
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,25 +11,17 @@
 FROM bitnami/minideb:buster AS base
 
 # TODO install_packages calls apt-get update and then nukes the list files after. We should avoid multiple calls to apt-get update.....
+# We could probably fix this by running the update and installs ourself with `RUN --mount type=cache` but that is "experimental"
 
-# Setup the repo for node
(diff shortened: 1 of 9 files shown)


nik9000:
That build failure is because `parcel` is listed as a `devDependency` which, it sort of is and sort of isn't. It isn't needed for the preview or the air gapped docs but it is is needed to build the docs proper. So the space savings isn't going to be as great as advertised. I haven't figured out how much the difference is though.

nik9000:
I don't feel *quite* so good about it with parcel in there, but I think this is the right way to get it out of there eventually. So I'm going to commit it.

PR #1404, opened 2019-11-13
Progress building chunked docs

nik9000:
This hooks the `chunker` extension added in #1390 to the docs build when
you do a multi-page book with `--direct_html`. Important: it crashes!
But I'll fix it with mostly ruby hacking and this is more perl stuff and
it is fairly selef contained. So I figure I should fix it in a follow up
change.

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -123,7 +123,6 @@ sub build_local {
         );
     }
     else {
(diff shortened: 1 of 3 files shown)


nik9000:
Progress!

PR #1412, opened 2019-11-15
Switch beats paths to "pathspec"s

nik9000:
These work properly with git to find the files we need. Hopefully.


nik9000:
@dedemorton I *think* these should work. I've added tests for this sort of thing and they look right. But, as I said, computers are sneaky.

PR #1413, opened 2019-11-15
Fix fairly strange error

nik9000:
Fixes a fairly esoteric error that you get when you do
```
./build_docs --keep_hash --sub_dir blah:blah:blah
```
without `--push`.

It fixes it by *always* passing the git username and email to the docker
image all the time. It also cleans up the tests a little bit in the ways
that that change implies.


nik9000:

PR #1415, opened 2019-11-15
Switch headings to Inter

nik9000:
We don't depend on the aktiv-grotesk font any more but we still declare
that headings should have it. That causes them to look radically
different on different platforms. We should just use Inter like we do
with anything else.

### DIFF
--- a/resources/web/style/heading.pcss
+++ b/resources/web/style/heading.pcss
@@ -2,7 +2,7 @@
   h1, h2, h3, h4, h5, h6 {
     /* Relative position allows laying out the edit me links inside the header. */
     position: relative;
-    font-family: aktiv-grotesk, sans-serif;
+    font-family: Inter;
     /* Override the "funny" line height from bootstrap. */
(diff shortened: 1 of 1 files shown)


nik9000:
> I couldn't find any other references to aktiv-grotesk in this repo, so I wonder if we could drop it from TypeKit? Unless it's being loaded in from the marketing site code and is used elsewhere outside this repo, of course.

@brianjolly can talk to that. I *believe* they've already dropped it actually.

JoshMock:
I vaguely recall it being on our list of fonts we were trying to drop before we could air-gap, and https://github.com/elastic/infra/issues/13669 corroborates, so I think you're right. :+1:

brianjolly:
Yes, on the marketing side we're still using Aktiv for headings so it shows up in both the nav and footer. Design was fine with docs using something different though to avoid font license issues.

nik9000:
Thanks @brianjolly and @JoshMock !

PR #1416, opened 2019-11-15
Keep the diff tool working

nik9000:
This adds a docker image for the diff tool. We used to be able to use
the regular build image for it but now that we stripped all the test
stuff out of it we'll need a named image for it.

Expect like, real tests and stuff for it to come in at some point. It is
an "expert" tool mostly used for migration, but it is important that it
works.


nik9000:
Thanks @JoshMock !

PR #1417, opened 2019-11-15
Speed up tests

nik9000:
Builds the testing docker images during the packer cached step of the
process that builds the images that run the docs build. So *usually*
we'll be able to use the cached copy. Which speeds up the tests.

### DIFF
--- a/build_docs
+++ b/build_docs
@@ -40,7 +40,6 @@ import webbrowser
 
 # The tag we use for our docker image. This should line up with the tags
 # in preview/Dockerfile and publish_docker.sh
-DOCKER_TAG = 'docker.elastic.co/docs/build:latest'
 DOCKER_BUILD_QUIET_TIME = 3  # seconds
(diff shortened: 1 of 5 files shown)


build_docs
```diff
@@ -628,10 +630,16 @@ if __name__ == '__main__':
             cmd.extend(argv[3:])
             returncode = subprocess.call(cmd)
             exit(returncode)
+        if '--docker-build' == argv[1]:
+            if len(argv) != 3:
+                raise ArgError(
+                    'Usage: build_docs --build-image <image>')
+            image = argv[2]
+            build_docker_image(image)
+            exit(0)
```
JoshMock: I do still wonder if something like [`click`](https://click.palletsprojects.com/en/7.x/) would tighten up a lot of this logic for you.

nik9000:
At this point I think it might! I'd have to vendor it in though because I
can't rely on folks having anything installed for this script.

nik9000:
Thanks for reviewing @olksdr and @JoshMock!

PR #1420, opened 2019-11-18
Create docbook workalike plugin

nik9000:
Creates a plugin to "work like" Elastic's docbook customizations. The
idea here is to replace docbook with as close to 0 diffs as possible.
But in some cases, that isn't worth it. So this also alters `html_diff`
to start to normalize all of the diffs that we don't care about. Right
now there is only one, but I'm sure we'll have a hand full.

The actual asciidoctor plugin munges the header and title into the
correct form with string manipulation. I don't feel *too* bad about that
because:
1. It is in the grand tradition of bubble gum and duct tape that holds
all of the important bits of the internet together.
2. We'll be able to *remove* it piece by piece once we're done. It'll be
*much* easier to visually verify that removing each piece is safe if we
can do it one by one.
3. Further work on the plugin will *probably* be cleaner because the
other parts of the document are generated much more precisely so we
*should* be able to override more carefully.

### DIFF
--- a/resources/asciidoctor/lib/docbook_compat/extension.rb
+++ b/resources/asciidoctor/lib/docbook_compat/extension.rb
@@ -1,0 +1,85 @@
+# frozen_string_literal: true
+
+require 'asciidoctor/extensions'
+require_relative '../delegating_converter'
+
+##
+# HTML5 converter that emulates Elastic's docbook generated html.
+module DocbookCompat
+  def self.activate(registry)
+    return unless registry.document.basebackend? 'html'
+
+    DelegatingConverter.setup(registry.document) { |d| Converter.new d }
+  end
+
+  ##
+  # A Converter implementation that emulates Elastic's docbook generated html.
+  class Converter < DelegatingConverter
+    def initialize(delegate)
+      super(delegate)
+    end
+
+    def convert_document(doc)
+      html = yield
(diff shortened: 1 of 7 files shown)


resources/asciidoctor/lib/docbook_compat/extension.rb
```diff
@@ -0,0 +1,85 @@
...
+      html.gsub!('</title>', "</title>\n#{meta}") ||
+        raise("Couldn't add dc meta to #{html}")
+    end
+
+    def munge_body(doc, html)
+      wrapped_body = <<~HTML.strip
+        <body>
+        <div class="#{doc.doctype}" lang="#{doc.attr 'lang', 'en'}">
+      HTML
+      html.gsub!(/<body[^>]+>/, wrapped_body) ||
```
JoshMock: Is there ever a case where there's going to be a `<body>` tag with no attributes? If so, `/<body[^>]*>/` might be a better choice.

nik9000: I don't *believe* so, but yeah.

nik9000:
> Parsing HTML with regex is brave, and I applaud your efforts.

It is! I feel ok about it because this is coming from asciidoct *and* we'll probably zap it as time goes on anyway.

Issue #1419, opened 2019-11-16
Sunset EOL versions of the docs


debadair:
It's helpful for folks to be able to reference docs for older versions, but we want to steer people who are new to the stack to the latest versions and encourage people who are running EOL versions to upgrade.

To this end, we should:

- [x] Omit EOL versions from the version drop-down. 
- [x] Update the banner on EOL versions to highlight that it is EOL, encourage upgrades, and direct folks to documentation for the latest version.
- [x] Add noindex tags to EOL versions. 
- [x] Submit a reindex request to Google. 

We should start with the 1.x series and monitor the SEO and traffic impacts as we sunset and remove each version. (Suddenly dropping thousands of pages from Google's search index could have unexpected/undesirable consequences.)

At some future date, we might opt to stop publishing the oldest versions entirely. These are basically the prerequisite steps to doing that.

Why keep the oldest versions around at all?
Historical interest. To preserve information about upgrading for folks running really old versions.

How does this encourage people to upgrade?
It lends some urgency to making a change if users see us phasing out documentation for older versions. Updating the banners will reinforce when versions reach EOL. 

How does this help steer users to the latest version?
Not having those versions show up in search results should boost the signal of the more recent versions a bit. Making it clearer when you've landed on an EOL version should also help people realize there's more-recent content. 

Depends on: https://github.com/elastic/docs/issues/1324, https://github.com/elastic/docs/issues/1431, https://github.com/elastic/docs/pull/1426

lcawl:
I had thought at the very least we should stop updating and building the out of support versions (i.e. leave a static copy of that content on the web). If we can get away with removing them entirely, however, that's even better.

debadair:
Having fewer duplicate pages has benefits from an SEO perspective, and reduces the likelihood of unsuspecting users landing on pages that don't apply to the version they're using. 

As part of the migration effort, @nik9000 raised a concern about publishing pages that we can't update. If we simply freeze them and stop routinely building them, we are likely to break the ability to build them without knowing it. If we're still publishing them, that leaves us exposed if some issue crops up that affects those pages.

zuketo:
I completely agree with:

> 
- Discontinue publishing EOL versions of the docs except for the last minor in the major version series. Today, for ES that would mean publishing: 0.90, 1.7, 2.4, 5.6, 6.3, 6.4, 6.5, 6.6, 6.7, 6.8, 7.0, 7.1, 7.2, 7.3, 7.4, 7.5, 7.x, master.
[...]

PR #1421, opened 2019-11-18
Create a "live" branches config for books

nik9000:
We've wanted for a while to mark old versions of books as "noindex" and,
probably, to remove the `edit_me` links from them. This creates a config
entry for `conf.yaml` to mark branches as "live". If this list is
specified then everything *not* in the list will be marked as `noindex`.
In the future we might remove the edit links from it was well.

This applies that config change to the installation and upgrade guide,
just to apply it to something for testing. I think we'll want to apply
it to most books eventually.

### DIFF
--- a/integtest/spec/helper/book_conf.rb
+++ b/integtest/spec/helper/book_conf.rb
@@ -1,0 +1,83 @@
+# frozen_string_literal: true
+
+##
+# Methods to build the config for a book.
+module BookConf
+  ##
+  # The configuration needed to build the book.
+  def conf
+    # We can't use to_yaml here because it emits yaml 1.2 but the docs build
+    # only supports 1.0.....
+    yaml = standard_conf
+    yaml += variable_conf
+    yaml += <<~YAML
+      sources:
+      #{sources_conf}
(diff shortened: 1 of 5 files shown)


integtest/spec/helper/book_conf.rb
```diff
@@ -0,0 +1,83 @@
...
+  def map_branches_conf(map_branches)
+    return '' unless map_branches
+
+    yaml = "map_branches:\n"
+    map_branches.each_pair do |key, value|
+      yaml += "  #{key}: #{value}\n"
+    end
+    yaml
+  end
+end
```
olksdr: you build here a lot of yaml by hand. Maybe it's worth pulling `yaml` lib for this and use ruby hashes/lists to build the structure you want and then just `to_yaml` it?

nik9000: I'm going to investigate that again.... The last time I tried it didn't work because the yaml library emitted yaml that perl couldn't read!

nik9000:
Relates to #1419.

PR #1422, opened 2019-11-19
Clean up yaml generation in tests

nik9000:
The integration tests used a lot of code to properly generate yaml that
the docs build could parse, mostly because the docs build used an old
yaml parser that had a few bugs. But we recently upgraded the yaml
parser and it looks like it doesn't have the bugs any more! So this
replaces the "by hand" yaml generation with ruby's built in `.to_yaml`
which is much simpler to do.

### DIFF
--- a/integtest/spec/all_books_sources_spec.rb
+++ b/integtest/spec/all_books_sources_spec.rb
@@ -71,7 +71,7 @@ RSpec.describe 'building all books' do
           extra_branches.each { |b| repo.switch_to_new_branch b }
           docs_repo = init_docs_repo src
           book = src.book 'Test'
-          book.branches << extra_branches
+          book.branches += extra_branches
           book.source repo, 'index.asciidoc'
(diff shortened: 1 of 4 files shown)


integtest/spec/all_books_sources_spec.rb
```diff
@@ -71,7 +71,7 @@ def self.init_docs_repo(src)
           extra_branches.each { |b| repo.switch_to_new_branch b }
           docs_repo = init_docs_repo src
           book = src.book 'Test'
-          book.branches << extra_branches
+          book.branches += extra_branches
```
nik9000: It looks like the old way was tolerant of the "funny" array this used to build: `[master, [5.5, 6.3, 7.2]]` but the new way isn't. So it caught a bug!

nik9000:
> Nice. So this looks much better now. What do you think?

I think it is better! I never liked generating the yaml by hand bug felt stuck doing it.....

nik9000:
Yay YAML!

PR #1423, opened 2019-11-19
Add "live" branches for APM agents

bmorelli25:
Full details in https://github.com/elastic/docs/pull/1421 & https://github.com/elastic/docs/issues/1419. To summarize:

> We've wanted for a while to mark old versions of books as "noindex" and,
probably, to remove the `edit_me` links from them. This creates a config
entry for conf.yaml to mark branches as "live".

This PR sets _live_ branches for the APM agents. Specifically, it marks only `master` and the `current` version of each agent as _live_. @lreuven (or maybe @axw?) can you confirm there are no other branches that should also be marked as _live_?

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -1292,6 +1292,7 @@ contents:
                     prefix:     go
                     current:    1.x
                     branches:   [ master, 1.x, 0.5 ]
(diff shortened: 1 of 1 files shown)


axw:
I think those are the right branches.

PR #1424, opened 2019-11-19
Configure "live" branches for stack books

nik9000:
Configures which branches are "live" for book that are versioned with
the stack.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -50,7 +50,7 @@ contents:
             current:    7.4
             index:      docs/en/install-upgrade/index.asciidoc
             branches:   [ master, 7.x, 7.5, 7.4, 7.3, 7.2, 7.1, 7.0, 6.8, 6.7, 6.6, 6.5, 6.4, 6.3, 6.2, 6.1, 6.0, 5.6, 5.5, 5.4, 5.3, 5.2, 5.1, 5.0 ]
-            live:       [ master, 7.x, 7.5, 7.4, 6.8 ]
+            live:       &stacklive [ master, 7.x, 7.5, 7.4, 6.8 ]
             chunk:      1
             tags:       Elastic Stack/Installation and Upgrade
             subject:    Elastic Stack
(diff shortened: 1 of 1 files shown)


nik9000:

conf.yaml
```diff
@@ -95,6 +95,7 @@ contents:
             current:    7.4
             index:      docs/en/getting-started/index.asciidoc
             branches:   [ master, 7.x, 7.5, 7.4, 7.3, 7.2, 7.1, 7.0, 6.8, 6.7, 6.6, 6.5, 6.4, 6.3 ]
+            live:       *stacklive
```
lcawl: Does this setting mean that this book uses the same values as are specified in &stacklive for the Installation and Upgrade Guide?  If so, that's a lot easier than copying the same values over and over.

conf.yaml
```diff
@@ -1253,6 +1274,7 @@ contents:
                 index:      docs/guide/index.asciidoc
                 current:    7.4
                 branches:   [ master, 7.5, 7.4, 7.3, 7.2, 7.1, 7.0, 6.8, 6.7, 6.6, 6.5, 6.4, 6.3, 6.2, 6.1, 6.0 ]
+                live:       *stacklive
```
bmorelli25: APM doesn't build `7.x` documentation. Will using `*stacklive` here (which lists `7.x` as a live branch) cause a problem?

conf.yaml
```diff
@@ -1269,6 +1291,7 @@ contents:
                 index:      docs/index.asciidoc
                 current:    7.4
                 branches:   [ master, 7.5, 7.4, 7.3, 7.2, 7.1, 7.0, 6.8, 6.7, 6.6, 6.5, 6.4, 6.3, 6.2, 6.1, 6.0 ]
+                live:       *stacklive
```
bmorelli25: Same question as above

conf.yaml
nik9000: It *ought* to be fine if the `live` list is "bigger" than the branches list.

conf.yaml
nik9000: It does.

PR #1425, opened 2019-11-19
Improve direct html generation

nik9000:
This improves our plugin for direct html generation in Asciidoctor to
the point where the diff for the perl client's docs is empty. Now, that
isn't as big a feat as it sounds because:
1. Most of the perl client's docs are on CPAM so the docs we actually
build are pretty small.
2. I pull a few tricks with `html_diff` to ignore a few changes that
don't *look* like they make any visual difference to me.

Even with those two caveates, this is pretty good progress. It switches
the perl client to direct html, because, well, I did all the work to
make sure the diff is empty.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -542,6 +542,7 @@ contents:
                 single:     1
                 tags:       Clients/Perl
                 subject:    Clients
+                direct_html: true
                 sources:
                   -
                     repo:   elasticsearch-perl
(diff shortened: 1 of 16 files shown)


nik9000:
> Looks sane to me. But handling/parsing html by hand scares me cold_sweat

I know! I'm pretty sure we'll be able to remove most of these after the migration which is the only reason I've done it this way. I'd really hoped that I could be more precise. All of the `doc_munging.rb` is quite brittle to the way asciidoctor generates its html.

I do have html parsing in `html_diff` which I'm using when I'm *super* sure the diff doesn't matter.

PR #1430, opened 2019-11-20
Improve docbook chunking emulation

nik9000:
Improves the docbook chunking emulation by generating proper headers for
the pages. This allows the templating system to work for them so
now multi-page books at least won't crash with `--direct_html`.


nik9000:
OK! I can start looking at all the books now!

PR #1432, opened 2019-11-20
Convert all single page books to direct html

nik9000:
This replaces the xmllint/xslt steps with the ruby implementation we've
been working on. It is generally much faster and much easier to
customize.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -320,6 +320,7 @@ contents:
             single:     1
             tags:       Elasticsearch/Resiliency Status
             subject:    Elasticsearch
+            direct_html: true
             sources:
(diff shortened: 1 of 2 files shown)


nik9000:
I don't expect the build to succeed, but it'll give me something to iterate on. If it does succeed I'll likely spend some time working to shrink the diffs. The upshot is that it'll be a while before I merge this.

nik9000:
#1436 and #1438 get pretty close to getting these working. I believe there is just juggling the table of contents left after that.

nik9000:
The diffs are pretty much empty! Time to merge!

Issue #1338, opened 2019-10-23
Docs homepage


zuketo:
Improvements to the docs homepage: https://www.elastic.co/guide/index.html

Initial notes and landing page outline from @debadair are [here](https://docs.google.com/document/d/1HZkPoZ8oZfidhXDMTWwV9gKG9D_83uBseYrUNJDMwIg/edit#heading=h.4klx2u1752i0)

@Adrian-J's [latest design](https://marvelapp.com/9e91008/screen/62324268)

Let's use this issue for tracking both the homepage changes and any new CSS styles needed for the homepage changes

Adrian-J:
The latest design for the homepage can be found [here](https://marvelapp.com/9e91008/screen/62324268).

cc for visibility: @zuketo @debadair @gchaps @nik9000 @nrichers @dedemorton

Adrian-J:
Updating this thread for visibility. The prototype is [here](https://docs-74e930.webflow.io/).

dedemorton:
Can we update the observability page in this design to reflect the navigation changes that @bmorelli25 and I made in our outline?  It would be nice to see how the nav looks at a high level.

bmorelli25:
@dedemorton I shared the updated nav with Adrian earlier today. He's going to update it for us.

PR #1434, opened 2019-11-21
Plumbing for "extra" HTML in TOCs

nik9000:
This adds the perl plumbing for "extra" HTML that is added to the table
of contents after it is rendered. This *probably* won't be enough
for #1338 but it is a good start.

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -56,7 +56,8 @@ use ES::LinkCheck();
 GetOptions($Opts, @{ command_line_opts() }) || exit usage();
 check_opts();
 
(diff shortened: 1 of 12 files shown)


nik9000:
Also CCing @DonNateR who'll pick up this process after this is in.

nik9000:
This makes invisible changes to https://www.elastic.co/guide/en/kibana/index.html and https://www.elastic.co/guide/index.html just so it is clear where we'll put the "extra" HTML.

nik9000:
Zoom!

PR #1435, opened 2019-11-21
Added the alternative examples for Cloud

philkra:
This PR adds the alternative examples to the cloud docs. Currently only the SaaS book is currently subjected, as the display is still in PoC.

/cc @nik9000 & @russcam

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -677,7 +677,7 @@ contents:
                 alternatives: { source_lang: csharp, alternative_lang: php }
                 repo:   clients-team
                 path:   examples/elastic-cloud/php
-                private: true
+                map_branches: *mapCloudSaasToClientsTeam
               -
                 alternatives: { source_lang: csharp, alternative_lang: go }
                 repo:   clients-team
@@ -727,6 +727,12 @@ contents:
                 repo:   clients-team
                 path:   examples/elastic-cloud/php
                 map_branches: &mapCloudEceToClientsTeam
+                  2.4: master
+                  2.3: master
+                  2.2: master
+                  2.1: master
+                  2.0: master
+                  1.1: master
                   1.0: master
               -
                 alternatives: { source_lang: csharp, alternative_lang: go }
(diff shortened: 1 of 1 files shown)


philkra:
From the `cloud` repo as root directory as well as the `clients-team` and `docs` repo checked ou, I'm running the following command to build the docs: 

```sh
../docs/build_docs --asciidoctor --open --doc docs/saas/index.asciidoc --resource docs/shared --chunk 1 --alternatives csharp:php:../clients-team/examples/elastic-cloud/php --conf ../docs/conf.yaml --verbose
```

Unfortunately the docs build does not pick up the alternative example (for only PHP in this case), I don't see a log message to indicate the missing link.

nik9000:
The failure is:
```
20:00:02 INFO:build_docs:Error executing: git archive --format=tar -o /tmp/docsbuild/y6uMX0Jf_L/clients-team/.temp_git_archive.tar release-ms-28 examples/elastic-cloud/php in GIT_DIR /docs_build/.repos/clients-team.git
```

It is because there isn't a `release-ms-28` in clients-team. I believe I can fix the problem. I'll push something in a moment.

nrichers:
Quick postscript: After you figure out how to address https://github.com/elastic/docs/pull/1435#issuecomment-557039831, we probably also need to update `doc_build_aliases.sh` for ESS, ECH (Heroku), and ECE.

nik9000:
> Quick postscript: After you figure out how to address [#1435 (comment)](https://github.com/elastic/docs/pull/1435#issuecomment-557039831), we probably also need to update `doc_build_aliases.sh` for ESS, ECH (Heroku), and ECE.

We may not want to update the normal alias as these can add a fair bit of time. I think it'd be worth adding a "with alternatives" version.

nik9000:
@philkra and @nrichers, I've got the build going. Basically I pointed everything to the `master` branch of your repo.

nrichers:
@nik9000 this is great news, thank you so much for your help! I learned something today, as I wasn't aware that you could map branches like this.

philkra:
Thank you so, so much @nik9000 🥇

PR #1438, opened 2019-11-21
Create a substitute for docinfo in direct html

nik9000:
We'd very much like to retire docbook, but we use its "docinfo" feature
quite a bit to add extra text to the header of the title page of a few
of our books. Asciidoctor *has* a "docinfo" feature but we can't use it
to emulate docbook's docinfo because the asciidoctor adds the text to
the `<head>` tag while docbook adds it below the `<h1>` tag. This change
causes us to look for files names `-extra-title-page.html` to find text
to add just below the `<h1>`.

### DIFF
--- a/integtest/spec/helper/dest.rb
+++ b/integtest/spec/helper/dest.rb
@@ -197,6 +197,16 @@ class Dest
       self
     end
 
+    def direct_html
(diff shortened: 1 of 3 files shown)


nik9000:
Let's go!

PR #1439, opened 2019-11-22
Add libbeat/outputs dirs to conf.yaml

dedemorton:
Adds path statements to get asciidoc source files from directories under libbeat/outputs/docs.

Prep work for more refactoring of libbeat.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -882,6 +882,10 @@ contents:
                 path:   libbeat/processors/*/docs/*
                 exclude_branches:   [ 7.4, 7.3, 7.2, 7.1, 7.0, 6.8, 6.7, 6.6, 6.5, 6.4, 6.3, 6.2, 6.1, 6.0, 5.6, 5.5, 5.4, 5.3, 5.2, 5.1, 5.0, 1.3, 1.2, 1.1, 1.0.1 ]
               -
+                repo:   beats
+                path:   libbeat/outputs/*/docs/*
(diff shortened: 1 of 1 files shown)


conf.yaml
```diff
@@ -881,6 +881,10 @@ contents:
                 repo:   beats
                 path:   libbeat/processors/*/docs/*
                 exclude_branches:   [ 7.4, 7.3, 7.2, 7.1, 7.0, 6.8, 6.7, 6.6, 6.5, 6.4, 6.3, 6.2, 6.1, 6.0, 5.6, 5.5, 5.4, 5.3, 5.2, 5.1, 5.0, 1.3, 1.2, 1.1, 1.0.1 ]
+              -
+                repo:   beats
+                path:   libbeat/outputs/*/docs/*
+                exclude_branches:   [ 7.x, 7.5, 7.4, 7.3, 7.2, 7.1, 7.0, 6.8, 6.7, 6.6, 6.5, 6.4, 6.3, 6.2, 6.1, 6.0, 5.6, 5.5, 5.4, 5.3, 5.2, 5.1, 5.0, 1.3, 1.2, 1.1, 1.0.1 ]
```
nik9000: I've started to use yaml anchors to make this kind of thing a little less copy-and-paste. Here is an example:
```
            live:       &stacklive [ master, 7.x, 7.5, 7.4, 6.8 ]
```

And I can reuse it like so:

```
            live:       *stacklive
```

You certainly don't have to do this, but I think it'd make things a little clearer if you did. Though maybe it'd make more sense in a follow up change. I'm not sure.

nik9000: I'm trying to go with `camelCase` for naming these, mostly because the yaml parser we use likes those a little better. So I should have named this `stackLive` .

dedemorton: Yah this is a cleanup step because the branch list for exclude_branches isn't complete yet. I'll need to update it again after I merge then backport changes.

PR #1441, opened 2019-11-22
Handle single-level TOCs in direct html

nik9000:
This modifies the docbook_compat plugin to add the table of contents to
the same location that docbook adds it and to generate it in the same
style that docbook uses. It only works for a single level, but that
should be fine for now. We'll improve it in further work.

### DIFF
--- a/resources/asciidoctor/lib/docbook_compat/convert_document.rb
+++ b/resources/asciidoctor/lib/docbook_compat/convert_document.rb
@@ -6,12 +6,17 @@ module DocbookCompat
   # scary but required at this point for docbook compatibility.
   module ConvertDocument
     def convert_document(doc)
+      # We'll manually add the toc ourselves if it was requested.
+      wants_toc = doc.attr?('toc') && doc.attr?('toc-placement', 'auto')
+      doc.attributes.delete 'toc' if wants_toc
+
       html = yield
       html.gsub!(/<html lang="[^"]+">/, '<html>') ||
         raise("Coudn't fix html in #{html}")
       munge_head doc, html
       munge_body doc, html
       munge_title doc, html
+      add_toc doc, html if wants_toc
       html
     end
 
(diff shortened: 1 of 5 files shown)


resources/asciidoctor/lib/docbook_compat/convert_document.rb
```diff
@@ -58,13 +63,26 @@ def munge_body(doc, html)
 
     def munge_title(doc, html)
       id = doc.id || 'id-1'
-      header_start = <<~HTML.strip
-        <div class="titlepage"><div><div>
+      # Important: we're not replacing he whole header - it still will have a
```
nik9000: I started by trying to munge the toc out of where it was in the header and realized that a lot more thing can end up in the header. I ended up not trying to get the toc out with string manipulation but I kept this part of the change so we could handle other things in the header without errors.

resources/asciidoctor/lib/docbook_compat/convert_lists.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module DocbookCompat
```
nik9000: Rubocop has started warning me that I'm bumping into classes that are "too big" so I'm starting to move groups of conversions to modules that I am mixing in to the converter. The class is still big and the tests are still big but at least the implementations are in "themed" files.

resources/asciidoctor/lib/docbook_compat/extension.rb
```diff
@@ -19,18 +21,12 @@ def self.activate(registry)
...
     include ConvertDocument
     include ConvertLinks
+    include ConvertLists
+    include ConvertOutline
 
-    def initialize(delegate)
-      super(delegate)
-    end
-
-    SECTION_WRAPPER_CLASSES = %w[unused chapter section].freeze
```
nik9000: I moved `wrapper_class` to a method because I use it in two places.

resources/asciidoctor/lib/docbook_compat/extension.rb
```diff
@@ -41,7 +37,7 @@ def convert_section(node)
 
     def convert_floating_title(node)
       tag_name = %(h#{node.level + 1})
-      # Asciidoctor's standard is to pu the id on the header tag but docbook
+      # Asciidoctor's standard is to put the id on the header tag but docbook
```
nik9000: Speeling.

resources/asciidoctor/lib/docbook_compat/extension.rb
```diff
@@ -66,19 +62,6 @@ def convert_listing(node)
       HTML
     end
 
-    def convert_ulist(node)
```
nik9000: Moved into the "themed" file to make rubocop happy.

nik9000:
Once this is in I can merge #1432!

nik9000:
> changes look sane to me :shipit:

Define sane. Some of the munging isn't all the sane if I planned to keep it forever!

resources/asciidoctor/lib/docbook_compat/convert_document.rb
olksdr:
```suggestion
      # Important: we're not replacing the whole header - it still will have a
```

PR #1443, opened 2019-11-22
Switch the Anthos book to direct html

nik9000:
I've picked this fairly small and self contained book to be the first
multi page book to switch to direct html *because* it is simple.


nik9000:
Let's see what kind of diff we get here......

nik9000:
Oooh! An unexpected failure. Neat!

nik9000:
The diff isn't *quite* empty so I'm going to hack on it a little more!

nik9000:
The diff is now empty!

PR #1448, opened 2019-11-22
Fix the title on chunked pages

nik9000:
In docbook we set the title to chunked pages to
"Chunk Title | Main Title | Elastic" so this does the same in
direct_html.


nik9000:
Step by step!

PR #1449, opened 2019-11-23
Fix direct_html chunking

nik9000:
`--direct_html`'s chunking was broken in two ways:
1. It'd only chunk when the section level matched the chunk level
exactly. It is supposed to chunk when it is less than or equal to the
level. This does that.
2. It wasn't using the standard table of contents mechanism to build the
table of contents. Which is trouble because our fancy table of contents
rendering code wasn't kicking in *and* because the table of contents was
getting scattered across the chunked pages. This fixes that too.


nik9000:
Much chunking!

PR #1450, opened 2019-11-23
Clean up direct_html's dc meta tags

nik9000:
Instead of adding the tags with regex manipulation add them by
overriding docinfo. This is the same mechanism that the chunked users to
get `<link rel="home"` into the head and it is *much* cleaner.


nik9000:
We're getting pretty close on the first multi-page book!

PR #1453, opened 2019-11-25
Generate related links for direct_html

nik9000:
Adds the `<link rel="home"`, `<link rel="up"`, `<link rel="pre"`,
and `<link rel="next"` tags to the header for direct_html.


nik9000:
This one is a little fiddly, mostly because "table of contents order" isn't really the order that the html files are generated or the order that the sections appear in the docs.

nik9000:
OK! I think we're getting super close to getting multi-page books now!

PR #1454, opened 2019-11-25
Wrap toc in comments in direct_html

nik9000:
Our docbook extensions wrap the table of contents in `<!--START_TOC-->`
and `<!--END_TOC-->` to make some manipulation by the docs build
simpler. This adds those comments to the when generating the html
directly from Asciidoctor.


nik9000:
Even small steps help!

PR #1455, opened 2019-11-25
Disable headers in direct_html subpages

nik9000:
This disables the "header" parts of `--direct_html`'s "subpages". These
pages only really *want* the contents of the subpage, not the
"titlepage" that comes with a "full" book.


nik9000:
Another step! Another step1

PR #1457, opened 2019-11-25
Add version drop downs to raw

nik9000:
Adds the version drop down to the raw docs so they show up in the
preview. It'd be nice if the templating system did this on the fly, but
that is a problem for another day. This is a simple solution that fixes
the preview so we can keep doing other things.


nik9000:
@olksdr could you look at this one too pretty please?

nik9000:
> Wow, it was requested 2 days ago.

Lol. You are super fast!

PR #1460, opened 2019-11-25
Handle the direct_html's different image class

nik9000:
`--direct_html` emits images wrapped in `<span class="image">` while
docbook emits them as `<span class="inlinemediaobject">`. This updates
the style so they both render the same and updates the `html_diff` tool
so we ignore that difference.


nik9000:
With this change the book I'm testing in #1443 starts to pretty similar. `html_diff` still disagrees, and there is certainly more work to do. But it is really coming along!

PR #1461, opened 2019-11-25
Normalize direct_html's ordered lists

nik9000:
This makes `--direct_html`'s ordered lists look visually the same as
docbook's ordered lists by removing the extra `<p`> tags on the entries
and adding docbook's standard classes. Docbook adds some extra "stuff"
to the lists which `--direct_html` doesn't even after this change but it
doesn't make a visual difference so we normalize those differences away
with `html_diff`.


nik9000:
<ol>
<li> Done!
</ol>

PR #1462, opened 2019-11-26
Fix a direct_html crash on link to inline anchor

nik9000:
Fixes an infinite loop that in the direct_html code that comes up when
you like to an inline anchor. These are *fairly* common and were causing
the Elasticsearch reference not to build with `--direct_html`.


nik9000:
> nice catch. :shipit:

It took quite a bit of print statements to find it!

PR #1463, opened 2019-11-26
Drop wrapper div in subpages

nik9000:
Drops a wrapper `<div>` that declares that we're converting a `book` or
an `article` for subpages produced by `--direct_html`. This lines them
up *much* better with docbook.


nik9000:
Zoom!

PR #1473, opened 2019-11-27
Switch installation guide direct_html

nik9000:
Switches the installation and upgrade guide to `--direct_html` which
should speed up building it quite a bit!


nik9000:
This one needs:
* [ ] Implement the xpack tag on headings.
* [ ] Support for titleabbrev
* [ ] Remove the title from a sidebar if one isn't set
* [ ] Inline admonitions
* [ ] Teach html_diff to ignore duplicate classes
* [ ] Support tables with titles like the one on beats-breaking-changes.html.

+ A few more after that is fixed, probably.

nik9000:
New list of needs:
* [ ] Teach html_diff to ignore duplicate classes
* [ ] Implement xpack tag on floating titles
* [ ] Change admonitions

nik9000:
Diff looks good.

PR #1474, opened 2019-11-27
Render deep toc in --direct_html

nik9000:
This enables rendering subsections in the table of contents in a way
that mimicks docbook. In doing so it fixes a crash when trying to
convert chunked books that have subsections caused by the chunker not
seeing the subsections in the toc. This adds them so it is happy.


nik9000:
This fixes the crash that happens when you try to build the Elasticsearch reference or #1473.

nik9000:
I write more tests than code!

PR #1475, opened 2019-11-27
Drop index from direct_html's nav

nik9000:
It turns out that docbook doesn't add the `index.html` as a "prev" page
in docbook's nav even though it *does* add it as `<link rel="prev"`.
Funny thing, that. Anyway, this makes us line up with docbook.


nik9000:
We're so close!

PR #1476, opened 2019-11-27
Fix override text on xref in direct_html

nik9000:
Stops us from ignoring explicit link text like `<<target,override text>>`
on cross references in direct_html. I was so concerned with getting he
text right in the complex cases I didn't do the obvious case where it is
specified....


nik9000:
I thinks this is it!

PR #1477, opened 2019-11-27
Switch the getting started guide to direct_html

nik9000:
This should speed up building it and get us more tests for
`--direct_html`.


lcawl:
LGTM!

nik9000:
I believe this'll need #1493, #1492, and #1491. It'll want #1483 too, but the difference without it is totally not a blocker.

nik9000:
My last diff wasn't entirely clean but the changes looked like in flight doc changes rather than things caused by --direct_html. I'm waiting on a [rebuild](https://elasticsearch-ci.elastic.co/view/Docs/job/elastic+docs+master+build/8664/console) I just kicked off. Once I finish that I'll kick off another build.

nik9000:
OK! One last thing, and it is only in 7.5 and 7.4: we need to add support for titles on code blocks.

nik9000:
The diff is now empty but for a couple of empty `<p>` tags being removed. Good enough for me.

PR #1478, opened 2019-11-27
Improve list rendering in direct_html

nik9000:
List items always have text and *sometimes* have complex content. If the
list item has complex content then docbook wraps the text in
`<p class="sempara>`. We should do that too for compatibility.


nik9000:
More steps! More better!

PR #1481, opened 2019-11-27
Add "other versions" link to versions drop down

nik9000:
This adds an "other versions" link to the versions drop downs if we
stripped any versions from the drop down because they aren't "live". We
remove non-"live" versions from the drop down because we don't want
folks accidentally picking up an old version and because the list just
gets too long with all the versions listed. But we still publish all the
old versions in case some unfortunate soul is stuck on that version.
This adds an "other versions" option to the versions drop down so those
folks can find their version.

### DIFF
--- a/resources/web/docs_js/utils.js
+++ b/resources/web/docs_js/utils.js
@@ -9,6 +9,7 @@ const VERSION_REGEX = /[^\/]+\/+([^\/]+\.html)/;
 export function get_current_page_in_version(version) {
   if (version === "other") {
     location.href = location.href.replace(VERSION_REGEX, "index.html");
+    return;
   }
   var url = location.href.replace(VERSION_REGEX, version + "/$1");
   return $.get(url).done(function() {
--- a/resources/web/docs_js/utils.js
+++ b/resources/web/docs_js/utils.js
@@ -5,9 +5,12 @@ export function get_base_url(href) {
              .replace(/^http:/, 'https:');
 }
 
+const VERSION_REGEX = /[^\/]+\/+([^\/]+\.html)/;
 export function get_current_page_in_version(version) {
-  var url = location.href;
-  var url = location.href.replace(/[^\/]+\/+([^\/]+\.html)/, version + "/$1");
+  if (version === "other") {
+    location.href = location.href.replace(VERSION_REGEX, "index.html");
+  }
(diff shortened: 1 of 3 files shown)


nik9000:
@JoshMock this one probably needs your eyes too. We don't have tests for that particular bit of javascript yet so I didn't add any.

resources/web/docs_js/utils.js
```diff
@@ -5,9 +5,12 @@ export function get_base_url(href) {
              .replace(/^http:/, 'https:');
 }
 
+const VERSION_REGEX = /[^\/]+\/+([^\/]+\.html)/;
```
JoshMock: because I like to verify regexes: this is matching on exactly one directory + one file, e.g. `foo/bar.html`, but not `foo/bar/baz.html` or `foo.html`

resources/web/docs_js/utils.js
```diff
@@ -5,9 +5,12 @@ export function get_base_url(href) {
...
 export function get_current_page_in_version(version) {
-  var url = location.href;
-  var url = location.href.replace(/[^\/]+\/+([^\/]+\.html)/, version + "/$1");
```
JoshMock: oh I guess you just moved an existing regex up. 😁 context, Josh. context.

lcawl:
I'm not sure how it happened, but I got into a bad state by clicking the "other versions":

image

nik9000:
> I'm not sure how it happened, but I got into a bad state by clicking the "other versions":

Huh. I saw a flash of something but the page reloaded any way. I'll play with it.

nik9000:
@lcawl, I think I fixed it. Could you have a look [here](http://niktest.docs-preview.app.elstc.co/guide/en/beats/filebeat/current/index.html)?

lcawl:
Thanks, if I select "other versions" it takes me to the expected website.  If I go back to the previous page it still says "other versions" in the drop-down, but it no longer generates the error I saw previously. It's simple enough to select a "real" version again, so I don't think this is a blocker anymore.

nik9000:
> Thanks, if I select "other versions" it takes me to the expected website. If I go back to the previous page it still says "other versions" in the drop-down, but it no longer generates the error I saw previously. It's simple enough to select a "real" version again, so I don't think this is a blocker anymore.

Yeah, I think the back button does that across the board though. It'd be nice to fix....

PR #1482, opened 2019-11-27
Tolerate direct_html's toc

nik9000:
We use perl code to hack the book's title into the table of contents.
Because we like it there. *Technically* we could use an Asciidoctor
customerization to do it in direct_html but in the spirit of keeping the
changes as small as possible I've just fixed the perl code to tolerate
direct_html output.


nik9000:
This should fix #1473 and #1477.

nik9000:
> lgtm

:heart:

PR #1483, opened 2019-11-27
Replace "simpara" style with one more explicit

nik9000:
Sometimes docsbook spits out `<p>` tags with `class="simpara"` and
sometimes it doesn't. I can't quite figure out the rules, but I'd like
it not to matter. And it *mostly* doesn't. There was just one css rule
that used `simpara` and of course it wasn't commented. But I've reverse
engineered it: it is designed to make lists that directly follow
paragraphs that are *inside* of `<dd>` or `<li>` closer to the
paragraph. Without it the lists look "lonely" and separate from the
paragraph.

So I replaced it with a css rule that explicitly looks for `<dd>` and
`<li>`.

### DIFF
--- a/resources/web/style/docbook.pcss
+++ b/resources/web/style/docbook.pcss
@@ -13,13 +13,6 @@
   .chapter div.section {
       margin-top: 2em;
   }
-
-  p.simpara+div {
-    &.orderedlist,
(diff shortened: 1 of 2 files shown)


nik9000:
Thanks @JoshMock !

PR #1489, opened 2019-11-29
Clean up html diff classes some more

nik9000:
Cleans up the html diff implementation's class handling a little:
1. Change one last class for screenshots
2. Always sort the class list. This came up when handling screenshots
but feels pretty generate.
3. Always remove empty class lists. This came up a while ago but feels
like something I should handle more generally.


nik9000:
Thanks @olksdr !!

PR #1491, opened 2019-11-29
Add version to some titles in direct_html

nik9000:
Docbook adds the version of the book to some title attributes on some
links. This teaches `--direct_html` to add them there too.


nik9000:
:heart: all around!

PR #1492, opened 2019-11-29
Implement noindex in direct_html

nik9000:
Implements the `noindex` flag in direct html which causes us to add a
header to the page that asks robots not to index or follow the page.
Docbook has this so we should too.


nik9000:
Step by step!

PR #1493, opened 2019-12-01
Add extra header support to direct_html

nik9000:
We use "extra" header elements on the page to convey things like "this
version is out of date" or "this version has yet to be released".
`--direct_html` was ignoring those extra headers. This adds support for
it.

### DIFF
--- a/resources/asciidoctor/lib/docbook_compat/convert_document.rb
+++ b/resources/asciidoctor/lib/docbook_compat/convert_document.rb
@@ -1,10 +1,15 @@
 # frozen_string_literal: true
 
+require_relative 'extra_docinfo'
+require_relative 'munge_body'
+
 module DocbookCompat
   ##
   # Methods to convert the document at the top level. All of these are a bit
   # scary but required at this point for docbook compatibility.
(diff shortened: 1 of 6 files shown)


resources/asciidoctor/lib/docbook_compat/convert_document.rb
```diff
@@ -17,21 +22,6 @@ def convert_document(doc)
       html + "\n"
     end
 
-    ##
```
nik9000: I moved this one to "make room" for the new features in this file. rubocop  doesn't like when files get too big. Neither do I, but it is better at noticing.

resources/asciidoctor/lib/docbook_compat/convert_document.rb
```diff
@@ -66,36 +56,6 @@ def munge_meta(html)
         raise("Couldn't remove generator in #{html}")
     end
 
-    def munge_body(doc, html)
```
nik9000: Same here!

nik9000:
Thanks @olksdr!

PR #1494, opened 2019-12-01
Switch glossary to direct_html

nik9000:
`--direct_html` is faster and much more customizable. It is the future
of docs.


nik9000:
Looks like this one needs some work on definition lists!

nik9000:
After #1497 the diff looks empty! I'm going to call it done!

PR #1495, opened 2019-12-01
Switch the legacy sense docs to direct_html

nik9000:
They are faster and more customizable. They are the future.


nik9000:
This one needs some work on definition lists and images with titles.

nik9000:
I believe this one will be fine once we get #1504. The diff isn't entirely empty, but it is close. The only visual difference is that links to images with titles no longer include the image number. This is so so rare in our docs that I'm fine letting it through.

PR #1496, opened 2019-12-01
Switch ecctl docs to direct_html

nik9000:
`--direct_html` is faster and more customizable. It is the future.


nik9000:
@nrichers, I'm not sure `--direct_html` is ready for the ecctl docs. I'll be iterating on `--direct_html` until it *is* ready though.

nik9000:
It looks like this one needs:
1. #1493
2. Fix for the html emitted when a snippet doesn't have a language.
3. Some work on the `<title>` when there is a subtitle on a section that generates a page.
4. Fix around the `beta` admonition. `--direct_html` isn't emitting some text that it should be emitting.

nik9000:
@nrichers the diff is now empty so I'm going to convert the ecctl docs to `--direct_html`. Please let me know if you find any issues!

PR #1497, opened 2019-12-02
Fix compatibility for direct_html's dl

nik9000:
Implement docbook compatibility for direct_html's description lists.
This should be enough to convert our glossary to direct_html.

This also removes the double trailing new lines from `</p>`. Asciidoctor
already seems to add a newline for those.

This also cleans up a bug in list rendering that I encountered while
doing real world testing of the description lists: if you form an
ordered list with `1.` instead of `.` then it wasn't getting the
override. This fixes that.

This looks to be enough to unblock #1494.


nik9000:
Thanks @JoshMock !

PR #1498, opened 2019-12-02
Clean up direct_html difference in titled images

nik9000:
`--direct_html` produces html for images with titles that is fairly
different than what docbook makes. But with a little CSS this makes them
*look* the same. And with a little more python code this makes the diff
tool ignore the changes to the html.


nik9000:
@JoshMock, I think this one is for you!

nik9000:
Thanks @JoshMock !

PR #1499, opened 2019-12-02
Be slightly less picky in normalizing links

nik9000:
There were some cases where we Asciidoctor and Docbook don't agree on
the classes on links. Those classes don't make and visual difference so
this ignores them. We actually already *tried* to ignore them, but we
didn't do it all the time.


nik9000:
Thanks @JoshMock !

PR #1501, opened 2019-12-02
Properly chunk footnotes in direct_html

nik9000:
Asciidoctor allows you to make footnotes and we use them. This makes
sure that they end up on the same page as they are declared. Without it
the footnotes end up on the "index" page which is super confusing.

NOTE: I'm note really making an effort to make this line up perfectly
with docbook's output. These *should* be rare enough that I can verify
them by hand. We really don't use *that* many footnotes.

### DIFF
--- a/resources/web/docs_js/index.js
+++ b/resources/web/docs_js/index.js
@@ -18,7 +18,7 @@ export function init_headers(right_col, lang_strings) {
   var ul = $('<ul></ul>').appendTo(this_page);
   var items = 0;
 
-  $('#guide a[id]').each(
+  $('#guide a[id]:not([href])').each(
     function() {
       // Make headers into real links for permalinks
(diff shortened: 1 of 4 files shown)


nik9000:
Required for #1473.

resources/web/docs_js/index.js
```diff
@@ -18,7 +18,7 @@ export function init_headers(right_col, lang_strings) {
   var ul = $('<ul></ul>').appendTo(this_page);
   var items = 0;
 
-  $('#guide a[id]').each(
+  $('#guide a[id]:not([href])').each(
```
nik9000: This was borking all of the footnotes.

nik9000:

nik9000:
:heart:

PR #1504, opened 2019-12-02
Properly handle level 3 sections in direct html

nik9000:
This fixes the wrapper class on level 3 sections in direct html.
Required for #1495.


nik9000:
Thanks @JoshMock !

PR #1505, opened 2019-12-02
Switch the stack overview to direct_html

nik9000:


nik9000:
I'm not sure if this is ready yet, but I'm going to use this to make a diff and take a look.

nik9000:
That build failure looks legit.

nik9000:
Now it looks like there are broken links! Neat!

nik9000:
@elasticmachine, run elasticsearch-ci/docs skip link check

nik9000:
@elasticmachine, run elasticsearch-ci/docs skiplinkcheck

nik9000:
Things to do
* [ ] Hunt down the broken links.
* [ ] Fix "up" links that to contain "Appendix A. " if linking to an appendix
* [ ] Fix inline anchors
* [ ] Remove the deleted pages appendix from the TOC. It has a `role=exclude` and we should use that to drop it from the index.

nik9000:
I've regenerate the diffs and it is nearly empty. Here is what is left, all are visible in the 6.3 branch:
* [ ] Handling block quotes
* [ ] Loss of text on the link to `security-privileges.html#privileges-list-cluster`
* [ ] Extra `<br/>` all over `ml-gs-forecast.html`

nik9000:
>  Extra `<br/>` all over ml-gs-forecast.html

The breaks are actually correct because of the trailing `+` on the end of the line! Docbook has been eating them. I guess we'll just be ok with that. They don't actually make a visual difference.

nik9000:
The diff for this one is empty but for the explict `<br/>` tags that docbook was destroying. They can pass through.

PR #1506, opened 2019-12-02
Switch azure docs to direct_html

nik9000:


nik9000:
There are two differences left:
1. Admonitions aren't getting titles.
2. The console widget isn't coming through.

nik9000:
The diff looks good now!

PR #1507, opened 2019-12-02
Handle complex content in tables in direct_html

nik9000:
This adds support for complex content inside of tables like lists and
things to `--direct_html`. Without this change complex content would
cause us to crash!

Required for #1505


nik9000:
Thanks @olksdr !

PR #1508, opened 2019-12-03
Render listing titles in direct_html

nik9000:
When a code listing has a title docbook wraps it in `<p><strong>`. So
should we.


nik9000:
:rocket: :computer:

PR #1509, opened 2019-12-03
Add some tests for a mysterious condition

nik9000:
It turns out that edit_me urls in the logstash plugin reference weren't
being generated quite right. I dug into it and found out that html5
fixes it which is lovely, if a little strange. This adds tests to make
sure it *stays* fixed.


nik9000:
Every test makes us stronger!

PR #1512, opened 2019-12-03
Switch logstash reference to direct_html

nik9000:
direct_html is faster and more customizable *and* it generates better
`edit_me` links for this book.


nik9000:
@karenzone, I'm opening this so that CI will make a preview and a diff for me. I don't imagine it'll be perfect, but we can look at it and I can fix whatever i wrong.

nik9000:
OK! Now that we aren't crashing I can make a list of things required for this:
* [ ] Fix rendering of code blocks without a language
* [ ] Remove the HTML from the title of inline links
* [ ] There are some places where we make `<h2>` docbook is making `<h1>`. I have no idea what is up.

The middle one in particular is *so* noisy that I can't get an accurate list without fixing it.

nik9000:
OK! The diff isn't quite empty but most of it is `<br>` tags that come from the `+` inside a line. There are a few edit_me url changes but they all look more right than what we had before.

PR #1513, opened 2019-12-03
Implement q and a style dlists in direct_html

nik9000:
Docbook handles "question and answer" style description lists
differently than regular description lists. So `--direct_html` has to do
the same.

This also adds protection around "missing" styles of description lists.
Instead of just crashing we'll report and error and refuse to convert
the dlist.

This is required for converting logstash's reference which is #1512.


nik9000:
:+1:

PR #1514, opened 2019-12-03
Ignore empty <p> when comparing direct_html

nik9000:
Docbook spits out empty `<p>` tags sometimes. I'm not sure why. But I'm
95% sure we don't want them. Sure enough that we can ignore them in the
diff and clean them up if anyone complains later.


nik9000:
Thanks @olksdr !

PR #1515, opened 2019-12-03
Support title on admonitions in direct_html

nik9000:
You can add a title to admonitions with
```
[WARNING]
.Title goes here
--
Content goes here
--
```

This makes direct_html produce the same output as docbook for those
titles.

Required for #1506.


nik9000:
Thanks @olksdr !

PR #1516, opened 2019-12-03
Retire 7.4 from the live stack releases

jrodewig:
With the release of 7.5.0, the 7.4.x branches passed their maintenance date:
https://www.elastic.co/support/eol

This retires the 7.4 branch from the list of "live" Stack release branches in our docs.

I believe this is our first time doing this so let me know if we want to formalize the process a bit more. I see benefits in both keeping the branch around longer _or_ updating the list as part of release (or shortly after).

Appreciate any feedback!


jrodewig:
I'll go ahead and merge this in the interest of shuffling 7.4 off the list. If there's need for further discussion, I'm happy to create a discuss issue. Thanks!

nik9000:
I think you *might* have to force a rebuild to get this to take effect. It is worth seeing what happens after the merge to check!

jrodewig:
> I think you _might_ have to force a rebuild to get this to take effect. It is worth seeing what happens after the merge to check!

Thanks! I'll force a manual rebuild.

jrodewig:
> I think you _might_ have to force a rebuild to get this to take effect. It is worth seeing what happens after the merge to check!

It looks like a full rebuild is needed after this change. Good to note. Thanks @nik9000!

PR #1517, opened 2019-12-04
Implement the console widget in direct_html

nik9000:
Implements that `console` widget and its friends the `kibana`, `sense`,
`ess`, and `ece` widgets in direct_html. This also enables copying the
snippets for `ess` and `ece` which we haven't done before. It isn't
*strictly* needed, but it is simpler to do for now and it doesn't hurt
anything.

Required for #1506


nik9000:
Thanks @olksdr !

PR #1520, opened 2019-12-04
Tweak diff algorithm for direct_html

nik9000:
When I built the diff for logstash I discovered a few small tweaks I
could make to the diff algorithm:
1. Totally ignore any "type" attributes on `ul` and `li` because we
override the type in css.
2. Ignore "double wrapping" of `<code>` tags that docbook can sometimes
make.

Required for #1512


nik9000:
Logstash is on its way!

PR #1521, opened 2019-12-04
Fix rendering of no lang listings in direct_html

nik9000:
When you declare a code listing without a language docbook renders it
quite differently than one *with* a language. This makes direct_html do
the same.


nik9000:
:baby: steps!

PR #1522, opened 2019-12-04
Don't include "role=exclude" entries from the toc

nik9000:
Docbook uses the `role=exclude` mechanism to exclude pages from the
table of contents. This makes direct_html do the same.


nik9000:
:baby: steps!

PR #1523, opened 2019-12-04
Add support for xpack tag in direct_html

nik9000:
We've customized docbook to add a spec "tag" to the title of pages that
are marked with `role=xpack`. This implements that customization in
direct_html. It was way, way, way, way easier than the docbook
customization.


nik9000:
X

PR #1525, opened 2019-12-04
Render ids for paragraphs in direct_html

nik9000:
It is possible to declare an id for a paragraph like so:

```
[[foo]]
Words in the paragraph.
```

Docbook renders this like:

```
<p><a id="foo"></a>Words in the paragraph</p>
```

So we should too.


nik9000:
Much cleaner diffs now!

PR #1528, opened 2019-12-04
Revised  default notice for older versions.

debadair:
This updates the default notice for versions older than current. With this change, we do not have to override the default notice in every OOM version, but we *will* have to override it for the maintenance version of the previous minor (currently 6.8).

image

That's simpler than having to do that for every OOM version, so I think it's a win. At some point, we could automatically apply the correct notice to "live" versions that aren't current and eliminate even  that manual step.

### DIFF
--- a/lib/ES/Book.pm
+++ b/lib/ES/Book.pm
@@ -15,8 +15,8 @@ use utf8;
 our %Page_Header = (
     en => {
         old => <<"HEADER",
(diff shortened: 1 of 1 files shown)


debadair:
I've merged the overrides for 6.8 in the docs that use the stack versioning. This default notice will be applied to versions that don't have an explicit page_header.html file.

PR #1529, opened 2019-12-05
Implement strange h level bumping in direct_html

nik9000:
Docbook bumps the `<h` tag levels in level 0 section and their children.
But *not* the their grand children. Which is strange. But we're
emulating docbook. So this does that too. Oh yeah, and only "bar" level
0 sections count. If you make a level 0 section an appendix then no bumping.


nik9000:
:bug:

PR #1530, opened 2019-12-05
Support default admonition text in direct_html

nik9000:
Our custom "experimental" and "beta" admonitions have default text if
you don't supply any. This adds that default text to direct_html and
cleans up the tests around admonitions a bit, including adding tests for
our custom admonitions.


nik9000:
One day all the diffs will be empty!

PR #1531, opened 2019-12-05
Ignore difference in direct_html's partintro

nik9000:
direct_html renders partintro blocks slightly differently than docbook
but they aren't visually distinct so we can ignoree that difference.
This teaches the html_diff tool to ignore it.


nik9000:
The diff for #1505 is almost readable now!

PR #1532, opened 2019-12-05
Handle tables with titles in direct_html

nik9000:
Docbook wraps tables with titles differently than tables without. This
adds that wrapping to direct_html.


nik9000:
*now* the diff is quite small!

PR #1533, opened 2019-12-05
Remove edit links form non-live versions

nik9000:
This removes the `edit` links from versions of books that aren't "live"
because, for the most part, we don't want folks opening PRs against
them. I mean, we'd *prefer* folks open PRs against the master branch,
but that isn't really a simple thing to do when fixing docs so we're
pretty happy with whatever. But we'd really prefer PRs to target live
branches.


nik9000:
Thanks @lcawl and @olksdr !

Issue #1536, opened 2019-12-06
Difficult to find older versions of documentation


sylvie777:
Provide the following information to help us identify and troubleshoot the issue.
- Page URL: e.g. https://www.elastic.co/guide/en/elasticsearch/reference/current/watcher-api-put-watch.html
- Device type: Laptop
- Specifics (e.g., OS, Hardware, Browser, etc.): Chrome Version 78.0.3904.108 (Official Build) (64-bit) on Ubuntu 18.04

List out steps on how to recreate the issue.
1. Go to the current version of a page of documentation.  Today it shows:
![screenshot-2019-12-06_11-42-09](https://user-images.githubusercontent.com/174291/70317171-86a06c00-181d-11ea-8b7d-f7c3b3ce12a7.png)
... but I want to see the docs for 6.6
2. So I click on "other versions"
3. I then see the list of "Elasticsearch Reference" for all of the previous versions and click on "Elasticsearch Reference 6.6":

![screenshot-2019-12-06_11-43-48](https://user-images.githubusercontent.com/174291/70317245-ae8fcf80-181d-11ea-94f4-af19a17297af.png)

But now I'm right at the root of the reference guide... many many clicks away from where I was in the documentation pages, and I probably would not know how to navigate there because I got there by Google in the first place.

This is new (because there used to be more older versions listed) and it makes finding the correct documentation for older (but still supported and pre EOL!) versions very difficult.
In support we _frequently_ have to find the correct version of documentation matching the version the customer is using to share with them, so this new reduced version drop-down is quite a pain.

nik9000:
Relates to #1419.

PR #1538, opened 2019-12-06
Refine "other versions" behavior

nik9000:
When you click the "other versions" option it used to take you to the
version link farm page. While it is nice to get all the versions this
loses the one of the nice behaviors of the versions drop down, namely
that it checks if the page you are on exists in the target version and
drops you right on that page if it does.

This change stops it from sending you to the link farm and instead
un-hides another drop down that contains *all* of the versions.

Relates to #1419 and #1536.

### DIFF
--- a/resources/web/docs_js/index.js
+++ b/resources/web/docs_js/index.js
@@ -128,14 +128,19 @@ function init_toc(lang_strings) {
   var v_selected = title.find('select option:selected');
   title
     .find('select')
-    .change(function() {
-       var version = title.find('option:selected').val();
-       utils.get_current_page_in_version(version).fail(function() {
-         v_selected.attr('selected', 'selected');
-         alert(lang_strings('This page is not available in the docs for version:')
-               + version);
-       });
-     });
+    .change(function(e) {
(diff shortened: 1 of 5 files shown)


nik9000:
This totally isn't the best solution but it feels a little better than what we have and it was pretty quick to put together.

resources/web/docs_js/index.js
```diff
@@ -128,14 +128,19 @@ function init_toc(lang_strings) {
...
-         v_selected.attr('selected', 'selected');
-         alert(lang_strings('This page is not available in the docs for version:')
-               + version);
-       });
-     });
+    .change(function(e) {
+      var version = $(e.target).find('option:selected').val();
+      if (version === "other") {
+        $("#other_versions").show();
+        $("#live_versions").hide();
```
JoshMock: Should this toggle the visibility in the other direction if `version !== "other"`? If so, using [`.toggle()`](https://api.jquery.com/toggle-event/) might be a more inclusive implementation.

nik9000: Once you pick "other" you can get all the versions so I don't think we need to toggle.

PR #1540, opened 2019-12-06
Handle "width" in tables in direct_html

nik9000:
Docbook handles `[width=50%]` on tables by adding the width to the
`<table>` tag. So we should too, just to be even.


nik9000:
There a zillion more table things to come one day!

PR #1541, opened 2019-12-06
Ignore difference between examples in direct_html

nik9000:
direct_html renders examples slightly differently than docbook but they
don't make any visual difference. So we ignore the difference.


nik9000:
Thanks!

PR #1543, opened 2019-12-09
Fix two connected issues with title in direct_html

nik9000:
* Pass the *caption* of the title into subpages. This adds things like
"Appendix A:" to the front of pages for the appendix.
* Remove the title-separator from subpages so their `<title>` renders
properly. Now we only use the separator for the *main* page.  Without
this `Appendix A: Foo` would get rendered as `Appendix A` because the
`Foo` part would be considered a subtitle.


nik9000:
Thanks @olksdr !

PR #1544, opened 2019-12-09
Handle inline admonitions in direct_html

nik9000:
Elastic's admonition extensions can be used inline with something like:
```
The foo is experimental::[some warning text].
```

This adds support for that to direct_html.


nik9000:
Thanks @olksdr and @JoshMock !

PR #1545, opened 2019-12-09
Handle attributes in `titleabbrev` in direct_html

nik9000:
You can use an attribute in a `titleabbrev` like so:
```
++++
<titleabbrev>{security}</titleabbrev>
++++
```

The `{security}` thing *should* get replaced with the attribute value.
This does that in direct_html and solves a *ton* of problems converting
logstash (#1512).


nik9000:
Thanks @olksdr !

PR #1546, opened 2019-12-09
Drop "out of date" from version drop down

nik9000:
I added "out of date" to the version drop down if you are on a page for
a non-live version of the docs but it turns out we're adding a heading
to the docs that says exactly that so this is duplicate information and
it can be confusing.


nik9000:
Thanks @olksdr !

PR #1547, opened 2019-12-09
Clean xref title in direct_html

nik9000:
When you make a cross reference to a section who's title contains html
we were generating invlid html. For example, a link to this:
```
[[foo]]
== `foo`
```

Would come out with `<a title="<code ....>foo</code>"`. That ain't
right. This strips the html which is *super* helpful for converting
logstash (#1512).


nik9000:
:baby: :steam_locomotive:

PR #1548, opened 2019-12-09
Fix ref text for inline anchors in direct_html

nik9000:
When you reference something like
```
== Section heading
This
[[privileges-list-cluster]]
page has moved.
```

direct_html *was* producting an empty string for the default text while
docbook was using the section heading as the default text. This fixes up
direct_html to match docbook.

### DIFF
--- a/resources/asciidoctor/lib/docbook_compat/convert_links.rb
+++ b/resources/asciidoctor/lib/docbook_compat/convert_links.rb
@@ -31,22 +31,16 @@ module DocbookCompat
     end
 
     def ref_text_for(ref, node)
-      if ref.node_name == 'inline_link'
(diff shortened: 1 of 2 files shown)


resources/asciidoctor/lib/docbook_compat/convert_links.rb
```diff
@@ -31,22 +31,16 @@ def convert_xref(node)
     end
 
     def ref_text_for(ref, node)
-      if ref.node_name == 'inline_link'
```
nik9000: It turns out we get this free by calling `xreftext`.

nik9000:
We're getting there slowly but surely!

PR #1549, opened 2019-12-09
Support title-less sidebar in direct_html

nik9000:
direct_html's implementation of `[sidebar]` produced an "empty" title if
the sidebar didn't have a title but docbook skipped it. This skips the
title to line up with docbook.


nik9000:
Thanks @olksdr !

PR #1551, opened 2019-12-09
Add support for colspan and rowspan to direct_html

nik9000:
Adds support for `colspan` and `rowspan` to direct_html's tables.


nik9000:
Thanks @olksdr !

PR #1552, opened 2019-12-09
Support quotes in direct_html

nik9000:
Docbook renders attributed quotes radically differenly than asciidoctor.
This gets direct_html to look like docbook.


nik9000:
Thanks @olksdr !

PR #1553, opened 2019-12-09
Don't add . to title ending in : in direct_html

nik9000:
When the title of a listing ends in a `:` we shouldn't add a `.` to it
because docbook doesn't and because it looks silly.


nik9000:
Thanks @olksdr !

PR #1555, opened 2019-12-09
Handle titles on paragraphs in direct_html

nik9000:
Adds support for titles on paragraphs to direct_html. These are written
like:
```
.Title
Paragraph body.
```


nik9000:
Thanks @olksdr !

PR #1556, opened 2019-12-10
Handle change admonitions in direct_html

nik9000:
This handles change admonitions and cleans up the care admonitions to
look like the change admonitions. For the most part that means moving
the "default text" of the care admonitions into the care admonition
extension itself. This will change the generated docbook xml but it
*shouldn't* change the resulting html.


nik9000:

I want to make double sure none of the docbook generated html changes and we don't have good tests for that so I just have to diff the world.

nik9000:

nik9000:
Thanks @olksdr!

PR #1557, opened 2019-12-10
Support more table formatting in direct_html

nik9000:
Asciidoctor supports explicitly setting the format on table cell so we
should respect that in direct_html. This sets up direct_html to mimic
how it all turns out in docbook.


nik9000:
Thanks @olksdr !

PR #1558, opened 2019-12-10
Switch the java rest client to direct_html

nik9000:
direct_html is the future!


nik9000:
The diff is *nearly* empty. In fact, the only real change is an example that docbook mangles and asciidoctor mangles less.

PR #1559, opened 2019-12-10
Switch ecs to direct_html

nik9000:
direct_html is faster and more customizable.

PR #1560, opened 2019-12-10
Rework table cell rendering in direct_html

nik9000:
It turns out that asciidoctor can do most of the job for us. Also, I
have to do something with paragraphs to fix errors I found while
converting ecs (#1559).

### DIFF
--- a/resources/asciidoctor/lib/docbook_compat/convert_table_cell.rb
+++ b/resources/asciidoctor/lib/docbook_compat/convert_table_cell.rb
@@ -41,48 +41,8 @@ module DocbookCompat
     end
 
     def cell_text(cell)
-      return cell.text unless (style = cell.attr 'style')
-
-      cell.document.converter.convert cell, "cell_text_#{style}"
-    rescue NoMethodError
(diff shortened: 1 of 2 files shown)


resources/asciidoctor/lib/docbook_compat/convert_table_cell.rb
```diff
@@ -41,48 +41,8 @@ def convert_cell_content(cell, allow_formatting)
...
-    end
-
-    private
-
-    def delegate_cell_text(cell, type)
-      Asciidoctor::Inline.new(
-        cell.parent, :quoted, cell.text, type: type
-      ).convert
+      cell.style = :strong if cell.style == :header
+      cell.content.join "</p>\n<p>"
```
JoshMock: How does this handle the first opening `<p>` and final closing `</p>`?

nik9000: It's in the method above. It is kind of weird. Let me clean that up.

JoshMock:
:+1: that helps!

PR #1561, opened 2019-12-10
Clean up direct_html's inline change admonitions

nik9000:
This fixes some small errors with direct_html's handling of inline
change admonitions that came up because I didn't have a great example to
compare against when I wrote them. I found an example though! This adds
a missing space between some classes, adds some more text to the
message, and fixes up the diff tool to ignore a small diff that isn't
visible.


nik9000:
Thanks @JoshMock !

PR #1562, opened 2019-12-10
Switch the painless book to direct_html

nik9000:
Direct html is faster and more customizable. It is the future!

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -358,6 +358,7 @@ contents:
             chunk:      1
             tags:       Elasticsearch/Painless
             subject:    Elasticsearch
+            direct_html: true
             sources:
               -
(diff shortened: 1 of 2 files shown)


nik9000:
I'm opening this to generate a diff and will update it once I have a look at the diff.

nik9000:
My diff found one thing which we should get fixed real soon now: #1563

nik9000:
Now that the fix is in I'm going to rebuild the diff and recheck.

nik9000:
The diff is now empty so I'm merging! @jdconrad, please update the command you use to build the book locally.

jdconrad:
Thanks @nik9000! Will change my docs flow accordingly.

PR #1563, opened 2019-12-10
Support ids on tables in direct_html

nik9000:
Adds support for specifying he id on a table in a docbook compatible way
to direct_html. Required by the painless reference (#1562).


nik9000:
Thanks @JoshMock !

PR #1564, opened 2019-12-10
Switch the ES plugins doc to direct_html

nik9000:
direct_html is the future.


nik9000:
This one needs work:
* [ ] Missing "This page is a place holder for the new plugins documentation coming with Elasticsearch"
* [ ] Pages with html in the title are breaking the `<title>` tags.
* [ ] Tables with ids need to get the id

nik9000:
>  Missing "This page is a place holder for the new plugins documentation coming with Elasticsearch"

I don't believe this is worth fixing because it is only a thing in the 1.7 version which is super, ultra, mega out of date.

nik9000:
Diff looks good!

PR #1565, opened 2019-12-10
Switch the transport client docs to direct_html

nik9000:
direct_html is the future


nik9000:
The diff is fine. There are some *interesting* results around duplicated callouts but there always has been.

PR #1566, opened 2019-12-10
Switch elasticsearch reference to direct_html

nik9000:

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -221,6 +221,7 @@ contents:
             chunk:      1
             tags:       Elasticsearch/Reference
             subject:    Elasticsearch
+            direct_html: true
             sources:
               -
                 repo:   elasticsearch
--- a/doc_build_aliases.sh
+++ b/doc_build_aliases.sh
@@ -12,13 +12,13 @@
 
 
 # Elasticsearch
-alias docbldesx='$GIT_HOME/docs/build_docs --doc $GIT_HOME/elasticsearch/docs/reference/index.asciidoc --resource=$GIT_HOME/elasticsearch/x-pack/docs/ --chunk 1'
+alias docbldesx='$GIT_HOME/docs/build_docs --direct_html --doc $GIT_HOME/elasticsearch/docs/reference/index.asciidoc --resource=$GIT_HOME/elasticsearch/x-pack/docs/ --chunk 1'
 
 alias docbldes=docbldesx
 
 # Elasticsearch 6.2 and earlier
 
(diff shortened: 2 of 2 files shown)


nik9000:
I've tracked the failure down to https://github.com/elastic/elasticsearch/pull/50098 and am fixing it.

nik9000:
For what it is worth I'm seeing time like 97 seconds -> 10 seconds and 69 seconds -> 11 seconds.

nik9000:
run  elasticsearch-ci/docs

nik9000:
run elasticsearch-ci/docs skiplinkcheck

nik9000:
It looks like we have some trouble with the links! I'll try to get to those first, but I'm rebuilding without link checking just to get a diff.

nik9000:
OK! Diff is ready. We need:
* [ ] Support for `center` on images.
* [ ] Fix the title rendering on an itemized list. I *thought* I had examples of this being right but it is worth checking.
* [ ] Ignore the difference between `width="700"` and `width="700px"` because Asciidoctor is more right.
* [ ] Figure out what is up with stuff like `<a class="indexterm" id="id-1.13.7.25.1"></a>`. We almost certainly don't need it. If so, ignore it in the diff.
* [ ] Remove duplicate callouts. Right now we actually generate *better* callouts than we had before, but they don't quite line up. We should sneak that fix in there. Compare [this](https://www.elastic.co/guide/en/elasticsearch/reference/1.3/modules-tribe.html) with [this](http://docs_1566.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/1.3/modules-tribe.html)
* [ ] `1.3/setup-configuration.html` looks funny around the "added" bit.
* [ ] Hunt down the difference in the example on `1.4/docs-termvectors.html`.
* [ ] `1.6/query-dsl-multi-match-query.html` is missing `<a id="operator-min"></a>`.


I'm going to call that enough for round 1.

nik9000:
None of those fix the bad links. Huh. I wonder what is up....

nik9000:
>     * `1.3/setup-configuration.html` looks funny around the "added" bit.

This is caused by the annotation being included into the id. Which it *shouldn't* do, but it is fairly difficult to exclude it.

Other than that, everything up until 2.0 is looking good. Still generating the rest of the diff.

nik9000:
> Other than that, everything up until 2.0 is looking good. Still generating the rest of the diff.

Because this is just one funny thing in one old version I'm ok with merging without fixing it. I will end up fixing it eventually though for some other books, I'm sure of it.

nik9000:
The diff took a while:
```
real	146m8.246s
```

It found 3 things:
* [ ] the funny thing I found before with the admonition in the id.
* [ ] "Alternative" language examples aren't being rendered properly.
* [ ] Floating titles with the xpack tag have their `class` attribute in the wrong place

nik9000:
The diff looks good. I'll merge later today when I can be sure I'll be around to support folks if something goes wrong.

PR #1567, opened 2019-12-10
Implement glossary style dlists in direct_html

nik9000:
This implements glossary style dlists in direct_html which is required
to convert the Elasticsearch reference (#1566).


nik9000:
Thanks @olksdr and @JoshMock ! I think I can get a diff for Elasticsearch now!

PR #1568, opened 2019-12-10
Strip html from titles in direct_html

nik9000:
If a page has extra html in the title we shouldn't add it to a
`title=""` tag because it'd be invalid. This stips it.


nik9000:
Thanks @olksdr !

PR #1569, opened 2019-12-11
Switch js client's docs to direct_html

nik9000:

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -487,6 +487,7 @@ contents:
                 index:      docs/index.asciidoc
                 tags:       Clients/JavaScript
                 subject:    Clients
+                direct_html: true
                 sources:
                   -
                     repo:   elasticsearch-js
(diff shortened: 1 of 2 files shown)


nik9000:
@delvedor, I'm creating this PR to track switching the elasticsearch-js docs to `--direct_html`. It'll build a diff and I'll review it. If it isn't good enough it'll iterate on `--direct_html` on my side until it is. Once it is ready we can switch. The goal is to make exactly the same html but do it faster.

nik9000:
The preview looks pretty ok to me. There are a whole bunch of explicit paragraph breaks that docbook was eating but that direct_html lets through and there is a difference in the html for the errors table in 16.x, but otherwise I think we're ok.

nik9000:
That table [looks](http://docs_1569.docs-preview.app.elstc.co/guide/en/elasticsearch/client/javascript-api/16.x/errors.html) ok to me.

nik9000:
OK! The diff is looking good so I'm going to merge this.

@delvedor, please use `--direct_html` to build the js api docs from here on out. I do plan to make it the default sometime in the next month, but for now you'll need to specify it to get the most accurate rendering of how the docs will look in production.

PR #1570, opened 2019-12-11
Switch the es ruby client to direct_html

nik9000:


nik9000:
This PR will make a preview which I'll check for accuracy. If everything is good I'll merge.

nik9000:
The diff here is empty. I think think we can merge once CI is happy.

PR #1572, opened 2019-12-11
Support ids on admonitions in direct_html

nik9000:
Adds support for ids on admonitions to direct_html in a way that lines
up with docbook. Required for the plugins and integrations guide
(#1564).


nik9000:
Thanks @olksdr !

PR #1573, opened 2019-12-11
Ignore funny quote in direct_html diff

nik9000:
This ignores `<span class="quote">` in the direct_html diffs. They don't
render differently and Asciidoctor doesn't make them so I don't believe
they are worth looking at.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -97,8 +97,8 @@ RUN rm -rf /var/log/nginx && rm -rf /run/nginx
 FROM base AS py_test
 RUN install_packages python3 python3-pip
 RUN pip3 install \
-  beautifulsoup4==4.7.1 \
-  lxml==4.3.1 \
+  beautifulsoup4==4.8.1 \
+  lxml==4.4.2 \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -97,8 +97,8 @@ RUN rm -rf /var/log/nginx && rm -rf /run/nginx
 FROM base AS py_test
 RUN install_packages python3 python3-pip
 RUN pip3 install \
-  beautifulsoup4==4.7.1 \
-  lxml==4.3.1 \
+  beautifulsoup4==4.8.1 \
```
nik9000: I need this to get `.smooth()`. Without smooth there is still a diff.

Dockerfile
```diff
@@ -97,8 +97,8 @@ RUN rm -rf /var/log/nginx && rm -rf /run/nginx
...
+  lxml==4.4.2 \
```
nik9000: I took this just because I was taking the other upgrade.

nik9000:
Thanks @olksdr !

PR #1574, opened 2019-12-11
Fix error in diff tool

nik9000:
Fixes a bug in the diff tool that causes it to crash when there is an
anchor tag inside a table without an id. Ooops.


nik9000:
Thanks @olksdr !

PR #1575, opened 2019-12-11
Switch the .net client to direct_html

nik9000:

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -530,6 +530,7 @@ contents:
                 tags:       Clients/.Net
                 subject:    Clients
                 chunk:      1
+                direct_html: true
                 sources:
                   -
                     repo:   elasticsearch-net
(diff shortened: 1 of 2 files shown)


nik9000:
This pull request will make a diff and I'll check the diff. If it looks good I'll make a comment on this PR and hopefully merge it. If it doesn't I'll iterate on `--direct_html` until it *does* look good.

nik9000:
Every single `<title>` tag has the wrong number of spaces between `and NEST:` and `the .NET`. All of them.

nik9000:
It looks like docbook squeezes spaces in the title. Easy enough to fix.

nik9000:
Now that I've fixed the titles I can read the diff and it is small enough that I was able to visually check everything. I think it is all "better" than it was before.

PR #1576, opened 2019-12-11
Switch the php client to direct_html

nik9000:

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -543,6 +543,7 @@ contents:
                 index:      docs/index.asciidoc
                 tags:       Clients/PHP
                 subject:    Clients
+                direct_html: true
                 sources:
                   -
                     repo:   elasticsearch-php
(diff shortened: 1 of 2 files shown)


nik9000:
This PR will make a diff and I'll check it. If the diff looks good I'll hopefully merge it. If it doesn't then I'll iterate on `--direct_html` until it does.

nik9000:
This one needs:
* [ ] Sidebars are missing ids.

There are some extra `<br>` tags but they are explicitly in the source, it is just that docbook was eating them.

nik9000:
> I'm not so familiar with `direct_html` parameter.

It's new. It speeds up building the doc by replacing xslt with a plugin to Asciidoctor. Removing xslt helps a ton from the "custsomizability" side too, but for the most part this change shouldn't make the docs *look* different.

Basically: you can use that new parameter when you build the docs now. Once I merge this then you *really* should use it.

nik9000:
Diff looks pretty good to me. Merging!

PR #1578, opened 2019-12-12
Switch logstash VPR to direct_html

nik9000:
direct_html is the future!


nik9000:
My diff spat out very little. Mostly some extra explicit `<br>` tags.

PR #1579, opened 2019-12-12
Clean up html in meta links in direct_html

nik9000:
Strips html from the `<link>` tag in the headings generated by direct
html. *and* squeezes extra spaces out of the titles as a side effect.

Required for .NET client (#1575).


nik9000:
Thanks @olksdr !

PR #1580, opened 2019-12-12
Force updating to a new version of yarn

nik9000:
We need at least 1.21.1 to fix
https://blog.npmjs.org/post/189618601100/binary-planting-with-the-npm-cli
. apt-get doesn't let us say "more than 1.21.1" so we have to pin it
*to* 1.21.1.


nik9000:
> Bummer you can't do `>=` but this works!

:+1:

PR #1581, opened 2019-12-12
Support id on sidebar in direct_html

nik9000:
Adds support for rendering the id on a sidebar in a docbook-like way to
direct_html.


nik9000:
:metal:

PR #1582, opened 2019-12-12
Improve rendering of callout lists in direct_html

nik9000:
When you write a callout list with duplicate entries we'd previously
duplicate a bunch of the callouts in strange ways. Both docbook and
direct_html had different, strange renderings. This changes direct_html
to have a fairly sane rendering. Say you write this, like we do in the
Elasticsearch reference:
```
      "tokens" : [ {
        "token" : "detail",
        "start_offset" : 0,
        "end_offset" : 8,
        "type" : "<ALPHANUM>",
        "position" : 0,
        "keyword" : false <1>
      }, {
        "token" : "output",
        "start_offset" : 9,
        "end_offset" : 15,
        "type" : "<ALPHANUM>",
        "position" : 1,
        "keyword" : false <1>
      } ]
```

The duplicate `<1>` elemenets now get a single callout which looks sort
of like:
```
<1> Output only "keyword" attribute, since specify "attributes" in
the request.
```

This is *much* cleaner than how docbook does it and how direct_html used
to do it.

### DIFF
--- a/resources/asciidoctor/lib/docbook_compat/convert_listing.rb
+++ b/resources/asciidoctor/lib/docbook_compat/convert_listing.rb
@@ -21,7 +21,9 @@ module DocbookCompat
       [
         '<div class="calloutlist">',
         '<table border="0" summary="Callout list">',
-        node.items.map { |item| convert_colist_item item },
+        node.items.each_with_index.map do |item, index|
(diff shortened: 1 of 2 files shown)


nik9000:
:squid:

PR #1584, opened 2019-12-12
Fixup the list titles in direct_html

nik9000:
This makes direct_html's rendering of the titles for ordered and
unordered lists line up with docbook's rendering.


nik9000:
1. better
2. lists
3. are
4. important!

PR #1585, opened 2019-12-12
Ignore unneeded tags in diffs for direct_html

nik9000:
Ignores two tags from in the html_diff. One of them docbook adds but
we don't need and another one asciidoctor adds that we're ok with it
adding.


nik9000:
:no_entry_sign: :eye:

PR #1586, opened 2019-12-13
Ignore some defaults in direct_html comparison

nik9000:
Docbook outputs `align=middle` and `align=center` for images but that is
how images look for us by default anyway and asciidoctor doesn't produce
those attributes. So this ignores them.


nik9000:
:no_entry_sign: :eyes:

PR #1588, opened 2019-12-13
Switch the curator docs to direct_html

nik9000:
direct_html is faster and more customizable. We're converting all of our
docs to it.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -617,6 +617,7 @@ contents:
             index:      docs/asciidoc/index.asciidoc
             tags:       Clients/Curator
             subject:    Clients
+            direct_html: true
             sources:
               -
                 repo:   curator
--- a/doc_build_aliases.sh
+++ b/doc_build_aliases.sh
@@ -70,7 +70,7 @@ alias docbldup='$GIT_HOME/docs/build_docs --doc $GIT_HOME/kibana/docs/uptime-gui
 alias docbldsec='$GIT_HOME/docs/build_docs --doc $GIT_HOME/stack-docs/docs/en/siem/index.asciidoc --chunk 1'
 
 # Curator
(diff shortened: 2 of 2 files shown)


nik9000:
I'm using this PR to make a preview. I'll check the preview for accuracy and, if it is accurate, merge this. Usually if the preview isn't perfect I'll hack on direct_html to make it perfect.

@untergeek this *shouldn't* change the docs at all, just build them faster. Once I've merged this PR you should add `--direct_html` when you build the docs.

nik9000:
There are some explicit `<hr>` and `<br>` tags that docbook was eating that have made it through. Those I think are ok. The footnotes render a little differently. This is also "ok".

I'll have to fix:
* [ ] `4.2/version-compatibility.html` isn't rendering right. Something is up with `&amp;emsp14;`.
* [ ] `3.3/indices-subcommand.html` is getting an invalid `title` attribute: `rel="prev" title="Command " tree=""`.

Those two should be pretty simple to fix.

nik9000:
> 4.2/version-compatibility.html isn't rendering right. Something is up with &amp;emsp14;.

The table is rendering visually just fine though. I think we're ok there.

untergeek:
I'm almost of a mind to remove 3.x and 4.x Curator docs. They are only for very old and no longer officially supported versions of Elasticsearch. If fixing those proves time consuming, I'm totally okay with that route. Let the odd use of 3.x or 4.x get picked up by the Wayback machine.

nik9000:
> When building locally for my own testing, will I add this `--direct_html` flag?

Yup! It'll be required for the next month or so. Once I finish the migration I'll memorialize that flag as a noop.

nik9000:
> I'm almost of a mind to remove 3.x and 4.x Curator docs.

I have to make the fix (#1592) anyway because some other page is *bound* to have this somewhere in our docs.

For what it is worth you can declare those old versions to be "not live" and we'll add a robots meta tag to make google stop indexing them. Check out the `live` attribute in conf.yaml if you are interested.

PR #1589, opened 2019-12-13
Switch cloud-on-k8s docs to direct_html

nik9000:
direct_html is faster and more customizable. We're switching all of our
docs to it.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -792,6 +792,7 @@ contents:
             branches:   [ master, 1.0, 0.9, 0.8 ]
             index:      docs/index.asciidoc
             chunk:      1
+            direct_html: true
             sources:
               -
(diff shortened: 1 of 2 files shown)


nik9000:
@alaudazzi, I'm using this PR to make a diff. I'll check the diff for accuracy and if everything is good I'll merge it. At that point folks should add `--direct_html` to the command line when building the cloud on k8s docs. There *shouldn't* be a difference in the docs. If there is I'll work on `--direct_html` so there isn't.

nik9000:
This one needs #1584. There are a bunch of explicit `<br>` tags that docbook was eating that this adds. If we don't like those then we should remove them from the source. Other than that the diff looks fine.

nik9000:
Diff looks good now!

PR #1590, opened 2019-12-13
Add a robots tag for Elastic Site Search

nik9000:
This adds a `<meta name="robot"` tag specifically for site search so we
*always* crawl the site, even if we've instructed other robots not to.
We'll use site search's customization to downplay the results for old
docs rather than excluding them entirely.

Here are the docs for the robots tag we use:
https://swiftype.com/documentation/site-search/crawler-configuration/meta-tags#swiftbot


nik9000:
Thanks @olksdr !

PR #1591, opened 2019-12-13
Make direct_html tables a little more docbook

nik9000:
Apparently docbook only uses the formal "table" wrapper when there is a
title and uses the informal "informaltable" wrapper even when the table
has an id. direct_html was using the formal wrapper when it *either* had
an id or a title. This fixes that.

Required by the Elaticsearch reference (#1566).


nik9000:
Thanks @olksdr !

PR #1592, opened 2019-12-13
Escape quotes in direct_html sibling links

nik9000:
We have `<link rel="next"` and `prev` and `up`, all of which have the
title of the target section. If the titles of those sections have a
quote in them we were emitting invalid html. This fixes that by escaping
the quotes as `&quot;`.


nik9000:
Thanks @olksdr !

PR #1593, opened 2019-12-13
Support inline xpack declaration in direct_html

nik9000:
Sometimes we declare the xpack tag inline like so:
```
[float]
=== [xpack]#Foo#
```

This renders the xpack tag in those cases.


nik9000:
Thanks @olksdr !

PR #1594, opened 2019-12-13
Ignore a few more differences in direct_html

nik9000:
Ignores the difference between `<p class="exclude exclude">` (docbook's
output) and `<p>` (asciidoctor's output). The `exclude` role doesn't
have any visual effect. And duplicate classes *never* have any visual
effect.

Required by the Elasticsearch reference (#1566).


nik9000:
Zoom!

PR #1595, opened 2019-12-13
Move id for admonitions in direct_html

nik9000:
Docbook puts ids *inside* the title of an admonition if there is a
title. This does that too.


nik9000:
Thanks @olksdr !

PR #1596, opened 2019-12-13
Ignore missing px in direct_html diff

nik9000:
Docbook doesn't add a `px` that direct_html does add. We're quite happy
to have the `px` but we don't need to see it on every time that gets it.
Thus, we ignore it in `html_diff`.


nik9000:
:heart:

PR #1597, opened 2019-12-13
Switch libbeat docs to direct_html

nik9000:

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -932,6 +932,7 @@ contents:
             tags:       Libbeat/Reference
             subject:    libbeat
             respect_edit_url_overrides: true
+            direct_html: true
             sources:
               -
                 repo:   beats
--- a/doc_build_aliases.sh
+++ b/doc_build_aliases.sh
@@ -87,7 +87,7 @@ alias docbldecctl='$GIT_HOME/docs/build_docs --direct_html --doc $GIT_HOME/ecctl
 alias docbldk8s='$GIT_HOME/docs/build_docs --direct_html --doc $GIT_HOME/cloud-on-k8s/docs/index.asciidoc --chunk 1'
 
 # Beats
-alias docbldbpr='$GIT_HOME/docs/build_docs --respect_edit_url_overrides --doc $GIT_HOME/beats/libbeat/docs/index.asciidoc --chunk 1'
+alias docbldbpr='$GIT_HOME/docs/build_docs --direct_html --respect_edit_url_overrides --doc $GIT_HOME/beats/libbeat/docs/index.asciidoc --chunk 1'
 
 alias docbldbdg='$GIT_HOME/docs/build_docs --respect_edit_url_overrides --doc $GIT_HOME/beats/docs/devguide/index.asciidoc --chunk 1'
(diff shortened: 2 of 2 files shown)


nik9000:
@dedemorton, I'm using this to generate a diff. I'll comment with how it goes. If it isn't perfect I *should* be able to modify `--direct_html` until it is good.

nik9000:
The diff contains two things. This one is fine with me, I think:
```
raw/en/beats/libbeat/7.5/toc.html
--- /dev/fd/63
+++ raw/en/beats/libbeat/7.5/toc.html
@@ -1120,7 +1120,7 @@
     <li>
      <span class="appendix">
       <a href="contributing-to-beats.html">
-       A. Contributing to Beats
+       Contributing to Beats
       </a>
      </span>
     </li>
```

And this one I'm going to fix:
```
raw/en/beats/libbeat/7.x/community-beats.html
--- /dev/fd/63
+++ raw/en/beats/libbeat/7.x/community-beats.html
@@ -90,11 +90,8 @@
    </p>
    <p>
     Want to contribute? See
-    <a class="link" href="contributing-to-beats.html" title="Appendix A: Contributing to Beats">
-     Appendix A,
-     <em>
-      Contributing to Beats
-     </em>
+    <a class="link" href="contributing-to-beats.html" title="Contributing to Beats">
+     Appendix A
     </a>
     .
    </p>
```

dedemorton:
@nik9000 I'm totally Ok with dropping the appendix numbering. It's there mainly because I wanted to include a one-pager at the end of the book, and that doesn't work with parts unless you make the page an appendix.

nik9000:
> @nik9000 I'm totally Ok with dropping the appendix numbering. It's there mainly because I wanted to include a one-pager at the end of the book, and that doesn't work with parts unless you make the page an appendix.

Yeah, that last link was just coming out as "Appendix A" without the "Contributing to Beats" part. But I've fixed it. I'll regenerate the diff. I'd love to merge this if the diff is empty.

Can you coordinate with the beats folk to update the build to include the new flag on this book? I'm expecting it to become the default in less than a month but for now I think it is nice to keep it lined up.

dedemorton:
> Can you coordinate with the beats folk to update the build to include the new flag on this book? 

Yes, but because of how are `make docs` command works, we'll need to make all the books compliant before I update the script.

nik9000:
> Yes, but because of how are `make docs` command works, we'll need to make all the books compliant before I update the script.

Got it! I'll hold off on merging these until they are *all* good.

dedemorton:
> Got it! I'll hold off on merging these until they are all good.

Not sure you need to wait unless merging will break the older command.

nik9000:
> Not sure you need to wait unless merging will break the older command.

It won't break the old command but there is a chance that running the old command won't produce the right output. But we have the universal PR tests so that is probably not a problem....

dedemorton:
> It won't break the old command but there is a chance that running the old command won't produce the right output

Yah, I think that's OK. I'll bet that no one even notices.

nik9000:
Sounds good!

PR #1598, opened 2019-12-13
Decorate link to appendix in direct_html

nik9000:
When we link to an appendix we docbook adds `Appendix A,` to the text.
This makes direct_html do that too.


nik9000:
Thanks @olksdr !

PR #1599, opened 2019-12-13
Fix examples in direct_html

nik9000:
Formal examples were rendering their titles in a docbook incompatible
way. This fixes that.


nik9000:
Thanks @olksdr !

PR #1602, opened 2019-12-13
Switch filebeat docs to direct_html

nik9000:


nik9000:
@dedemorton, another one to make a diff.

nik9000:
@dedemorton the diff for this one also look good. It is ready whenever you are.

PR #1603, opened 2019-12-13
Support roles on paragraphs in direct_html

nik9000:
Docbook copies the `role` on paragraphs into the class. This implements
that in direct_html.

Required for beats.


nik9000:
Thanks @olksdr !

PR #1604, opened 2019-12-13
Switch winlogbeat to direct_html

nik9000:
direct_html is the future!


nik9000:
The diff on this one also looks good. Ready when you are, @dedemorton !

PR #1605, opened 2019-12-13
Switch metricbeat to direct_html

nik9000:
direct_html is the future!


nik9000:
@dedemorton the diff looks ok. This one is ready when you are!

PR #1609, opened 2019-12-13
Switch journalbeat to direct_html

nik9000:
direct_html is the future!


nik9000:
The diff for this one also looks good. @dedemorton , this is ready when you are.

PR #1610, opened 2019-12-13
Switch the definitive guide to direct_html

nik9000:
direct_html is the future! The definitive guide is the past....

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -1910,6 +1910,7 @@ contents:
             tags:       Legacy/Elasticsearch/Definitive Guide
             subject:    Elasticsearch
             suppress_migration_warnings: true
+            direct_html: true
             sources:
               -
                 repo:   guide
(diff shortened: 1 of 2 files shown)


nik9000:
There are quite a few small spacing differences that the diff tool doesn't seem to ignore for some reason. There are a couple of other small changes that I don't think are worth fixing up because this is a legacy book. I do think there is one thing I need to fix::
* [ ] `elasticsearch/guide/1.x/index.html` is missing `authorgroup`. Without this we don't mention the authors at all.

nik9000:
>     * `elasticsearch/guide/1.x/index.html` is missing `authorgroup`. Without this we don't mention the authors at all.

We had added this with a docinfo.xml file. I pushed a titlepage-extra file.

nik9000:
Diff looks good now. Merging.

PR #1611, opened 2019-12-13
Switch the legacy xpack reference to direct_html

nik9000:
direct_html is the future! The xpack reference is a thing of the past.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -1884,6 +1884,7 @@ contents:
             index:      docs/en/index.asciidoc
             private:    1
             branches:   [ 6.2, 6.1, 6.0, 5.6, 5.5, 5.4, 5.3, 5.2, 5.1, 5.0 ]
+            direct_html: true
             sources:
               -
                 repo:   x-pack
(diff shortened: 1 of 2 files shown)


nik9000:
Look like we generate links to some of these images:
```
-     <img alt="Node View" src="images/monitoring/monitoring-kibana-instances.jpg"/>
+     <a class="image" href="images/monitoring-kibana-instances.jpg">
+      <img alt="Node View" src="images/monitoring/monitoring-kibana-instances.jpg"/>
+     </a>
```

I'm fine making the images clickable but the url should be right.....

nik9000:
That happens, on, for example, `5.3/monitoring-logstash.html`

nik9000:
These seem to be caused by code that looks like:
```
image::monitoring/images/monitoring-kibana-instances.jpg["Kibana Instances",link="images/monitoring-kibana-instances.jpg"]
```

nik9000:
So I dug into this book and found that we were generating totally the wrong `img` tags which https://github.com/elastic/docs/pull/1630 fixes. Sadly, the `link` bit of these images is actually something that we requested in the source and the link is *wrong*. This doesn't fail the link checker because that only checks absolute links to `/` or to `www.elastic.co` and this is a link relative to the page. So these links are bad. If the docs were current I'd say we should fix them. But they aren't so I'm ok with it. I guess.....

nik9000:
The diff for this one looks fine now. Modulo some of those image links I mentioned before.

PR #1612, opened 2019-12-13
Switch legacy infrastructure guide to direct_html

nik9000:
direct_html is the future!


nik9000:
This one's diff is empty so I'm converting it.

PR #1613, opened 2019-12-13
Switch the groovy api to direct_html

nik9000:
direct_html is the future!


nik9000:
The diff is empty so I merged!

PR #1620, opened 2019-12-15
Switch shield to direct_html

nik9000:
direct_html is the future.


nik9000:
The failure here is very similar to #1626, though maybe I can just fix it by hand. I'm not sure.

nik9000:
#1648 should fix this.

nik9000:
> #1648 should fix this.

Now that that is in the diff looks good!

PR #1625, opened 2019-12-15
Switch siem docs to direct_html

nik9000:


nik9000:
The diff looks good to me. I can merge this anytime you are ready @karenzone.

nik9000:
Thanks @karenzone and @benskelker !

PR #1626, opened 2019-12-16
Drop admonitions from generated ids in direct_html

nik9000:
When a section or a floating title doesn't have an id asciidoctor
generates one. When that section includes an admonition it *was*
included in the generated id. Docbook also generates ids but doesn't
include the admonition, which makes a ton of sense. So this modifies
direct_html not to generate the admonition into the id. *how* we do that
is a little complex because of the extension points that asciidoctor
gives us:

1. If we're generating an admonition inside of a section that doesn't
yet have an id then return an empty string. This should make the
admonition a noop when generating the title used to generate the
automatic id and *only* then. All other times the section has an id.
2. Before actually converting the document to html we clear that cached
copy of the generated title. Now that the section has an id the next
time we regenerate its title it'll get generated correctly.

### DIFF
--- a/resources/asciidoctor/lib/docbook_compat/clear_cached_titles.rb
+++ b/resources/asciidoctor/lib/docbook_compat/clear_cached_titles.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+require_relative '../scaffold'
+
+module DocbookCompat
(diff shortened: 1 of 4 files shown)


nik9000:
Thanks @olksdr !

PR #1627, opened 2019-12-16
Fix a bug in direct_html' floating titles

nik9000:
direct_html's floating titles were rendering classes *outside* of the
tag. Oops. This fixes that.


nik9000:
Thanks @olksdr !

PR #1628, opened 2019-12-16
Support alternative examples in direct_html

nik9000:
This adds support for examples in alternative languages to direct_html.
The alternative examples plugin works by adding a bunch of extra
listings and callout lists to the parse tree. This makes sure that they
are all converted properly, including updating the integration test for
them to work with `--direct_html`.


nik9000:
Thanks @olksdr !

PR #1630, opened 2019-12-16
Fix support for imagesdir

nik9000:
This fixes our image copier to properly copy images referenced by
`imagesdir`. We *were* copying the images correctly be we weren't
copying them to the right spot. This is because our image copier does a
`find | grep <path>` to find the image for compatibility with the old
`a2x` utility that we used to build the asciidoc before asciidoctor. We
just weren't getting the paths right.

Here is some documentation for the attribute:
https://asciidoctor.org/docs/user-manual/#setting-the-location-of-images

### DIFF
--- a/resources/asciidoctor/lib/copy_images/copier.rb
+++ b/resources/asciidoctor/lib/copy_images/copier.rb
@@ -57,7 +57,7 @@ module CopyImages
         to_check += subdirs(dir)
       end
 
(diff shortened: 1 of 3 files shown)


nik9000:
The failure is because I added something to the warning. Let me fix that up....

nik9000:
Thanks @olksdr !

PR #1631, opened 2019-12-17
Switch Chinese translations to direct_html

nik9000:
direct_html is the future.


nik9000:
Oh boy.

```
00:35:29 INFO:build_docs:/var/lib/gems/2.5.0/gems/asciidoctor-2.0.10/lib/asciidoctor/cli/options.rb:87:in `encode': "\xE8" from ASCII-8BIT to UTF-8 (Encoding::UndefinedConversionError)
```

nik9000:
There are quite a few small differences in the guide but I think that is ok given how out of date it is anyway.

PR #1634, opened 2019-12-17
Switch RUM apm agent to direct_html

nik9000:

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -1585,6 +1585,7 @@ contents:
                     tags:       APM Real User Monitoring JavaScript Agent/Reference
                     subject:    APM
                     chunk:      1
+                    direct_html: true
                     sources:
                       -
                         repo:   apm-agent-rum-js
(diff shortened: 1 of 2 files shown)


nik9000:
@bmorelli25, I'm using this PR to make a preview which I'll check for accuracy.

bmorelli25:
Is the [diff](http://docs_1634.docs-preview.app.elstc.co/diff) supposed to be that big? I've been clicking around a bit, but can't see anything that looks wrong.

nik9000:
The diff from the preview is a line by line diff. My diff that I make on my
laptop parses the html then runs the diff. So, yeah, I expect a big diff in
the preview but for the one I run to be much smaller.

nik9000:
When I run the diff with the [diff tool](https://github.com/elastic/docs/blob/master/integtest/html_diff) it is empty. But the tool has a bunch of "normalize the html" stuff that the docs-preview's diff doesn't have.

nik9000:
So I'm ready to merge this anytime you are @bmorelli25 .

bmorelli25:
Opened https://github.com/elastic/apm-agent-rum-js/pull/538 to update the build script in the RUM repo.

PR #1635, opened 2019-12-17
Switch ruby apm agent to direct_html

nik9000:

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -1567,6 +1567,7 @@ contents:
                     tags:       APM Ruby Agent/Reference
                     subject:    APM
                     chunk:      1
+                    direct_html: true
                     sources:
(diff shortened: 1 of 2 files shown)


nik9000:
@bmorelli25 I'll leave a comment once I have a chance to analyze the diff this one makes.

nik9000:
@bmorelli25 the diff for this one contains only some new explicit `<br>` tags that were being eaten by docbook but make it through with direct_html. You can see them [here](http://docs_1635.docs-preview.app.elstc.co/guide/en/apm/agent/ruby/2.x/opentracing.html#operation-modes)

nik9000:
I'm been letting those new `<br>` tags through. So I'm ready to merge this anytime.

PR #1640, opened 2019-12-17
Switch .NET apm agent to direct_html

nik9000:


nik9000:
This one has a problem with admonitions being included in `title` tags. I'll open a PR to remove them today if possible.

nik9000:
I've merged a fix for the titles and regenerated. Now the diff looks good. This is ready when you are @bmorelli25 .

PR #1641, opened 2019-12-17
Switch Java apm agent to direct_html

nik9000:


nik9000:
This one also has admonitions that sneak into `title`s. I'll fix it.

nik9000:
I've merged the fix for the title and now the diff looks good! This is ready when you are @bmorelli25 .

PR #1644, opened 2019-12-17
Switch apm getting started docs to direct_html

nik9000:


nik9000:
The diff for this one is empty so it is ready whenever you are, @bmorelli25.

PR #1645, opened 2019-12-17
Switch kibana to direct_html

nik9000:
direct_html is fater and more cutomizable. It is the future!


nik9000:
Retest this please

nik9000:
The diff looks good! I'll merge in the morning so I can be around in case I need to revert.

PR #1648, opened 2019-12-17
Drop admonitions from titles

nik9000:
This removes admonitions from the `<title>` and `title=""` tags. It does
this by hacking it out with regexes which is pretty lame, but
asciidoctor doesn't give us another extension point.


nik9000:
:cry:

PR #1650, opened 2019-12-17
Switch cloud's heroku docs to direct_html

nik9000:
`--direct_html` is faster and more customizable. It is the future.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -691,6 +691,7 @@ contents:
             chunk:      1
             noindex:    1
             private:    1
+            direct_html: true
             sources:
               -
                 repo:   cloud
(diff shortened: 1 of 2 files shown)


nik9000:
@nrichers I'm using this to make a preview and a diff. I'll leave a comment once I know how the diff looks.

nik9000:
The diff looks ok. There is a small thing but I don't think it is big enough to block merging. I'll fix it anyway though:
```
raw/en/cloud-heroku/release-ms-31/ech-users-manage-x-pack.html
--- docbook
+++ direct_html
@@ -87,7 +87,8 @@
    </p>
    <p>
     <span class="image">
-     <img alt="The admin user in X-Pack" src="images/x-pack-users.png"/>
+     <img alt="The &lt;code class=" literal="" src="images/x-pack-users.png"/>
+     admin user in X-Pack"&gt;
     </span>
    </p>
    <p>
```

PR #1652, opened 2019-12-17
Switch cloud enterprise docs to direct_html

nik9000:


nik9000:
@nrichers, I'm using this to generate a preview and a diff. I'll leave a comment once I've looked over the diff.

nik9000:
I'm building the diff for this one too. So far it only has the same alt tag.

nik9000:
@nrichers this diff looks good too!

nik9000:
I talked to @nrichers out of band and he urged me to merge. So here goes!

PR #1654, opened 2019-12-17
Add cloud-assets to conf.yaml and build aliases for ECE

nrichers:
We're moving to a more sustainable way of maintaining current Docker image info for ECE that requires pulling from the `master` branch of the elastic/cloud-assets repo. This PR makes the necessary doc build updates.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -11,6 +11,7 @@ repos:
     beats:                https://github.com/elastic/beats.git
     clients-team:         https://github.com/elastic/clients-team.git
     cloud:                https://github.com/elastic/cloud.git
+    cloud-assets:         https://github.com/elastic/cloud-assets.git
     cloud-on-k8s:         https://github.com/elastic/cloud-on-k8s.git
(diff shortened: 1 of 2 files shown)


conf.yaml
```diff
@@ -750,6 +751,17 @@ contents:
                 repo:   cloud
                 path:   docs/shared
                 exclude_branches: [ 1.0 ]
+              -
+                repo:   cloud-assets
+                path:   docs
+                map_branches: &mapCloudEceToCloudAssets
```
nik9000: You only need to add the ` &mapCloudEceToCloudAssets` if you will reuse the source.

nik9000:
The failure looks like it is caused because that repo doesn't contain that path.

nrichers:
jenkins test this please

nrichers:
> The failure looks like it is caused because that repo doesn't contain that path.

Yes, there's https://github.com/elastic/cloud-assets/pull/315 that adds the files that we need.

PR #1656, opened 2019-12-18
Switch "all books" integ tests to direct_html

nik9000:
Switch the "all books" integration tests to direct_html now that almost
all of the books have switched.


nik9000:
Thanks @olksdr !

PR #1659, opened 2019-12-18
Only build with `--direct_html`

nik9000:
This removes support for building docs without direct_html. You no
longer need to specify `--direct_html` on the command line or
`direct_html` in conf.yaml for it ot build with direct_html.

This doesn't drop the configuration from conf.yaml or the change
tracker or all of the files that we need to build with docbook. We'll do
that in a followup.

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -69,6 +69,12 @@ The Asciidoctor migration is complete! --asciidoctor will emit this message
 forever in honor of our success but otherwise doesn't do anything.
 MSG
 }
+if ( $Opts->{direct_html} ) {
+   say <<MSG
+The direct_html migration is complete! --direct_html will emit this message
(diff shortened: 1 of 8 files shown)


nik9000:
This needs to wait on #1559.

build_docs.pl
```diff
@@ -69,6 +69,12 @@ BEGIN
 forever in honor of our success but otherwise doesn't do anything.
 MSG
 }
+if ( $Opts->{direct_html} ) {
+   say <<MSG
+The direct_html migration is complete! --direct_html will emit this message
+forever in honor of our success but otherwise doesn't do anything.
+MSG
+}
```
JoshMock: 😍

nik9000: I don't want to drop the parameter or else it'll break everyone that added it1

nik9000:

PR #1661, opened 2019-12-18
Clean up the chunker test

nik9000:
Drops some "funny" `gsub` calls from the chunker test in favor of being
more explicit.


nik9000:
Zoom!

PR #1664, opened 2019-12-19
Share a test utility

nik9000:
This moves the `file_exist` matcher into a place so the Asciidoctor
plugin unit tests and the integration tests can share it.


nik9000:
Thanks @olksdr !

PR #1665, opened 2019-12-19
Fix remaining rubocop

nik9000:
This fixes all of the remaining rubocop warnings in our code and removes
the suppresions. The last suppression is for the docbook45 converter
which mostly comes from Asciidoctor's code base and we're in the process
of removing it.


nik9000:
:robot: :policewoman:

PR #1666, opened 2019-12-19
Move ruby style helper

nik9000:
Move the ruby style helper into the `resources` directory so we don't
have so many top level directories.


nik9000:
:man_dancing:

PR #1669, opened 2019-12-19
Unify some test dsl

nik9000:
Moves `file_context` into a shared space because we use it in both the
unit tests and integration tests.


nik9000:
Thanks @olksdr !

PR #1670, opened 2019-12-19
Remove docbook

nik9000:
Removes all the docbook dependencies now that we don't need them any
more.


nik9000:
Needs #1670.

nik9000:
I've merged #1659 and included this in the PR. Let's see if it works now!

PR #1672, opened 2019-12-20
Vanish ugly grey bars unless needed

nik9000:
Switches the overflow strategy for tables from "always scroll" to
"scroll if needed".


nik9000:
Thanks @webmat !

PR #1673, opened 2019-12-20
Drop docbook dependencies from docker

nik9000:
We don't depend on dockbook any more so we don't need its dependencies.


nik9000:
Thanks @olksdr !

PR #1674, opened 2019-12-20
Drop direct_html from conf.yaml

nik9000:
It is the default now.


nik9000:
This should force a rebuild of everything. Which is fine. I'll recheck the diff to make sure it isn't crazy while I'm at it. But we *can't* build with docbook any more. All the xslt is gone!

nik9000:
Change list look good.

PR #1675, opened 2019-12-20
Drop environment customization for xml

nik9000:
Since we don't depend on docbook we don't need to set up the environment
for xml any more.

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -726,10 +726,6 @@ sub push_changes {
 #===================================
 sub init_env {
 #===================================
-    $ENV{PATH}
-        . ":$FindBin::RealBin:"
-        . $ENV{PATH};
-
     if (exists $ENV{SSH_AUTH_SOCK}
             && $ENV{SSH_AUTH_SOCK} eq '/tmp/forwarded_ssh_auth') {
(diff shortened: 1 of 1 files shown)


build_docs.pl
```diff
@@ -726,12 +726,7 @@ sub push_changes {
 #===================================
 sub init_env {
 #===================================
-    $ENV{SGML_CATALOG_FILES} = $ENV{XML_CATALOG_FILES} = join ' ',
-        file('resources/docbook-xsl-1.78.1/catalog.xml')->absolute,
-        file('resources/docbook-xml-4.5/catalog.xml')->absolute;
-
     $ENV{PATH}
-        = dir('resources/asciidoc-8.6.8/')->absolute
         . ":$FindBin::RealBin:"
```
olksdr: should be a `=` sign here maybe?

nik9000: Yeah! I wonder why the tests passed.....

nik9000: I've entirely dropped the path customizations. I guess we didn't need them!

nik9000:

Just for a little extra paranoia.

nik9000:
Dropping the path customization did the trick.

PR #1677, opened 2019-12-20
Remove change tracking code for direct_html

nik9000:
Removes all of the "change tracking" code for `--direct_html`. It is the
default now.


nik9000:
Thanks @olksdr !

PR #1678, opened 2019-12-20
Drop remaining tests for direct_html

nik9000:
It is now the default so there is no need to test it explicitly.


nik9000:
Thanks @olksdr !

PR #1679, opened 2019-12-20
Temporarily change the default banner text for old versions.

debadair:
Changing the default "old version" banner until @nik9000 merges a change to enable a different default for live versions older than current (actively maintained versions).

Once we rebuild, this will replace the erroneous out of maintenance message on the Cloud docs. OOM stack docs will also get the "milder" message until Nik's fix is merged and this is reverted, but we're okay with that.


debadair:
image

PR #1680, opened 2019-12-20
Default to "non-live" header on non-live old books

nik9000:
If a book is not `live` and is "older" than the current book this
defaults its header to an "end of life" style message.

### DIFF
--- a/lib/ES/Book.pm
+++ b/lib/ES/Book.pm
@@ -15,7 +15,12 @@ use utf8;
 our %Page_Header = (
     en => {
         old => <<"HEADER",
-A newer version is available. For the latest information, see the 
+A newer version is available. For the latest information, see the
+<a href="../current/index.html">current release documentation</a>.
+HEADER
+        dead => <<"HEADER",
+<strong>IMPORTANT</strong>: No additional bug fixes or documentation updates
+will be released for this version. For the latest information, see the
 <a href="../current/index.html">current release documentation</a>.
 HEADER
         new => <<"HEADER"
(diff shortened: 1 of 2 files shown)


lib/ES/Book.pm
```diff
@@ -27,15 +32,22 @@ HEADER
     zh_cn => {
         old => <<"HEADER",
 你当前正在查看的是旧版本的文档。如果不是你要找的，请点击查看 <a href="../current/index.html">当前发布版本的文档</a>。
+HEADER
+        dead => <<"HEADER",
+你当前正在查看的是旧版本的文档。如果不是你要找的，请点击查看 <a href="../current/index.html">当前发布版本的文档</a>。
```
nik9000: This is the same as the `old` message.

lib/ES/Book.pm
```diff
@@ -27,15 +32,22 @@ HEADER
...
     ja => {
         old => <<"HEADER",
-You are looking at documentation for an older release.
-Not what you want? See the
+A newer version is available. For the latest information, see the
+<a href="../current/index.html">current release documentation</a>.
+HEADER
+        dead => <<"HEADER",
+<strong>IMPORTANT</strong>: No additional bug fixes or documentation updates
+will be released for this version. For the latest information, see the
```
nik9000: These were never translated.

nik9000:

debadair:
If I'm reading this right, it does what we discussed--attaches the "newer version" notice to live versions older than current, and attaches the OOM notice to not-live versions older than current. 

We really have three states if version < current:
- live (maintained version)
- not live, not EOL (out of maintenance)
- not live, EOL (end of life)

If we could specify an additional "eol" parameter the way we do "live", we could totally automate the addition of the notices, instead of manually having to add the page headers to the affected books when things go EOL--update conf.yaml and you're done. 

Just sayin'...

nik9000:
You mean the version who's text is:
```
<p>
  <strong>WARNING</strong>: Version {version} of {product} has passed its 
  <a href="https://www.elastic.co/support/eol">EOL date</a>. 
</p>  
<p>
  This documentation is no longer being maintained and may be removed. 
  If you are running this version, we strongly advise you to upgrade. 
  For the latest information, see the 
  <a href="../current/index.html">current release documentation</a>. 
</p>
```
?

I think maybe we should get this one in as is and then think about if we should rename `live` to `maintained` and add a `live` or `eol` list. Or something like that.

nik9000:
Thanks @olksdr !

PR #1682, opened 2019-12-23
 Begin dropping docbook support from plugins

nik9000:
This starts dropping docbook support from our Asciidoctor plugins and
their tests. The tests end up being a lot more code to get right. But
now that we don't use docbook it is silly for us to make it and test
that we make it.


nik9000:
Thanks @olksdr ! I'll see about dropping the rest of the docbook!

PR #1683, opened 2019-12-23
Drop docbook 4.5 from plugins

nik9000:
Drops support for docbook 4.5 from our Asciidoctor plugins, including
the docbook 4.5 converter that we pulled from an older version of
Asciidoctor's source and customized to work with the modern version of
Asciidoctor.

The majority of the changes here are in the tests to assert against HTML
rather than docbook. I also moved the testing for the custom admonitions
from the docbook compat plugin's tests into the custom admonition
plugin's tests. You still need docbook compat to render the custom
admonition plugins though. But without that move the custom admonition
plugins wouldn't have *any* tests. Or, at least, they wouldn't look like
they had any tests.


nik9000:
@elasicmachine, run elasticsearch-ci/docs

nik9000:
Thanks @olksdr ! I think that is it!

PR #1687, opened 2020-01-02
Add support for "collapsible" examples

### LINKED ISSUE #1279: Add support for collapsible sections
To improve our breaking changes documentation, the Elasticsearch writers would like to use collapsible sections in our HTML output.

As of writing, it doesn't appear this is supported. I've documented my attempts to implement collapsible sections below.

### Option 1: Use the Asciidoctor`[%collapsible]` block
I first tried to use the [`[%collapsible]` block](https://mrhaki.blogspot.com/2019/03/awesome-asciidoctor-collapsible-content.html) added in Asciidoctor 2.0:

```
[%collapsible]
.Details
====
Content to collapse.
====
```

However, our docs build uses Asciidoctor 1.5 which doesn't support this block type.

### Option 2: Use an html passthrough

I then tried to use passthrough blocks to use HTML details and summary tags.

```
++++
<details><summary>
Details
</summary><div>
++++

Content to collapse.

++++
</div></details> 
++++
```

However, docbook doesn't allow us to pass these HTML elements on directly. After looking at the docbook DTD, I don't believe any collapsible elements are supported.

nik9000:
This adds support for collapsible examples to our docs. It turns out
that asciidoctor just uses `<details>` and `<summary>` which isn't
supported by all browsers, but we've got a polyfill.

Closes #1279

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -1274,6 +1274,25 @@ in an error.  To resolve this, add newlines
 between the delimiter and the content
 before and after it.
 
+Examples can be made collapsible:
+
+[source,asciidoc]
+----------------------------------
+[%collapsible]
+.An example
+====
+Lots of text can go in here.
+====
+----------------------------------
+
+Which renders as:
+
+[%collapsible]
+.An example
+====
+Lots of text can go in here.
+====
+
 [[includes]]
 == Including files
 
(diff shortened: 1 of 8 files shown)


nik9000:
Thanks @JoshMock !

PR #1705, opened 2020-01-21
Add Github link support to beta and experimental macros

jrodewig:
We add the `experimental::[]` and `beta::[]` macros to document features that are not yet ready for GA. However, these tags don't currently indicate what steps are needed to get a feature to GA.

With these changes, users can include a GitHub issue for the feature in the `experimental::[]` or `beta::[]`. When provided, a concise sentence and link is generated for the issue.

Relates to https://github.com/elastic/elasticsearch/issues/51250

### Examples
http://test__care-admon-links.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/_beta_experimental_macro_examples.html

### DIFF
--- a/resources/asciidoctor/lib/care_admonition/extension.rb
+++ b/resources/asciidoctor/lib/care_admonition/extension.rb
@@ -42,24 +42,34 @@ class CareAdmonition < Asciidoctor::Extensions::Group
       @default_text = default_text
     end
 
-    def generate_issue_text(text, issue_url)
-      issue_no = issue_url.split('/').last.chomp('/')
-      issue_text = <<~TEXT
-        For feature status, see #{issue_url}[\##{issue_no}].
-      TEXT
-      text + ' ' + issue_text
-    end
-
     def generate_text(text, issue_url)
-      if text&.start_with?('http')
+      if text&.start_with?('http', '{issue}')
         issue_url = text
         text = @default_text
       else
         issue_url = issue_url
         text ||= @default_text
       end
-      text = generate_issue_text(text, issue_url) if issue_url
+      text = add_issue_text(text, issue_url) if issue_url
       text
+    end
+
+    def add_issue_text(text, issue_url)
+      issue_num = get_issue_num(issue_url)
+      issue_text = <<~TEXT
+        For feature status, see #{issue_url}[\##{issue_num}].
+      TEXT
+      text + ' ' + issue_text
+    end
+
+    def get_issue_num(issue_url)
+      if issue_url.start_with?('http')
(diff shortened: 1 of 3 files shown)


resources/asciidoctor/lib/care_admonition/extension.rb
```diff
@@ -34,7 +34,7 @@ def activate(registry)
   # Block care admonition.
   class ChangeAdmonitionBlock < Asciidoctor::Extensions::BlockMacroProcessor
     use_dsl
-    name_positional_attributes :passtext
+    name_positional_attributes :passtext, :github
```
nik9000: I'd call it `issue_url` instead.

resources/asciidoctor/lib/care_admonition/extension.rb
```diff
@@ -43,9 +43,25 @@ def initialize(role, default_text)
     end
 
     def process(parent, _target, attrs)
+      text = attrs[:passtext]
```
nik9000: That is super kind of you!

I think it's be cool just to check if it starts with http.  No need for the rest, I think.

jrodewig:
Thanks for taking a look @nik9000.

While I was rewriting this to pass the Rubocop tests, I decided to add support for `{issue}` links as well. There are also unit tests for everything in`care_admonition_spec.rb`. It was fun learning more about Ruby and Rspec!

resources/asciidoctor/lib/care_admonition/extension.rb
```diff
@@ -34,19 +34,48 @@ def activate(registry)
...
+    end
+
+    def get_issue_num(issue_url)
+      if issue_url.start_with?('http')
+        issue_num = issue_url.split('/').last
+        issue_num.chomp!('/')
+      else
+        issue_num = issue_url.sub('{issue}', '')
+      end
+      issue_num
```
nik9000: I believe it is a little more idiomatic in ruby to do
```
    def get_issue_num(issue_url)
      if issue_url.start_with?('http')
        issue_url.split('/').last.issue_num.chomp('/')
      else
        issue_url.sub('{issue}', '')
      end
    end
```

Though I tend to prefer "early returns" from my time doing java:
```
  def get_issue_num(issue_url)
     return issue_url.split('/').last.chomp('/') if issue_url.start_with?('http')
     issue_url.sub('{issue}', '')
  end
```

jrodewig: Thanks @nik9000. I'll go with the second option. Rubocop seems fine with it.

Issue #1629, opened 2019-12-16
Make the Edit button work for Cloud docs! 


nrichers:
Enabling the **Edit** button for Cloud docs would be easy, if it weren't for the fact that those docs today live in a private repo. But there are still many people who do have access to the elastic/cloud repo, and so ... 

... if we could figure out a way to enable the Edit button for those who do have access to the elastic/cloud repo, that would make quick docs changes so much simpler, not to mention that it would remove the Cloud writers as gatekeepers from the process. 

Two possible options: 
* Fancy: Figure out if you are a part of the Elastic org and display the **Edit** button if you are. I would think that if you use the new Elastic Cloud home page and are authenticated, we'd know who you are. Essentially, we'd be enabling the Edit button if you made it past SSO. 
* Minimally viable: Just enable the **Edit** button but mark it as Elastic-only or restricted with an icon or text if it's displayed on a Cloud docs page. Could be a simple as this:
   
   ![image](https://user-images.githubusercontent.com/15148011/70946501-17d8c500-200c-11ea-809b-443b04e69f5f.png)

@ppf2 @Leaf-Lin @nik9000 FYI

(I hope this is in the right repo, since we're talking about docs in a private repo. If this is not acceptable, let me know and I will move the issue.)

ppf2:
LOVE IT ❤️

nrichers:
@kellyemurphy raised a good point today: we also have the tooltip we can update, so that it clearly indicates you need to have access to the closed elastic/cloud repo:

![image](https://user-images.githubusercontent.com/15148011/71288478-e7e93480-231f-11ea-80b1-0e9c6808c677.png)

We could say something like `Requires access to edit this page on GitHub` or something like this.

PR #1719, opened 2020-01-31
Enable custom "edit" icon for private repos

nyurik:
If a repository is marked as private,
show a custom icon and tooltip, explaining
that this page is editable only by authorized users.

This change also add support for building private single-books
with the `--private` CLI parameter.

Resolves https://github.com/elastic/docs/issues/1629

For everyone else, it will result in 404.


build_docs.pl
```diff
@@ -946,6 +947,7 @@ sub usage {
                             Suppress warnings about Asciidoctor migration
                             issues. Use this when building "old" branches.
           --toc             Include a TOC at the beginning of the page.
+          --private         Indicate that the github repo is private.
```
nrichers: Perhaps say ` Indicate that editing on GitHub is for a private repo.`?

resources/asciidoctor/lib/edit_me/extension.rb
```diff
@@ -74,8 +74,16 @@ def link_for(block)
       url = edit_url block
       return '' unless url
 
+      if block.document.attr 'private_edit_urls'
+        css_class = 'edit_me edit_me_private'
+        title = 'Editing on GitHub is available to Elastic'
```
nrichers: I updated the text for the tooltip after we talked; given in how many places we show the edit icon, I wanted to make the message more positive (i.e. just say who CAN edit rather than that you need to be an authorized user).

nrichers:
@Adrian-J a quick question for you: we are adding the option to edit docs on GitHub for the Cloud repo, but since the repo is private, we'll display a different edit icon and a tooltip that tells you editing is available only to Elastic. 

I'm wondering if you would have any ideas for an edit icon that indicates that editing is available but restricted? The icon is only 24 x 24 pixels, so there is not a whole lot to play with.

My initial idea to use a lock icon or a key gives off way too much of a negative vibe:

image

The usual edit icon: 
image

The edit icon for private repos as it is now (the usual icon but inverted and desaturated slightly--it's too subtle):
image

nrichers:
@nik9000 there's one other thing I noticed: When I was playing around with the edit icon for the private repo, my local doc build **refused** to pick up anything other than the very first version I edited. 

I expected it to use the latest local version of `resources/web/style/img/edit-me-private.png`, but no. It's almost as if the icon got cached somewhere and then additional local changes got ignored. I've not seen this behavior before.

nik9000:
> I expected it to use the latest local version of `resources/web/style/img/edit-me-private.png`, but no. It's almost as if the icon got cached somewhere and then additional local changes got ignored. I've not seen this behavior before.

Strange! I wouldn't have written the CSS quite the way @nyurik did but I figured it works. Maybe something is wedged? That'd be odd though.

Adrian-J:
@nrichers Of course, I will have those ready sometime this week.

nyurik:
I am currently trying to add some unit tests, and may need to reorg some tests further to allow for what is being done...

PR #1724, opened 2020-02-05
Add Workplace search docs to the build.

debadair:
Preview: http://docs_1724.docs-preview.app.elstc.co/guide/en/workplace-search/current/index.html

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -36,6 +36,7 @@ repos:
     sense:                https://github.com/elastic/sense.git
     stack-docs:           https://github.com/elastic/stack-docs.git
     swiftype:             https://github.com/elastic/swiftype-doc-placeholder.git
+    enterprise-search:    https://github.com/elastic/enterprise-search-pubs.git
     x-pack:               https://github.com/elastic/x-pack.git
     x-pack-elasticsearch: https://github.com/elastic/x-pack-elasticsearch.git
     x-pack-kibana:        https://github.com/elastic/x-pack-kibana.git
@@ -1346,7 +1347,7 @@ contents:
                 path:   libbeat/docs
 
     -
-        title:      Swiftype: Site, App, and Enterprise Search
+        title:      Swiftype: Site and App Search
         sections:
           -
             title:      Site Search Reference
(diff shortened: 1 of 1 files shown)


conf.yaml
```diff
@@ -1376,6 +1377,25 @@ contents:
...
+            private:    1
+            current:    master
+            branches:   [ master ]
+            single:     1
+            tags:       Workplace Search/Reference
+            subject:    Workplace Search
+            sources:
+              -
+                repo:   enterprise-search
+                path:   workplace-search-docs
```
bmorelli25: Doesn't this also need `chunk: 1`? Without it, I'm seeing [all content on the same page](http://docs_1724.docs-preview.app.elstc.co/guide/en/enterprisesearch/workplacesearch/current/index.html).

conf.yaml
```diff
@@ -1376,6 +1377,25 @@ contents:
...
+            index:      workplace-search-docs/index.asciidoc
+            private:    1
+            current:    master
+            branches:   [ master ]
+            single:     1
+            tags:       Workplace Search/Reference
+            subject:    Workplace Search
+            sources:
+              -
+                repo:   enterprise-search
```
goodroot:
```suggestion
                repo:   enterprise-search-pubs
```

https://github.com/elastic/enterprise-search-pubs

conf.yaml
bmorelli25: Or is that on purpose? I'm just now noticing the `single: 1`, which I haven't seen before.

debadair: Yes--and the "single" option needed to be removed.

conf.yaml
debadair: It's based on the repo name set above. I changed it in both places. (Originally left the "pubs" off because it is longer than all the other names.)

debadair:
After speaking with @AnneB-SEO about the URL, I dropped enterprisesearch from the URL and hyphenated workplace-search. The hyphenation is preferable from an SEO perspective, and we don't include the solution names in the URLs for the other products.

goodroot:
LGTM!... As far as I understand things. 😆 

@debadair What is the sequence of events?

1. PR is merged.
2. Elastic build runs at `x`
3. Docs are live on elastic.co?

goodroot:
@bmorelli25 Fixed upstream. Nice catch.

PR #1725, opened 2020-02-06
Fix console selector for ESS client docs

nrichers:
This PR fixes the language selector dropdown for the Elasticsearch Service docs. Previously, no selector was shown at all. With a small fix to conf.yaml, the selector now appears as expected:

image

Before this is ready for review: 

* With the selector enabled, none of the language examples can be found. Need to investigate further to see what's amiss. ~~Possibly related to us needing to map to the `master` branch in the clients-team repo?~~ EDIT: After investigating further, I think this might be because of how the examples in the elastic/cloud repo are written, which will require a separate PR to fix. 

  image

* Added the `master` branch to the ESS doc build for testing; need to remove before merging

Also added an alias for `docbldess`. Unrelated but helpful, since `docbldec` is dates back to the pre-renaming days of our SaaS.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -687,7 +687,8 @@ contents:
                 alternatives: { source_lang: console, alternative_lang: php }
                 repo:   clients-team
                 path:   examples/elastic-cloud/php
-                map_branches: *mapCloudSaasToClientsTeam
+                map_branches: &mapCloudSaasToClientsTeam
+                  release-ms-31: master
               -
                 alternatives: { source_lang: console, alternative_lang: go }
                 repo:   clients-team
--- a/conf.yaml
+++ b/conf.yaml
@@ -656,6 +656,11 @@ contents:
                 repo:   clients-team
                 path:   examples/elastic-cloud/python
                 map_branches: *mapCloudSaasToClientsTeam
+              -
+                alternatives: { source_lang: console, alternative_lang: csharp }
+                repo:   clients-team
+                path:   examples/elastic-cloud/csharp
+                map_branches: *mapCloudSaasToClientsTeam
 
           -
             title:      Elasticsearch Add-On for Heroku - Hosted Elasticsearch and Kibana for Heroku Users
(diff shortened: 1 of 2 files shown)


philkra:
good catch with changing `csharp` back to ` console`, I forgot that one! thanks

nrichers:
@philkra can you approve this PR? I think the remaining fix will need to go into the elastic/cloud repo.

conf.yaml
```diff
@@ -626,33 +626,33 @@ contents:
                 repo:   cloud
                 path:   docs/shared
               -
```
philkra: we're missing .NET, need to add the following block:
```
alternatives: { source_lang: console, alternative_lang: csharp }
repo:   clients-team
path:   examples/elastic-cloud/csharp
map_branches: *mapCloudSaasToClientsTeam
```

nrichers: Fixed, thank you for pointing this out!

nik9000:
run elasticsearch-ci/docs

conf.yaml
```diff
@@ -751,30 +761,35 @@ contents:
...
-                alternatives: { source_lang: csharp, alternative_lang: python }
+                alternatives: { source_lang: console, alternative_lang: python }
                 repo:   clients-team
                 path:   examples/elastic-cloud/python
                 map_branches: *mapCloudEceToClientsTeam
+              -
+                alternatives: { source_lang: console, alternative_lang: csharp }
+                repo:   clients-team
+                path:   examples/elastic-cloud/csharp
+                map_branches: *mapCloudSaasToClientsTeam
```
nik9000: s/mapCloudSaasToClientsTeam/mapCloudEceToClientsTeam/

nrichers: Good catch! In my pre-meeting hurry to address Philip's comment, I messed this up. Fixed now.

nik9000:
Looks like it is working now!

nrichers:
Tested ESS, Heroku and ECE docs to make sure everything works as expected. Removed `master` from conf.yaml. This PR is ready to merge.

PR #1730, opened 2020-02-11
Removes old shared files

lcawl:
Related to https://github.com/elastic/docs/issues/804

This PR removes old shared version attribute files that are no longer required.


lcawl:

PR #1733, opened 2020-02-12
Update Azure ARM template current version to 7.6

russcam:
This PR updates the Azure ARM template current version to 7.6


russcam:
This PR must be merged only after 7.6 is live on the Azure Marketplace; I'll update this PR when that happens

russcam:
7.6.0 is now live on the Marketplace. Merging this in.

PR #1736, opened 2020-02-19
Revert "Enable custom "edit" icon for private repos"

nrichers:
This PR reverts elastic/docs#1719 until we can tweak the implementation to address some concerns that have been raised. 

Context: We told people about the **Edit** link for Cloud docs last week and generally it's been very well received, especially by Support. But I also got some pretty strong feedback that we shouldn't be showing the **Edit** link by default to people who'll get 404s if they try to follow the link. There are also some concerns around linking to private repos from our public docs. 

I will continue working with our amazing infra team to sort out this out. Possible options might include: 

* Provide an alternate URL that enables the **Edit** link. For example: if you add `?edit` at the end, we could show the **Edit** link. This would be especially helpful if we could remember the state--once you switch to the edit mode, you always see the link. (A plugin to enable the edit button was also suggested, but might be more involved.)
* Fork of the docs to a public repo to enable public editing for everyone and sync the commits back to the Cloud repo, but that would be more involved.

FYI @ppf2 @Leaf-Lin @nyurik @drewr


ppf2:
Thanks for the ping and continuing to look for alternate solutions. I have reverted the process change on the Edit button on our support wiki.

PR #1737, opened 2020-02-20
Add standard ESS trial attribute

nrichers:
We'd like to start collecting some metrics for ESS trial sign-ups from the docs, for which we need a standard attribute. 

Relates to https://github.com/elastic/marketing-ops/issues/7367.


karenzone:
@dedemorton FYI. This applies to us, too.

PR #1747, opened 2020-02-25
Correct CSS for h5 headings

jrodewig:
Fixes a typo in the CSS that breaks the font-size for h5 headings.


jrodewig:
Thanks @JoshMock!

PR #1749, opened 2020-02-25
Remove alternatives for Heroku and ECE

nrichers:
We now have new client docs for ESS, Heroku, and ECE. Just one gotcha that @philkra pointed out: The examples rely on the Cloud ID, which is not currently supported for Heroku and ECE.

I looked at reverting at the old docs we had for connecting, but they are just not that great. So, I came up with an idea that involves a couple of PRs:

* This PR removes the alternatives from conf.yaml for Heroku and ECE. This removes the selector, but leaves the default console example that should work just fine.
* In a second PR, provide an alternate lead-in text for the "Work with language clients" section that doesn't mention the alternative languages that we just removed.
 
Relates to https://github.com/elastic/docs/pull/1725

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -685,42 +685,6 @@ contents:
               -
                 repo:   cloud
                 path:   docs/heroku
-              -
-                alternatives: { source_lang: console, alternative_lang: php }
-                repo:   clients-team
-                path:   examples/elastic-cloud/php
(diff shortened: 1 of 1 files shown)


nrichers:
I tinkered with this some more and I don't think this is the right approach--the content left over is not entirely useful. Closing.

PR #1758, opened 2020-02-27
Update elasticsearch-py document path

sethmlarson:
Alternative example docs are in a new location.
Depends on https://github.com/elastic/elasticsearch-py/pull/1122 landing first.


sethmlarson:
Cycling to see that the docs CI job passes now

PR #1764, opened 2020-03-06
Enable custom "edit" icon for private repos, take 2

### LINKED ISSUE #1629: Make the Edit button work for Cloud docs!

emanuil-tolev:
Resolves #1629 after PR #1719 was [reverted](https://github.com/elastic/docs/commit/83c35dbe6f9364a67f2dccf54025d267c2a76890).

Changes on top of the original PR discussed with @nyurik: the Edit button is only shown on Cloud docs if `edit` is present in the query string or hash. Furthermore, all internal links are rewritten to keep the toggle on while navigating the docs. Removing it manually from the URL stops rewriting of course.

I've preserved whatever existing query strings and hashes there may be in the relative links in the docs during the link rewrite. Unfortunately, to do that in an IE-compatible manner, I had to add an extra function to parse the existing query portion - jQuery can serialise that with `$.param`, but not parse it. I used the most tested and relied upon function I could find, since writing one has all sorts of odd corner cases. A modern alternative is https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams but alas, IE.

EDIT: To check locally: 

`./build_docs --private --open --doc $GIT_HOME/cloud/docs/saas/index.asciidoc --resource=$GIT_HOME/cloud/docs/shar --chunk 1` (note `--private` as opposed to usual cloud build alias - I can change the build alias if that's a good idea at this stage.)


emanuil-tolev:
To check locally: 

`./build_docs --private --open --doc $GIT_HOME/cloud/docs/saas/index.asciidoc --resource=$GIT_HOME/cloud/docs/shar --chunk 1` (note `--private` as opposed to usual cloud build alias - I can change the build alias if that's a good idea at this stage.)

resources/web/docs_js/index.js
```diff
@@ -267,5 +267,23 @@ $(function() {
     document.head.appendChild(hotcss);
   }
 
+  // If you're have been given access to the private repository these docs
+  // relate to *and* you use the 'edit' in the query string or hash, we'll show
+  // you an edit button taking you to the private repo.
```
nyurik: The comment is not correct - this code doesn't test if you have access to the private repo.

```suggestion
  // For the private docs repositories, the edit button is hidden
  // unless there is an '?edit' in the query string or hash.
```

resources/web/docs_js/index.js
```diff
@@ -267,5 +267,23 @@ $(function() {
...
+  if (window.location.search.indexOf('edit') > -1 || window.location.hash.indexOf('edit') > -1) {
```
nyurik: I think we should explicitly test for `?edit` in the hash, because I suspect we often use hashes to link to sections of the documentation, and there is a high chance the word `edit` would be there already. The `?edit` on the other hand is much less likely.  I think having `edit` as is in the query part is ok, as I don't think we use query params (unless we use it for searching in the docs?)

resources/web/docs_js/index.js
```diff
@@ -267,5 +267,23 @@ $(function() {
...
+    $('a.edit_me_private').show();
+
+    // Carefully rewrite the query string of each relative docs link to add an
+    // 'edit' param while preserving any existing parameters.
+    $('a').each(function(index) {
+      if (this.href.startsWith(window.location.origin)) {
+        var query_obj = utils.deparam(this.search);
+        query_obj.edit = "";
+        console.log(query_obj);
```
nyurik: log?

nyurik:
Thank you so much for working on this!  @nik9000 might have more thoughts in addition to what I wrote.

resources/web/docs_js/index.js
emanuil-tolev: > there is a high chance the word `edit` would be there already

agreed - amended

>  I don't think we use query params (unless we use it for searching in the docs?)

I asked myself that a lot as I wondered whether the param preserving code was needed! Best to assume we do. Given that with this PR we have the ability to `var query_obj = utils.deparam(this.search);` maybe just use that and check for the presence of `query_obj.edit`?

resources/web/docs_js/index.js
emanuil-tolev: 😳 oops, done.

resources/asciidoctor/lib/edit_me/extension.rb
```diff
@@ -74,8 +74,16 @@ def link_for(block)
       url = edit_url block
       return '' unless url
 
+      if block.document.attr 'private_edit_urls'
+        css_class = 'edit_me edit_me_private'
```
nik9000: s/css_class/css_classes/ ?

resources/web/docs_js/index.js
```diff
@@ -267,5 +267,22 @@ $(function() {
     document.head.appendChild(hotcss);
   }
 
+  // For the private docs repositories, the edit button is hidden
+  // unless there is an '?edit' in the query string or hash.
+  if (utils.deparam(window.location.search).edit !== undefined
+      || window.location.hash.indexOf('?edit') > -1) {
```
nik9000: Do we need to do this for the anchor? That feels a little funky.

resources/web/docs_js/utils.js
```diff
@@ -139,3 +139,103 @@ export const getCurlText = ({consoleText,
...
+
+/*
+  Parse query strings in a battle-tested, good practice manner.
+  An extraction of the deparam method from Ben Alman's jQuery BBQ plugin
+  http://benalman.com/projects/jquery-bbq-plugin/
+  Exact source: https://github.com/cowboy/jquery-bbq/blob/8e0064ba68a34bcd805e15499cb45de3f4cc398d/jquery.ba-bbq.js#L466
+  Unit tests maintained separately: http://benalman.com/code/projects/jquery-bbq/unit/
+  Licence: dual-licensed under MIT and GPL, we use https://github.com/cowboy/jquery-bbq/blob/8e0064ba68a34bcd805e15499cb45de3f4cc398d/LICENSE-MIT
+*/
+export function deparam(params, coerce) {
```
nik9000: I don't know the right thing, but I know we have Redux and React in the project. Is there a good way to get them to do this?

resources/asciidoctor/lib/edit_me/extension.rb
nyurik: you already reviewed this last time :D , but sure, i'm fine with that too.

resources/web/docs_js/index.js
nyurik: I think we should handle both for a simple (yet weird) reason:  The workflow for this feature is to navigate to a page, see that something is broken, click the URL, add `?edit` at the end, hit enter, and click the edit button (not ideal, i know).  The problem with the above is that if the URL ends with an anchor, adding `?edit` to it wouldn't do anything, so the user would have to find the beginning of the anchor and edit there. If we can support this weird case out of the box, user won't have to worry about it.

resources/asciidoctor/lib/edit_me/extension.rb
nik9000: :man_shrugging: Sorry. Either way is fine I guess.

resources/web/docs_js/utils.js
emanuil-tolev: Apparently not. I did some digging and [they decided to explicitly not process querystr](https://github.com/ReactTraining/react-router/issues/4410) in Router v4 (and later), where this kind of functionality would naturally fall. Instead they allow for the use of any 3rd party lib.

I'm not sure how Preact v8 interacts with this but have been unable to find a Preact solution to it. A project (providing routing among other functionality) uses our exact version of Preact in its docs examples and explicitly states a 3rd party library is required for querystr parsing with that setup: http://meiosis.js.org/docs/routing.html (search "Again, there are several options"). It appears Preact doesn't do it either.

When searching for solutions around Redux it's pretty clear people expect you to use React Router to pass querystr data to Redux for state mgmt.

I've narrowed down the options:

1. One of the 4 libraries from [this list](http://meiosis.js.org/docs/routing.html). From these https://github.com/sindresorhus/query-string seems most popular on NPM and well maintained. It will parse querystr and hash.

2. Use the new-ish [universal web API URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams). No IE support at all, so requires [a seemingly popular and well maintained polyfill](https://github.com/jerrybendy/url-search-params-polyfill).

3. The deparam function I used before I found the above. Upon further digging into solutions around react, this approach has been superseded and should be replaced by 1. or 2. 

I believe 2. is a good choice resulting in the least amount of extra code to maintain and using a fairly well established browser API. The polyfill can be dropped whenever we drop IE support, much as any of the other polyfills we use, and leave us with no custom querystr parsing code at all. My previous implementation had a further problem, so I replaced it with this option. Happy to change it though, naturally.

resources/asciidoctor/lib/edit_me/extension.rb
emanuil-tolev: I'd use `css_class_string` or `css_classes` so changed to the plural.

emanuil-tolev: I'd use `css_class_string` or `css_classes` so changed to the plural.

emanuil-tolev:
@nrichers and I had a quick Zoom to figure out what's going on. The local build problem went away on its own for him when he retried - an unrelated error likely related to docker caching. We did find a different problem which broke when the private edit btn was shown. That's fixed.

Additionally, internal links in the ToC on the side (fetched by AJAX as toc.html) were not getting rewritten, only links in the body of the current doc page were. Fixed.

@nyurik re external copy / paste: for some reason functionality which has been needed since the dawn of the web - parsing query strings in URLs - has several different approaches in JS. Reimplemented using a more modern solution found on NikE's prodding. https://github.com/elastic/docs/pull/1764#discussion_r390085982

PR should be ready for re-review. @nrichers: if you could re-test in the same way that'd be great!

resources/web/docs_js/utils.js
nik9000: I guess I'd go with either 1 or 2, adding it to `package.json` to get it in. I *think* the polyfill approach is a little better, but I super don't have strong opinions.

resources/web/docs_js/index.js
emanuil-tolev: Now using URLSearchParams web API to achieve this with polyfill for IE, it works very well.

resources/web/docs_js/index.js
emanuil-tolev: This ok with you @nik9000 ? I don't see how `?edit` in the hash part would cause problems with our own use of anchors.

resources/web/docs_js/utils.js
emanuil-tolev: I did end up [adding the polyfill to package.json and yarn.lock](https://github.com/elastic/docs/pull/1764/commits/fc9e2a9d6e5f5aa291426aeeda654e017a97be14) and [invoking the web API (polyfill in IE)](https://github.com/elastic/docs/pull/1764/commits/6fac8be75a605ff1685df03bceba7f1acb0db16c).

I have also tested this in IE9-IE11 on browserstack.com on Win 7 - Win 10. All work.

Does the PR look ok to you now code-wise?

nrichers:
@emanuil-tolev @gtback Just checking in on this PR ... what would you need to get this PR working soon-ish? I'd love to have the edit links working for Cloud—just got another ping this morning about a doc issue that got opened that could have been addressed directly and we run into this several times a week, every week. Love this PR and would love to have it see the light of day. ❤️

gtback:
@emanuil-tolev If there's anything I can do to move this forward, let me know. You understand this code a lot better than I do at this point, but I'm happy to help if you want me to.

nrichers:
Just checking in again. Got pinged by @111andre111 today about the **Edit** button. We definitely still want this and the request continues to come up. 😄

gtback:
@emanuil-tolev will you be able to take a look at this in the next couple weeks, or do you need me to?

emanuil-tolev:
> @emanuil-tolev will you be able to take a look at this in the next couple weeks, or do you need me to?

Sorry Greg, I have reviewed this. The problem here was the link rewriting - I had no way of testing it and it [did cause an unexpected problem](https://github.com/elastic/docs/pull/1764#pullrequestreview-372481538). I just didn't have time to find out whether we had tests that ran after the full DOM was loaded and could check the result (like Selenium would).

Then you said:
> It would certainly be a nice touch if users with access to the private repo could "browse" the site and only have to worry about manually adding `?edit` once, but I don't think it's the worst experience for them to just have to add it whenever they see a section they want to change.

You're right! I've removed the bit that rewrites links. 

To test I used the docs repo and the cloud docs repo: `./build_docs --private --open --doc $GIT_HOME/cloud/docs/saas/index.asciidoc --resource=$GIT_HOME/cloud/docs/shared --chunk 1`, opening http://localhost:8000/guide/ec-getting-started.html with and without ?edit , and the Edit icon appears only when the param is present without propagating further.

A bit of admin: I've merged docs master into my private branch to make sure there are no conflicts. Happy to do a rebase against docs master and force push to `emanuil-tolev/docs:private-edit` instead if that's preferred.

gtback (approved):
This is working great!

gtback:
@emanuil-tolev Thanks for taking the time to wrap this up. Merging in `master` is perfect; this repo isn't high-traffic enough that we worry about rebasing or squash-merging anything. 

@nrichers can you take a look as well and confirm this is working? [Preview Link](https://docs_1764.docs-preview.app.elstc.co/guide/index.html)

111andre111:
I have tested from my side as a third party and it is working for me as well.

nrichers (approved):
This PR is important and we should merge it, with a huge thank you to @emanuil-tolev for implementing the `?edit` functionality! ❤️ I did notice a few quirks during testing that we can look at separately or decide to live with. 

Quirks: 
* You shouldn't be able to edit our generated API reference (to make changes, you need to regenerate it after making changes to the Scala annotations), but the edit button is available. Trying to edit it gives you a blank edit window. Not sure why this happens, but the blank edit window is preferable over letting you edit Asciidoc source that's just going to get overwritten anyway. (To reproduce, try the edit button on [this page](https://docs_1764.docs-preview.app.elstc.co/guide/en/cloud-enterprise/current/search-clusters.html?edit).)
* You can't edit content we pull in from other repos, which is fine. E.g. if you go to [Manage Elastic Stack versions](https://docs_1764.docs-preview.app.elstc.co/guide/en/cloud-enterprise/current/ece-manage-elastic-stack.html) and click **Expand to view full list** , the content from that list lives in the cloud-assets repo and is imbedded here. You can edit only the parent file that does the imbedding.  
* Since this PR was first opened, the Cloud team has changed its git workflow and we now backport rather than forward port PRs. To follow the workflow, you would need to edit files in `master` and then use the automated backporting we provide. The edit button edits the file in whatever branch you're on (and correctly edits in MS 41 for ECE 2.6, which is nice!), but this does sort of break our workflow. We can ask the Cloud writers to make sure they change the base branch to `master` when they review or to do a `git rebase --onto` if that doesn't work, or to do a git cherry pick. There is a chance that changes will go into one branch other than `master`, though, and we effectively lose them in the long run.

gtback:
Thanks, @nrichers. My vote would be to merge this as is, and deal with any of the other quirks separately (if at all).

> You shouldn't be able to edit our generated API reference...

Unfortunately, we don't have any metadata in the docs right now (that I know of) to indicate which content is auto-generated and which isn't. This is an existing problem with (for instance, the `ecctl` command reference). If we try to fix this, it's beyond the scope of this PR.

> You can't edit content we pull in from other repos...

This might actually be pretty hard, since I'm not sure if we know, at the time we're putting these links in, where the actual content is coming from (it's already been `include::`-ed). This is also an issue affecting existing content, wherever it's included from, so I'd want to handle this in a separate PR.

> Since this PR was first opened, the Cloud team has changed its git workflow...

I don't know of a good way to generalize this behavior across all repos. Asciidoctor and our Ruby code generally aren't aware of any of the other branches of a book, so we'd have to pass in a "edit prefix" that could be customized per-book in conf.yaml. As you said, the edit button edits the file in whatever branch you're on, which is probably pretty decent default behavior. 

The last two issues will essentially go away once we formalize a one-input-file-to-one-output-page model, so I'd be tempted to wait to see if they're actually a problem before putting a lot of effort into fixing them.

nrichers:
@emanuil-tolev will you do the honours of merging this is shall I? 😁

PR #1761, opened 2020-03-03
Improving Visual Content Hierarchy and Legibility

jonasll:
# Improving Visual Hierarchy and Legibility

This PR aims at improving the consumability across all guides. Impact of these changes have been reviewed across the `elasticsearch`, `kibana`, `beats` and `workplace-search` books. 

Here's a list of the main tweaks (before and after included):

## 1. Header and Paragraph Hierarchy
This change aimed and creating better distinctions between sections and group information together with their headers. The previous iteration made the documents feel crowded and difficult to parse when jumping from section to section.

hierarchy

---

## 2. Table Legibility
The current styles are very tight, and information is hard to consume for tables, large and small.

tables

---

## 3. Miscellaneous Alignment and Padding Tweaks / Bug Fixes
Devil is in the detail, and some of the misaligned or inconsistent objects have been fine-tuned. Here are a few examples:

spacing

align

crumbs

---

## Bonus: Improved Step Process for Image-Intensive Guides (i.e. Solutions Guides)
Solutions Guides tend to be more visually rich, especially around flows and sequences. While not entirely related to this PR, it highlights some of the tweaks around Figure Caption hierarchy, and overall displays the combination of many of the aforementioned tweaks into a single view. The final product is strikingly more consumable.

ws_service

---

@dedemorton @debadair for review. Please add anyone else who you may think can add value to this review.

PR #1770, opened 2020-03-10
Improve styling for multi-paragraph admonitions

jrodewig:
PR #1761 set the bottom margin for paragraphs inside admonitions to
zero. This "squishes" all paragraphs within admonitions into a
semi-paragraph.

This tweak sets a zero bottom margin only on the _last_ paragraph
within an admonition. This eliminates unneeded grey space within the
admin without squishing paragraphs.

### Before
before

### After
after


jrodewig:
Thanks @jonasll!

PR #1782, opened 2020-03-23
[DOCS] Add function reference template

jrodewig:
Adds a reference template for function documentation.

I ended up needing this while documenting functions for
EQL in the Elasticsearch docs. Even within the Elasticsearch
docs, we have several conventions for documenting functions.

The goal of this template is to standardize the way Elastic
documents functions. Inspiration was taken from the Kibana docs:

https://www.elastic.co/guide/en/kibana/master/canvas-function-reference.html

Any and all feedback welcome!

### DIFF
--- a/shared/fn-ref-ex.asciidoc
+++ b/shared/fn-ref-ex.asciidoc
@@ -49,7 +49,7 @@ Guidelines for parameter documentation
 * Use a definition list.
 * End each definition with a period.
 * Include whether the parameter is Optional or Required and the data type.
-* Add `{multi-arg}` to parameters that accept multiple arguments
+* Add `{multi-arg}` to parameters that accept multiple arguments.
 * Include default values as the last sentence of the first paragraph.
 * Include a range of valid values, if applicable.
 * If the parameter requires a specific delimiter for multiple values, say so.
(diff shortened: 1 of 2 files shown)


shared/fn-ref-ex.asciidoc
```diff
@@ -0,0 +1,83 @@
...
+substring(<source>, <start_pos>[, <end_pos>])
+----
+
+////
+Guidelines for parameter documentation
+***************************************
+* Use a definition list.
+* End each definition with a period.
+* Include whether the parameter is Optional or Required and the data type.
+* A † denotes a parameter can be passed multiple times.
```
lcawl: > A † denotes a parameter can be passed multiple times.

I actually had to hunt for that legend in the Kibana example, since I missed see that at the top.  Would it be feasible to have these as callouts or footnotes instead so they're linked?

jrodewig: I think that's a great idea. I added this as a footnote with ec18644ef072354dbeecb91b838d9a9f80040996. Let me know if you have any suggestions on the implementation.

jrodewig:
I'm going to go ahead and merge this since it's already live in the EQL function docs:
https://www.elastic.co/guide/en/elasticsearch/reference/master/eql-function-ref.html

However, please continue to provide feedback and/or iterations via future PRs or issues. Thanks!

PR #1786, opened 2020-03-25
[DOCS] Adds child_attributes CSS role

lcawl:
This PR creates a stylesheet with a border around collapsed sections (detail/summary objects). It's intended for use in nested objects on API reference pages.

### DIFF
--- a/resources/web/style/child_attributes.pcss
+++ b/resources/web/style/child_attributes.pcss
@@ -4,7 +4,7 @@
     border: 1px solid #A9A9A9;
     border-radius: 10px;
     background: #fbfbfb;
-    padding: 15px 20px 0;
-    margin: 10px 15px 20px;
+    padding: 15px 10px 0;
+    margin: 10px 5px 20px;
   }
 }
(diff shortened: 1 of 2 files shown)


bmorelli25:
@lcawl, This is super cool! Question, would it be worth reconsidering the background color used for these collapsed sections? The color used, `#fbfbfb` is very similar to the color we use for inline code snippets: `#f8f8f8`. (Holy moly, how cool is it that GitHub automatically shows you the colors!)
Screen Shot 2020-04-01 at 11 15 13 AM

IMO, this makes it difficult to differentiate between inline code and regular text inside of collapsed sections. At the same time, code is still the monospacing, which does help, so I'm not sure how important this really is 🤷‍♂. Here's an example:
Screen Shot 2020-04-01 at 11 16 07 AM

lcawl:
@bmorelli25 Thanks for taking a look!  As far as I'm concerned, improvements to the stylesheet are very welcome.  I just copied the "sidebar" stylesheet then changed the blue outline to grey and minimized the padding.  If you create a PR that makes it prettier or more easily to differentiate, I'm all for it.

PR #1794, opened 2020-04-03
docs: update child_attributes background color

bmorelli25:
## Summary
IMO, the background color for `child_attributes` makes it difficult to differentiate between inline code and regular text inside of collapsed sections. @lcawl also pointed out that admonition blocks don't seem to be showing up, as they probably use the same grey. **This PR explores some alternative background colors.**

Here's an example of the current background color:
Screen Shot 2020-04-01 at 11 16 07 AM

And here's an enlarged comparison:
Screen Shot 2020-04-01 at 11 15 13 AM

## Alternatives

Some of these are just terrible, but I took screenshots, so I might as well share them. I like `#ebf7fa` the best, but I'm also not in love with it. Open to any and all suggestions.

### `#f1f8ff`
Screen Shot 2020-04-03 at 12 53 15 PM

### `#b7c9e2`
Screen Shot 2020-04-03 at 12 54 32 PM

### `#cafffb`
Screen Shot 2020-04-03 at 1 03 12 PM

### `#ebf7fa`
Screen Shot 2020-04-03 at 1 02 08 PM

### `#e0f9ff`
Screen Shot 2020-04-03 at 12 57 34 PM

---
cc @jrodewig / @KOTungseth

### DIFF
--- a/resources/web/style/child_attributes.pcss
+++ b/resources/web/style/child_attributes.pcss
@@ -3,7 +3,6 @@
   details > div {
     border: 1px solid #A9A9A9;
     border-radius: 10px;
-    background: #ebf7fa;
     padding: 15px 10px 0;
     margin: 10px 5px 20px;
   }
--- a/resources/web/style/child_attributes.pcss
+++ b/resources/web/style/child_attributes.pcss
@@ -1,9 +1,9 @@
 #guide .child_attributes {
-  
+
   details > div {
     border: 1px solid #A9A9A9;
     border-radius: 10px;
-    background: #fbfbfb;
+    background: #ebf7fa;
     padding: 15px 10px 0;
     margin: 10px 5px 20px;
   }


bmorelli25:
And yes, I'm aware of how awful I am at design

KOTungseth:
I've requested the expertise of the Kibana Design Team to recommend a color that will be loved by all.

snide:
👋 Since these seem to open in an accordion I'd suggest forgoing the color / border all-together. Use a left margin to "inset" the content and demarcate it as child content. If you have a link to a page that has this, I can likely munge up the css to show what I mean.

bmorelli25:
@snide Thank you! [Here](https://www.elastic.co/guide/en/elasticsearch/reference/master/get-dfanalytics-stats.html) is the page I've been testing on.

The current margin value of 5px puts the border right underneath the triangle, which makes sense to me. Anything else I do looks goofy, so I'll let you play around with that. Removing the background color looks good though!

Screen Shot 2020-04-06 at 8 23 00 AM

lcawl:
I like the lack of background colour but keeping the border.  Thanks, @bmorelli25 !

bmorelli25:
See http://docs_1794.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/get-dfanalytics-stats.html. I'm merging this now, but happy to reassess if Dave comes back with a better idea.

PR #1799, opened 2020-04-08
Add ECK 1.1

anyasabo:
Adds ECK 1.1 to the dropdown. Will followup with a PR in the future to make it the current branch.

It wasn't clear to me if I should also add it to the excludes array just below it that includes the 1.0 branch. It was added here https://github.com/elastic/docs/pull/1297

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -823,7 +823,7 @@ contents:
               -
                 repo:   docs
                 path:   shared/versions/stack/7.x.asciidoc
-                exclude_branches:   [ master, 1.0, 1.0-beta ]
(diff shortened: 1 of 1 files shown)


gtback:
@lcawl might also be able to help with the dependencies between branches of the various repos.

lcawl:
I chatted with @alaudazzi and checked https://raw.githubusercontent.com/elastic/cloud-on-k8s/1.1/docs/index.asciidoc, which shows that 1.1 branch using the current.asciidoc file.  I've therefore added a commit that updates the conf.yaml accordingly.

PR #1804, opened 2020-04-10
Add kibana observability app names to shared attribute list

dedemorton:
These attributes are defined in the obs docs, but they need to be defined in the shared file so all projects can use them.

@gchaps Let me know what else you want in this list, or feel free to push commits to my fork, if you want to list other app names.

### DIFF
--- a/shared/attributes.asciidoc
+++ b/shared/attributes.asciidoc
@@ -92,6 +92,15 @@ ifdef::elasticsearch-root[]
 endif::elasticsearch-root[]
 
 //////////
+Kibana app names
+//////////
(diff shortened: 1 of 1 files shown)


dedemorton:
@andrewvc FYI... After this PR is merged, you can use these attributes in your attribute definitions. Just make sure you put your attribute definitions _after_ the include for `shared/attributes.asciidoc[]`. So you can have:

:beat_kib_app: {kib} {logs-app}

Then you won't have to worry about updating the locally-defined Beats attributes if the app names changes.

PR #1807, opened 2020-04-13
Remove outdated info about edit links

gtback:

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -1770,10 +1770,6 @@ We automatically generate `edit` links for most sections to make it easier for
 folks to contribute simple fixes and to help folks find the asciidoc file that
 generated a particular section. It should appear next to every title-like thing.
 
-Books built with AsciiDoc will automatically pick the correct url for files
-from the "root" repository. It guesses the wrong url for files from other
-repositories so you have to manually set `edit_url` at the top of each file.
-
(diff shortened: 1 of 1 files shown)


README.asciidoc
```diff
@@ -1770,10 +1770,6 @@ We automatically generate `edit` links for most sections to make it easier for
 folks to contribute simple fixes and to help folks find the asciidoc file that
 generated a particular section. It should appear next to every title-like thing.
 
-Books built with AsciiDoc will automatically pick the correct url for files
-from the "root" repository. It guesses the wrong url for files from other
-repositories so you have to manually set `edit_url` at the top of each file.
-
```
karenzone, line 1776: >It guesses the wrong url for files from other
repositories so you have to manually set `edit_url` at the top of each file.

Is this no longer an issue for files outside of the `elastic` org?  I'm wondering about all of the Logstash plugins.

nik9000: This is the same edit url that asciidoctor has had for  long time. I don't recall how good a job it does with the logtash org.

karenzone: I tested, and think we can safely merge this change.  There's still a bit of Logstash weirdness, but I can manage those edge cases with comments in the index files.  Thanks!

PR #1808, opened 2020-04-15
[Endpoint]: Add shared attributes repo to security docs config

DonNateR:

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -1490,6 +1490,12 @@ contents:
               -
                 repo:   security-docs
                 path:   docs
+              -
+                repo:   docs
+                path:   shared/versions/stack/{version}.asciidoc
+              -
+                repo:   docs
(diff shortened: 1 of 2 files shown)


conf.yaml
```diff
@@ -1478,7 +1478,7 @@ contents:
         sections:
           -
             title:      Elastic Endpoint Security
-            prefix:     en/endpoint/guide
+            prefix:     en/endpoint
```
gtback, line 1481: Depending on how long this guide has been up, we may want to set up redirects from the current URLs to the new ones.

DonNateR: Honestly, only people internally have tried to access this link so far... And there's no officially published documentation yet.

Not sure it's needed.

gtback: Ok, then I'd agree they aren't needed 😀

PR #1814, opened 2020-04-22
WIP Add Cloud icon

nrichers:
This PR adds an icon for Cloud features that you need to highlight in your docs, similar to the X-Pack one that we still use in some of our docs. If you need to indicate what Elasticsearch or Kibana user settings are supported on Elasticsearch Service, for example, you can use this icon. 

**Sample usage** 

Prefix the chunk of content you want to highlight with `[role="cloud"]`: 

```
[role="cloud"]`xpack.http.ssl.cipher_suites` setting is now whitelisted in {n}. This setting can be useful when using outgoing
TLS traffic to external services with restrictive supported cipher suites. One example would be using Telegram from Watcher.
```

The `[role="cloud"]` is supported wherever the X-Pack icon works, such as section headings, paragraphs, list items, and so forth.

To dos before this is ready for review:
- [x] Add the CSS for the new icon based on how the X-Pack one is currently implemented.
- [ ] Complete code changes to the Ruby files to support the icon. The CSS work is done, but you currently need to add the class to the HTML output manually:
   image 
- [ ] Replace the temporary Cloud icons with ones that include some text and look more like the X-Pack ones.
- [ ] Add some instructions to the elastic/docs readme about this reusable content.


bmorelli25:
As a side note, it is already possible to do something similar to this. In the APM Agent documentation, we use inline svg images to indicate dynamic configuration options. The same concept could be applied here.

Add an attribute to `shared/attributes.asciidoc`:
```
// With a local link
:cloud-config: <<dynamic-configuration, image:./images/cloud.svg[] >>

// Or, without a link
:cloud-config: image:./images/cloud-default.svg[]
```

Then, just add `{cloud-config}` to each relevant configuration:
```
==== `capture_body` {cloud-config}
```

Here's the end result:

Screen Shot 2020-04-23 at 12 48 53 PM


With that being said, I do think your idea is better. Mine is more of a hack but might be useful if this is too difficult to implement (which I don't think will be the case).

gtback:
@nrichers Let me know if you'd like me to help out with this. If this is a pattern we imagine extending (beyond the X-Pack and Cloud logos), I'll try to come up with a more scalable solution. But if it's just going to be these two, I'm fine with this as is.

nrichers:
@lcawl @nik9000 when you have a moment, can you shed some light on the X-Pack icon that is used in some of our documentation?

image

I started implementing (imitating) something very similar for Cloud and got the CSS and about half of the Ruby side o' things done. There's probably an hour or two of work left over, assuming I can figure out the remaining Ruby bits. (Not having had prior exposure to Ruby, I'm enjoying the chance to learn more.)

At our cloud-first workgroup today, Brandon mentioned that he uses a simple attribute for a graphic in the APM docs. He's added https://github.com/elastic/docs/pull/1814#issuecomment-618626942 to suggest that something very similar might work for our cloud content. 

**Can you tell me why you chose the implementation for the X-Pack icon that we use today?** I imagine there is a very good reason for it and I found a number of PRs to enable the icon to work in a wide range of cases that I suspect might not be supported by a simple attribute. I'd love to know.

lcawl:
Honestly, we implemented it so long ago that I don't recall all the things we tried and discarded. If this other solution works for all the scenarios you need, go for it!

nik9000:
> Honestly, we implemented it so long ago that I don't recall all the things we tried and discarded. If this other solution works for all the scenarios you need, go for it!

We did it in the before times so it was all xslt hackery. And I carried forward support for it so it was *exactly* the same. If we want to add another icon I think it'd be worth figuring out if there is an asciidoctor-native way to support it. And, if we've broken that way in the name of backwards compatibility, then maybe we should fix *that*.

nrichers:
After thinking about @nik9000's https://github.com/elastic/docs/pull/1814#issuecomment-619223405 a bit more, I decided to take a look at the current Asciidoctor docs. There's some additional icon support in there, but the gist of what I found is that a solution like the one suggested by @bmorelli25 is likely right. I'll open a new PR with the changes.

PR #1815, opened 2020-04-23
Single-sourced content for cloud and cloud icon

nrichers:
This PR:

* Adds standard topics for spinning up the Elastic Stack on cloud and accessing Kibana, e.g.:

image

* Moves the various cloud attributes into a separate section to make them easier to spot:

   image

* Adds some standard lead-in text you can use to point people to our Elasticsearch Service (`:ess-leadin:` ^^)

* Adds new cloud icons for ECE and ESS that anyone can use to indicate if something is supported on cloud

   image

The content in this PR isn't intended to be the final say, but rather to create something different books can imbed now, to be reworked over time.  Some improvements we should handle in the future:

* Add more single-sourced cloud tasks, e.g. for upgrading deployments
* Figure out how we can source the cloud icons out of elastic/docs without requiring `--resource=$GIT_HOME/docs/` for every book (or the conf.yaml equivalent)
* If the previous bullet is addressed, to stop using the cloud icons stored on S3

Sample usage for single-sourced content is described in the source files themselves and will also be highlighted in the cloud-first writing guidelines we're working on.


shared/cloud/ess-getting-started.asciidoc
```diff
@@ -0,0 +1,25 @@
+[cloud-ess-setup]
+== Set up on Cloud
+
+// Need this at a different level? Use leveloffset: +1 to change to H3, 
+// for example, and then change the offset back with leveloffset: 0 afterwards.
```
bmorelli25, line 5: This is awesome! My comment from the weekly meeting: To keep these as flexible as possible, I'm not sure that we should include headers in these files. Instead, we could comment out a suggested header, which would encourage harmony across documentation, but allow each product to customize headings and anchor tags as necessary.

bmorelli25: A side effect of this is that we wouldn't be able to include links live you done at the bottom of this file: https://github.com/elastic/docs/pull/1815/files#diff-934228d98950ff95c8c53444966daeaeR25

bmorelli25: Another side effect is the "edit" button linking to the product documentation instead of the docs repo documentation. That might actually be a good thing though?

gtback:
@nrichers you can use `include::{docs-root}/shared/cloud/ess-getting-started.asciidoc[]` rather than traversing up and out of the current book with `{asciidoc-dir}`.

Like @bmorelli25 said, I think the warning about being unable to generate a header link is because the heading is [in the shared file, not in the book](https://github.com/elastic/docs/blob/8c93dcf1eeb10f80ca3f7924e86fd61469ca2dbc/resources/asciidoctor/lib/edit_me/extension.rb#L93):
> source_location.path doesn't work for relative includes outside of the base_dir which we use when we build books from many repos.

I'm sure we can adapt that to work with shared-reuse snippets like this that we want to include in the main docs repo, *if* we decide that we want to keep the headers in the shared content, rather than everywhere that content is used.

shared/cloud/ess-getting-started.asciidoc
nrichers: I tinkered with an `ifeval::` statement to enable or disable the heading in your book, but in the end decided that your suggestion is simply the most straightforward ... and commented out the heading. I think keeping it as flexible and simple as possible is the right thing to do here.

nrichers:
> @nrichers you can use `include::{docs-root}/shared/cloud/ess-getting-started.asciidoc[]` rather than traversing up and out of the current book with `{asciidoc-dir}`.

Excellent suggestion, thank you! I changed to using `{docs-root}`.

shared/cloud/ess-getting-started.asciidoc
```diff
@@ -0,0 +1,22 @@
...
+// :leveloffset: 0
+ 
+//[[cloud-ess-getting-started]]
+//== Get started with {ess}
+
+There's no faster way to get started than with our hosted {ess} on Elastic Cloud:
+
+. {ess-trial}[Get a free trial]. 
+
+.	Log into {ess-console}[Elastic Cloud].
```
nrichers, line 16: Fix this!

shared/cloud/ess-access-kibana.asciidoc
```diff
@@ -0,0 +1,26 @@
+// Include this file in your docs:
+// include::{docs-root}/shared/cloud/ess-access-kibana.asciidoc[]
+//
+// Use a different heading level, e.g. H3? 
+// :leveloffset: +1 
+// include::{docs-root}/shared/cloud/ess-getting-started.asciidoc[]
+// :leveloffset: 0
```
bmorelli25, line 7: Now that headings aren't being included in these files, is this necessary?

shared/attributes.asciidoc
```diff
@@ -87,6 +81,21 @@
...
+:ecloud:      Elastic Cloud
+:ess:         Elasticsearch Service
+:ece:         Elastic Cloud Enterprise
+:cloud:       https://www.elastic.co/guide/en/cloud/current
+:ess-trial:   https://www.elastic.co/cloud/elasticsearch-service/signup?baymax=docs-body&elektra=docs
+:ess-product: https://www.elastic.co/cloud/elasticsearch-service?baymax=docs-body&elektra=docs
+:ess-console: https://cloud.elastic.co?baymax=docs-body&elektra=docs
+:ess-baymax:  ?baymax=docs-body&elektra=docs
+:ece-ref:     https://www.elastic.co/guide/en/cloud-enterprise/current
+:ess-leadin: You can run Elasticsearch on your own hardware or use our hosted Elasticsearch Service that is available on AWS, GCP, and Azure. {ess-trial}[Try the Elasticsearch Service for free].
```
karenzone, line 96: `:ess-leadin:` is the piece I need for Logstash.  It works like a champ! 
I've opened two Logstash PRs to make use of this attribute when this work gets merged.  Thanks for your work on this, @nrichers.  <3

shared/cloud/ess-getting-started.asciidoc
nrichers: Fixed stray tab character here and in a couple other places.

shared/cloud/ess-access-kibana.asciidoc
nrichers: Yes, good point. Removed these :leveloffset: instructions here and in the other single-sourced file.

shared/attributes.asciidoc
nrichers: Glad it works, @karenzone, and let me know if you want additional text snippets!

PR #1818, opened 2020-04-27
Clean up conf.yaml

gtback:
- Add .git suffix to repo URLs for consistency
- Add quotes around section titles that contain colons
- reuse rather than redefine `mapCloudSaasToClientsTeam` variable.

The last two changes are needed to make Python's YAML library able to
load the file.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -16,16 +16,16 @@ repos:
     curator:              https://github.com/elastic/curator.git
     ecctl:                https://github.com/elastic/ecctl.git
     ecs:                  https://github.com/elastic/ecs.git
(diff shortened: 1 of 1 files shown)


conf.yaml
```diff
@@ -698,8 +698,7 @@ contents:
                 alternatives: { source_lang: console, alternative_lang: php }
                 repo:   clients-team
                 path:   docs/examples/elastic-cloud/php
-                map_branches: &mapCloudSaasToClientsTeam
-                  release-ms-34: master
+                map_branches: *mapCloudSaasToClientsTeam
```
gtback, line 701: Already defined on line 643

PR #1840, opened 2020-05-13
Update conf.yaml with new ecctl branches

karencfv:
This patch updates the ecctl config to include the master branch and the branch from latest release.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -794,7 +794,7 @@ contents:
             prefix:     en/ecctl
             tags:       CloudControl/Reference
             subject:    ECCTL
-            current:    v1.0.0-beta3
+            current:    1.0.0-beta3
             branches:   [ master, 1.0.0-beta3, 1.0.0-beta2, v1.0.0-beta3, v1.0.0-beta2 ]
             index:      docs/index.asciidoc
             chunk:      1
--- a/conf.yaml
+++ b/conf.yaml
@@ -795,7 +795,7 @@ contents:
             tags:       CloudControl/Reference
             subject:    ECCTL
             current:    v1.0.0-beta3
-            branches:   [ master, v1.0.0-beta3, v1.0.0-beta2 ]
+            branches:   [ master, 1.0.0-beta3, 1.0.0-beta2, v1.0.0-beta3, v1.0.0-beta2 ]
             index:      docs/index.asciidoc
             chunk:      1
             sources:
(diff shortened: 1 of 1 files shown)


gtback:
It looks like there's only one branch being built right now, which gets the label `current` (`https://www.elastic.co/guide/en/ecctl/current/index.html`). You can also access the same documentation using the branch name (`https://www.elastic.co/guide/en/ecctl/v1.0.0-beta2/index.html`), but there shouldn't be links directly to these pages. Once we make this change, each branch will be accessible via these types of URLs, and it will be harder to change them. Since the branch names are the same as the tags for the releases, this can cause confusion when using the git command line. Would it be OK to change the branches to just be (for example) `1.0.0-beta2` (no `v`)? We can add the branches in the ecctl repo at the same commits as their counterparts for now, then update conf.yaml to build the new branch names, then remove the branch labels (not the actual commits of course) from the ecctl repo. 

I'm trying to avoid needing to build redirects if we end up changing our minds later. I've also been talking to @nrichers about building directly from tags, which is another avenue to pursue.

karencfv:
Hi @gtback, I am following the instructions @nrichers gave us for when we release. I think naming the branch exactly as the version is intended. Can you confirm @nrichers?

I don't feel too strongly about branch names, but on this occasion we are releasing today, so it's somewhat time sensitive.

conf.yaml
```diff
@@ -794,8 +794,8 @@ contents:
             prefix:     en/ecctl
             tags:       CloudControl/Reference
             subject:    ECCTL
-            current:    v1.0.0-beta2
-            branches:   [ v1.0.0-beta2 ]
+            current:    v1.0.0-beta3
```
nrichers, line 797:
```suggestion
            current:    1.0.0-beta3
```

conf.yaml
```diff
@@ -794,8 +794,8 @@ contents:
...
+            branches:   [ master, v1.0.0-beta3, v1.0.0-beta2 ]
```
nrichers, line 798:
```suggestion
            branches:   [ master, 1.0.0-beta3, 1.0.0-beta2 ]
```

nrichers, line 798:
```suggestion
            branches:   [ master, 1.0.0-beta3, 1.0.0-beta2, v1.0.0-beta3, v1.0.0-beta2 ]
```

(Temporary fix until we can add a redirect in to the docs in `1.0.0-beta3` and `1.0.0-beta2`.)

PR #1861, opened 2020-06-04
docs: Guide order

bmorelli25:
## Summary

@KOTungseth pointed out that our solution docs are hidden away at the bottom of our current documentation landing page. To better 3+1ify the page, this PR moves our solution documentation above the Logstash and Beats reference. The page still isn't perfect, but in its current structure, it will never be.

@dedemorton asked if I could also update the order the Beats documentation while I was working with the file. Beats are now sorted alphabetically, with the Beats platform reference at the top of the list, and the Beats developer reference at the bottom.

## Screenshot

screencapture-docs-1861-docs-preview-app-elstc-co-guide-index-html-2020-06-03-20_38_30

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -837,542 +837,6 @@ contents:
                 path:   shared/attributes62.asciidoc
                 exclude_branches:   [ master, 7.x, 7.8, 7.7, 7.6, 7.5, 7.4, 7.3, 7.2, 7.1, 7.0, 6.8, 6.7, 6.6, 6.5, 6.4, 6.3, 5.4, 5.3, 5.2, 5.1, 5.0, 4.6, 4.5, 4.4, 4.3, 4.2, 4.1, 4.0, 3.0 ]
 
-    -   title:      "Logstash: Collect, Enrich, and Transport"
-        sections:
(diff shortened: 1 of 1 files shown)


bmorelli25:
Ugh, I know. The diff is pretty bad 😞

chriscressman:
@bmorelli25 - When Kibana is transformed into Elastic Console, the solutions will be ordered

1. Enterprise Search
2. Observability
3. Security

to align with the marketing copy "Search. Observe. Protect.".

So probably makes sense to use that same order here, while we're changing things?

conf.yaml
```diff
@@ -2092,6 +2072,26 @@ contents:
                 repo:   elasticsearch
                 path:   docs/Versions.asciidoc
                 exclude_branches:   [ 2.4, 2.3, 2.2, 2.1, 2.0, 1.7, 1.6, 1.5, 1.4, 1.3, 0.90 ]
+          - title:      Legacy Topbeat Reference
```
dedemorton, line 2075: Sorry, one change I missed. Let's remove "Legacy" from this title.

bmorelli25: Good find

PR #1862, opened 2020-06-05
Add attributes for ingest management

dedemorton:
Adds attributes (variables) for names related to new ingest management capabilities. 

@mostlyjason Can you verify the names here? Not sure if I got these right, although I did read through 3 different issue.  I couldn't figure out the final decision on a couple names, like the package registry. :-) 

@skh After we merge this PR, please use these attributes in place of names in the ingest management docs. For example, instead of Elastic Agent, use `{agent}`.

The Kibana Reference already includes the file. To include the file in new docs, use:

`include::{docs-root}/shared/versions/stack/{source_branch}.asciidoc[]`

### DIFF
--- a/shared/attributes.asciidoc
+++ b/shared/attributes.asciidoc
@@ -122,7 +122,7 @@ Ingest terms
 :agent: Elastic Agent
 :fleet: Fleet
 :ingest-manager: Ingest Manager
-:package-manager: package manager
+:package-manager: Elastic Package Manager
 :integrations: Integrations
 :package-registry: Elastic Package Registry
 
--- a/shared/attributes.asciidoc
+++ b/shared/attributes.asciidoc
@@ -116,6 +116,17 @@ Kibana app names
 :siem-app: SIEM app
 
 //////////
+Ingest terms
+//////////
+
+:agent: Elastic Agent
+:fleet: Fleet
+:ingest-manager: Ingest Manager
+:package-manager: package manager
+:integrations: Integrations
+:package-registry: Elastic Package Registry
+
+//////////
 Common words and phrases
 //////////
 :stack:           Elastic Stack


mostlyjason:
@dedemorton thanks I need to follow up on naming for the package registry let me kick off an email. I think integrations can be lowercase except when referring to the Integrations page? Also, in what context would you use the term "package manager?" I don't think that phrase is shown in the UI anywhere.

mostlyjason:
I checked our old naming issue and it looks like we decided to call it Elastic Package Registry and not spend a lot of time thinking about it since it's mostly an internal name. https://github.com/elastic/observability-dev/issues/659

This name does show up in a few places where users are creating their own packages or when hosting their own local registries. I think it's advanced use case though most customers don't need to know about. We could think about renaming it but I'm not sure on the cost vs. benefit.

I would recommend removing package manager since its not used anywhere.

dedemorton:
There's a setting in Kibana to enable "EPM." What should we use instead of EPM? (Using abbreviations is not recommended unless they are common, industry-standard abbreviations.)

See https://github.com/elastic/kibana/pull/66736/files#diff-e4b2f7481ede0e997ab0d290ac183acfR22

dedemorton:
@mostlyjason I've changed Elastic Package Manager to match the Kibana settings, but I expect we will update this later before 7.8 ships. I mainly just need to get this merged so that we can use the attributes in the docs we're writing this week. Then we can just update the shared file before 7.8 and all the docs will be updated accordingly. Sound fair?

mostlyjason:
@dedemorton sounds good I approved it for now. I also started a conversation on the other issue you mentioned to remove any mention of Elastic Package Manager. Hopefully we can swing back to remove this later.

PR #1873, opened 2020-06-23
Sets apm branches for 7.x to master

dedemorton:
APM doesn't build the 7.x branch, so we need to set the URL to master to avoid breaking the build.

### DIFF
--- a/shared/attributes.asciidoc
+++ b/shared/attributes.asciidoc
@@ -32,6 +32,14 @@
 :apm-server-ref-62:    https://www.elastic.co/guide/en/apm/server/6.2
 :apm-server-ref-64:    https://www.elastic.co/guide/en/apm/server/6.4
 :apm-server-ref-70:    https://www.elastic.co/guide/en/apm/server/7.0
+///////
+APM does not build n.x documentation. Links from .x branches should point to master instead
+///////
+ifeval::["{source_branch}"=="7.x"]
+:apm-server-ref:       {apm-server-ref-m}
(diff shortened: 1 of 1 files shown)


shared/attributes.asciidoc
```diff
@@ -32,6 +32,14 @@
 :apm-server-ref-62:    https://www.elastic.co/guide/en/apm/server/6.2
 :apm-server-ref-64:    https://www.elastic.co/guide/en/apm/server/6.4
 :apm-server-ref-70:    https://www.elastic.co/guide/en/apm/server/7.0
+///////
+APM does not build n.x documentation. Links from .x branches should point to master instead
+///////
+ifeval::["{source_branch}"=="7.x"]
+:apm-server-ref:       {apm-server-ref-m}
+:apm-server-ref-v:     {apm-server-ref-m}
+:apm-overview-ref-v:   {apm-overview-ref-m}
```
bmorelli25, line 41: We should also add the old `get-started` attribute

```suggestion
:apm-overview-ref-v:   {apm-overview-ref-m}
:apm-get-started-ref:  {apm-overview-ref-m}
```

dedemorton: Ruh roh. I merged prematurely. I'll put up another PR for this.

bmorelli25: I don't think you need to. I'm pretty sure that attribute was only still used in the Beats repo--and you just removed it!

PR #1874, opened 2020-06-24
Set branch for apm-get-started-ref attribute

dedemorton:
Missed this one in https://github.com/elastic/docs/pull/1873


dedemorton:
At some point, we should a full search and remove old attributes. But it's better to have this in until we do (just in case someone like me bumbles along and uses the old one).

gtback:
I'd love to enable Asciidoctor's warnings for missing attributes:

https://github.com/elastic/docs/blob/bc6f559330bc649694b1f6ccd6848c857cc14023/lib/ES/Util.pm#L89-L91

PR #1875, opened 2020-06-24
Add kibana dependency to Glossary

lcawl:
Related to https://github.com/elastic/kibana/pull/69721

This PR adds the kibana repo as a resource for the Glossary so that it can pull tagged sections from that repo.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -130,10 +130,10 @@ contents:
                 path:   docs/en
               -
                 repo:   elasticsearch
-                path:   docs/reference
+                path:   docs/reference/glossary.asciidoc
               -
                 repo:   kibana
-                path:   docs
+                path:   docs/glossary.asciidoc
               -
                 repo:   docs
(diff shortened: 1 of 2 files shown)


conf.yaml
```diff
@@ -131,6 +131,9 @@ contents:
               -
                 repo:   elasticsearch
                 path:   docs/reference
+              -
+                repo:   kibana
+                path:   docs
```
gtback, line 136: This is OK as is, but if we don't want to rebuild the glossary anytime *any* of the Kibana docs change, we can use this instead:
```suggestion
                path:   docs/glossary.asciidoc
```

Do I understand that correctly, @nik9000 ?

lcawl: Ah, if that's the case I'll make the same change to the Glossary's elasticsearch dependency. It only cares about one file there too.

lcawl:

PR #1893, opened 2020-07-14
Add ECK 1.2 

charith-elastic:
This adds the ECK 1.2 branch to the build list. ECK 1.2.0 is planned for release on July 21. 

I followed #1799 and added 1.2 to the exclude list for `7.x`. I hope that is still the correct thing to do as 1.2 also uses `current.asciidoc` (https://raw.githubusercontent.com/elastic/cloud-on-k8s/1.2/docs/index.asciidoc)

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -776,7 +776,7 @@ contents:
             tags:       Kubernetes/Reference
             subject:    ECK
             current:    1.1
-            branches:   [ master, 1.1, 1.0, 1.0-beta, 0.9, 0.8 ]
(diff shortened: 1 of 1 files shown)


gtback:
@elastic/cloud-writers can you review this?

charith-elastic:
@gtback I think what you are suggesting makes sense. We should pin the stack version used by older ECK releases because some of the newer stack applications probably wouldn't work on those. I will do that in a separate PR.

gtback:
Thanks @charith-elastic 👍

Issue #1212, opened 2019-09-24
Inline pop-ups for glossary terms


debadair:
It would be nice to be able to display glossary definitions as inline pop-ups, rather than having to click a link to the separate glossary pages. (@kellyemurphy)

bmorelli25:
Unasigning myself as we've closed #1894.

bmorelli25:
Agh, sorry. Zube bot is awful.

PR #1894, opened 2020-07-14
Inline popups for glossary terms widget

bmorelli25:
## Summary

This PR adds inline pop-ups for glossary terms to the documentation build.

**GIF (outdated style)**
ezgif com-video-to-gif

**Hover (updated style)**
Screen Shot 2020-07-16 at 3 31 23 PM

**No hover**
Screen Shot 2020-07-14 at 11 30 46 AM

## How to use the widget

The syntax for using this widget is: `definition:[word,definition]`. For example:
```asciidoc
definition:[snapshots,A backup taken from a running Elasticsearch cluster. A snapshot can include backups of an entire cluster or only data streams and indices you specify.]
```
Definitions can be assigned to attributes and used anywhere the attribute is defined:
```asciidoc
// definition
:snapshots-def: definition:[snapshots,A backup taken from a running Elasticsearch cluster. A snapshot can include backups of an entire cluster or only data streams and indices you specify.]

// use
This example shows you how to register a shared file system repository and store {snapshots-def}. Before you begin, you must register the location of the repository in the path.repo setting on your master and data nodes. You can do this in one of two ways:
```

All a user needs to do is hover over the word to display a definition popup 💥 🧙 

## Limitations
* This widget is for inline use only. Block didn't seem necessary, and we can always iterate if needed.
* Because tagged regions resolve _after_ these widgets are created by the build, tagged regions cannot be used within a definition block 😢 
* ~Commas cannot be used in definitions 😢~ They do work! @lcawl pointed me [here](https://wiki.elastic.co/display/DOC/Troubleshooting+documentation+errors#Troubleshootingdocumentationerrors-Added,coming,anddeprecated). Sure enough, wrapping the definition in `"definition"` allows the use of commas.

## How does it work?

* The popup functionality was extended from the `change_admonition` asciidoctor extension: https://github.com/elastic/docs/blob/master/resources/asciidoctor/lib/change_admonition/extension.rb
* The style of the popup was adapted from the `admonishment.pcss` style sheet:
https://github.com/elastic/docs/blob/master/resources/web/style/admonishment.pcss
* Conversion to HTML occurs much in the same way as admonitions, but with different class names: https://github.com/elastic/docs/blob/master/resources/asciidoctor/lib/docbook_compat/convert_admonition.rb

## Related PRs
Closes https://github.com/elastic/docs/issues/1212

## Todo

- [x] Clean up code
- [x] Better CSS
- [x] Accessibility
- [x] Add tests
- [x] Add documentation

## Glossary working group crew?

cc @debadair @gchaps @lcawl


bmorelli25:
@gtback. Is this just saying I need to remove variables that I don't use? I'm guessing that's happening because this is s simpler implementation of the ChangeAdmonition.

```
11:36:16 Offenses:
11:36:16 
11:36:16 lib/definition_hover/extension.rb:19:32: W: Lint/UnusedBlockArgument: Unused block argument - revisionflag. If it's necessary, use _ or _revisionflag as an argument name to indicate that it won't be used.
11:36:16     MACRO_CONF.each do |(name, revisionflag, tag, message, title_class)|
11:36:16                                ^^^^^^^^^^^^
11:36:16 lib/definition_hover/extension.rb:19:46: W: Lint/UnusedBlockArgument: Unused block argument - tag. If it's necessary, use _ or _tag as an argument name to indicate that it won't be used.
11:36:16     MACRO_CONF.each do |(name, revisionflag, tag, message, title_class)|
11:36:16                                              ^^^
11:36:16 lib/definition_hover/extension.rb:40:17: C: Style/UnneededInterpolation: Prefer to_s over string interpolation.
11:36:16       message = "#{@message}" + attrs[:text] if attrs[:text]
11:36:16                 ^^^^^^^^^^^^^
11:36:16 lib/docbook_compat/convert_admonition.rb:58:5: C: Metrics/MethodLength: Method has too many lines. [21/10]
11:36:16     def convert_inline_admonition_for_real(node) ...
11:36:16     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
11:36:16 lib/docbook_compat/convert_admonition.rb:59:1: C: Layout/EmptyLinesAroundMethodBody: Extra empty line detected at method body beginning.
```

resources/asciidoctor/lib/docbook_compat/convert_admonition.rb
```diff
@@ -56,16 +56,32 @@ def skip_inline_admonition(node)
...
+        #   "Definition-#{node.attr 'title_type'} #{node.attr 'title_class'}"
+        [
+          %(<span class="Definition Definition--#{node.type}">),
+          %(<span class="Definition-word">#{node.attr 'title'}</span>),
+          '<span class="Definition-defined">',
+          node.text,
+          '</span>',
+          '</span>',
+        ].join "\n"
+      # For all other admonitions, add custom admonishment style
```
bmorelli25, line 72: I'm not sure if adding an `if` statement here is the right way to go about this. Would it be better to have a new `dockbook_compat` file named `convert_definition`?

dedemorton:
This is very cool! Would this replace the glossary completely, or just provide a short definition?

If we're providing short definitions and plan to maintain a glossary, too, I wonder if we can define the short definition as a shared attribute that gets referenced. I'm can't remember the precedence for angle brackets vs curly braces (sorry I'm being lazy), but can we do this?

```
:snapshots-def: definition:[snapshots,{snapshots-short-def}]`
```

Then in the glossary, we would have something like:

```
snapshots::
{snapshots-short-def}. Blah blah blah and more blah.
```

bmorelli25:
> This is very cool! Would this replace the glossary completely, or just provide a short definition?

I don't think it would replace it, but perhaps @gchaps or @debadair can comment on that.

> I wonder if we can define the short definition as a shared attribute that gets referenced.

I just mocked up a quick test of your example locally and it works!

gchaps:
This is great!

> This is very cool! Would this replace the glossary completely, or just provide a short definition?

We still need a glossary, where readers can see all terms in one place.  

@bmorelli25 Does this meet accessibility requirements?

bmorelli25:
> @bmorelli25 Does this meet accessibility requirements?

What are our accessibility requirements?

myasonik:
@bmorelli25 A little simplified, but essentially it's to ensure everything is available to both keyboard-only and screen reader users.

If you're looking for official requirements and specs, we shoot for [WCAG AA](https://www.w3.org/TR/WCAG21/) (warning: WCAG specs are dense, to say the least).

If you need any help making the tooltips more accessible or want help testing, let me know. As a starter, you can use [EUI's tooltip implementation](https://elastic.github.io/eui/#/display/tooltip) as an example and/or refer to [this recent blog post](https://sarahmhigley.com/writing/tooltips-in-wcag-21/#best-practices-summary) for a broad and thorough in-depth dive on all things tooltip-related.

debadair:
This is cool! At least for the time being, I don't think this would replace the Glossary in the docs. 

I use short definitions for some of the ES terms already, but use tagged regions to do it. Switching to the attribute approach @dedemorton suggested is certainly possible. (And I'm not very enamored of using tagged regions for re-use. It's really clunky.)

We'll have to take this into account in the efforts to get to "one glossary to rule them all". If our glossary tooltips rely on short defs defined as asciidoc attributes, we'll have to figure out how/where the short and long defs are maintained and pulled into the appropriate places. @gchaps might know more about the current glossary thinking. 

This is a nice enhancement that shouldn't be held up by how we might reuse the content later, though. We *should* decide where they should live and agree on a naming convention.

dedemorton:
I used VoiceOver to do a quick accessibility test of our docs (not Brandon's glossary widget...just some random pages). In general, our docs have a lot of accessibility challenges. We don't make it easy for folks using screen readers. I wonder if we can get Brandon's glossary widget added as an "experimental" feature in one or two books and get some user feedback? IMO we shouldn't require full accessibility when our docs themselves need much love in that area. On the other hand, we don't want anything that makes accessibility _worse_ so we'll need to do some minimal testing.

myasonik:
Yeah, the docs site and elastic.co have some significant shortcomings but continuing to dig our hole seems dangerous too. If we're already doing some work, best make the new features accessible. 

If @bmorelli25 investigates and it's prohibitively difficult for some reason, I'd agree that we probably don't need to stop everything to do it but I'm optimistic that it won't be too bad 🤞

gtback:
> I'm not very enamored of using tagged regions for re-use. It's really clunky.

Agree 💯% 

I like the idea of reusing the content as much as possible. To the extent that we're willing to have the tooltip just show the first "paragraph" of the full glossary entry, we might be able to get away with not maintaining two defs for a term. I'm at least glad we have a single (unversioned) Glossary book.

@bmorelli25 does this handle hyperlinks within the definitions?

resources/asciidoctor/lib/definition_hover/extension.rb
```diff
@@ -0,0 +1,49 @@
...
+    name_positional_attributes :version, :text
+    format :short
+
+    def initialize(message, extra_title_class)
+      super(nil)
+      @message = message
+      @extra_title_class = extra_title_class
+    end
+
+    def process(parent, _target, attrs)
```
bmorelli25, line 38: Based on examples in the [asciidoctor reference](https://asciidoctor-docs.netlify.app/asciidoctor/1.5/api/block-processor/), we should be able to access a `_target.lines` function that returns the content underneath a custom-defined block. I'm not sure why that doesn't work here. It returns an empty array.

bmorelli25:
> @bmorelli25 does this handle hyperlinks within the definitions?

@gtback Surprisingly, it does! Links render and work correctly:

---

Screen Shot 2020-07-15 at 8 32 38 AM

bmorelli25:
```
09:55:16 Offenses:
09:55:16 
09:55:16 lib/docbook_compat/convert_admonition.rb:58:5: C: Metrics/MethodLength: Method has too many lines. [21/10]
09:55:16     def convert_inline_admonition_for_real(node) ...
09:55:16     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
09:55:16 
09:55:16 70 files inspected, 1 offense detected
09:55:16 Makefile:12: recipe for target 'rubocop' failed
```

OK, so it looks like [The Ruby Style Guide](https://github.com/bbatsov/ruby-style-guide) says "Avoid methods longer than 10 LOC (lines of code). Ideally, most methods will be shorter than 5 LOC. Empty lines do not contribute to the relevant LOC.", and that's why RuboCop has 10 as the default limit.

Any recommendations as to how to reduce the size of `def convert_inline_admonition_for_real(node)`?

bmorelli25:
Met with @gtback who helped me clean up the code considerably. Thanks! I'll tackle tests later this week.

resources/asciidoctor/lib/definition_hover/extension.rb
bmorelli25: Resolved with the help of @gtback. Accessing block content only works within a BlockProcessor. See docs [here](https://www.rubydoc.info/github/asciidoctor/asciidoctor/Asciidoctor/Extensions/BlockProcessor), and an example below:

```
class Shout < Asciidoctor::Extensions::Group
  def activate(registry)
    registry.block ShoutBlock.new :shout
  end

  ##
  # Inline change admonition.
  class ShoutBlock < Asciidoctor::Extensions::BlockProcessor
    use_dsl

    # Stores content passed in from asciidoc in `attrs[:x]`
    # name_positional_attributes :input_word, :input_def
    name_positional_attributes 'vol'
    parse_content_as :simple

    # I don't know what this does, but it's necessary for this widget to work
    # format :short

    def process(parent, _target, attrs)
      # Access attributes passed in from raw asciidoc
      input_word = attrs[:input_word]
      message = attrs[:input_def]

      # on_context :paragraph
      # name_positional_attributes 'vol'


      volume = ((attrs.delete 'vol') || 1).to_i
      create_paragraph parent, (_target.lines.map {|l| l.upcase}), attrs

      # Create a new line block
      # Asciidoctor::Block.new(
      #   parent, :admonition, message, type: 'definition', attributes: {
      #     # 'input_word' => input_word,

      #     # These aren't needed. I'll delete them soon enough
      #     # 'word' => @word,
      #     # 'definition' => @definition,
      #   }
      # )
    end
  end
end
```

resources/asciidoctor/lib/docbook_compat/convert_admonition.rb
bmorelli25: Removed `if` statement

gtback:
@bmorelli25 Let me know if you want me to take a pass at reviewing now or wait until there are tests.

bmorelli25:
You can hold off for now, @gtback. I'll remove the draft status when I'm ready.

gtback:
@bmorelli25 feel free to request a review when you're ready. I just un-assigned myself to get it out of my queue 😁

gtback:
Another suggestion: If you add an example of this in README.asciidoc, it's a good way to explain how to use it an also make it easy for people to test.

bmorelli25:
> Another suggestion: If you add an example of this in README.asciidoc, it's a good way to explain how to use it an also make it easy for people to test.

Beat you to it by about 30 seconds 😄

gtback:
> Beat you to it by about 30 seconds 😄

Wow, impressive. It's like you read my mind (also, scary!)

resources/asciidoctor/spec/definition_hover_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+          <<~ASCIIDOC
+            Words #{key}:[#{word},] words.
+          ASCIIDOC
+        end
+        it "renders with Elastic's custom template" do
+          expect_definition_popup(
+            word.to_s, ''
+          )
+        end
+      end
```
bmorelli25, line 66: This should really produce an error, but I'm not sure how to make that happen. A word and definition should be required

bmorelli25:
@gtback Following up: Is there anything you need from me on this before you can review?

gtback:
Nope, sorry. I'll look at this this afternoon.

README.asciidoc
```diff
@@ -1375,6 +1375,32 @@ https://github.com/elastic/docs/blob/master/conf.yaml[`conf.yaml`].
...
+// Assign the definition to an attribute
+:run-short-def: To move at a speed faster than a walk. See <<run-link,run>> to learn more.
+
+// Create a definition pop-up and assign it to an attribute
+:run-def: definition:[run,"{run-short-def}"]
+
+// Reference the definition attribute to create a pop-up in the docs
+run::
+{run-short-def}. I like to {run-def}.
+----
```
gtback, line 1402: Maybe also include this as actual Asciidoc (outside the `[source]` tag) so people previewing the README can see it.

README.asciidoc
```diff
@@ -1375,6 +1375,32 @@ https://github.com/elastic/docs/blob/master/conf.yaml[`conf.yaml`].
...
 changes to the referenced repository don't change the referenced path.
 
+[[glossary-pop-ups]]
+== Inline pop-ups for glossary terms
+
+Rather than linking to glossary pages, you can use an inline definition pop-up.
+
+[source,asciidoc]
+----
+definition:["word", "definition"]
```
gtback, line 1385: Should the default example include quotes, given that (in this case) they aren't needed?

resources/asciidoctor/lib/definition_hover/extension.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+require 'asciidoctor/extensions'
+
+##
+# Extension for adding an inline definition to key terms in the docs
+#
+# Usage
+#
+#   Foo definition:["word", "definition"]
```
gtback, line 10: I wouldn't use "definition" in the example, since it might confuse people as to what is expected to be copied and what people can change.
```suggestion
#   This is a paragraph about definition:[Elasticsearch, "a distributed, JSON-based search and analytics engine"].
```

resources/asciidoctor/lib/definition_hover/extension.rb
```diff
@@ -0,0 +1,41 @@
...
+    format :short
+
+    def process(parent, _target, attrs)
+      # Access attributes passed in from raw asciidoc
+      input_word = attrs[:input_word]
+      message = attrs[:input_def]
+
+      # Create a new line block
+      Asciidoctor::Inline.new(
+        parent, :admonition, message, type: 'definition', attributes: {
```
gtback, line 35: What is `:admonition` used here for?

gtback: Oh, I think maybe we need to move `convert_inline_definition_for_real` into a separate module under `docbook_compat`. A definition is distinct from an admonition, right?

resources/asciidoctor/lib/docbook_compat/convert_admonition.rb
```diff
@@ -55,6 +59,20 @@ def skip_inline_admonition(node)
       end
     end
 
+    # If creating a defintiion admonition, maintain normal style
```
gtback, line 62:
```suggestion
    # If creating a definition admonition, maintain normal style
```

resources/asciidoctor/lib/docbook_compat/convert_admonition.rb
```diff
@@ -55,6 +59,20 @@ def skip_inline_admonition(node)
...
+    def convert_inline_definition_for_real(node)
+      [
+        %(<span class="Definition Definition--#{node.type}" \
```
gtback, line 65: I don't know if we need the `node.type` here. That's more for Admonitions: the same reason we got rid of that MACRO_LOOP (we only have one type).

Also, I personally don't like capital letters in CSS class names, but that's possibly just a personal preference.

resources/asciidoctor/spec/definition_hover_spec.rb
gtback: Can we have the `Definition.process` method raise an error if either argument is a blank string? I don't know what built-in errors Asciidoctor defines, but it would be best to follow an existing pattern there.

resources/web/style/definition_hover.pcss
```diff
@@ -0,0 +1,36 @@
+#guide {
+  .Definition {
+    position: relative;
+
+    &:hover {
+      color: black;
+      transition: all 0.3s ease;
+      .Definition-defined {
+        z-index: 10000;
```
gtback, line 9: How much of these styles are borrowed from elsewhere? Is there an opportunity to reuse them so they aren't specific to this element?

bmorelli25:
After today's doc redesign update, I'm not sure it makes sense to continue working on this. I'm going to close this out. If anyone disagrees, please let me know.

gtback:
If this isn't urgently needed, then I agree we can wait an re-implement this functionality in the new system.

PR #1902, opened 2020-07-15
Add attribute for eck html path

dedemorton:
I need this attribute to resolve links in the Beats documentation that point to the ECK docs.

### DIFF
--- a/shared/attributes.asciidoc
+++ b/shared/attributes.asciidoc
@@ -108,7 +108,7 @@ Elastic Cloud
 :ess-console: https://cloud.elastic.co?baymax=docs-body&elektra=docs
 :ess-baymax:  ?baymax=docs-body&elektra=docs
 :ece-ref:     https://www.elastic.co/guide/en/cloud-enterprise/current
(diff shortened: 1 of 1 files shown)


shared/attributes.asciidoc
```diff
@@ -108,6 +108,7 @@ Elastic Cloud
 :ess-console: https://cloud.elastic.co?baymax=docs-body&elektra=docs
 :ess-baymax:  ?baymax=docs-body&elektra=docs
 :ece-ref:     https://www.elastic.co/guide/en/cloud-enterprise/current
+:eck-ref:     https://www.elastic.co/guide/en/cloud-on-k8s/current/
```
david-kow, line 111: Should we remove trailing `/` for consistency with `ece-ref` above?

dedemorton: good catch. I goofed.

dedemorton, line 111:
```suggestion
:eck-ref:     https://www.elastic.co/guide/en/cloud-on-k8s/current
```

PR #1906, opened 2020-07-16
Add 7.10 attributes

jrodewig:
Updates the 7.x versions file to reference 7.10.
Creates a 7.10 versions file.


shared/versions/stack/7.x.asciidoc
```diff
@@ -1,14 +1,14 @@
-:version:                7.9.0
+:version:                7.10.0
```
gtback, line 1: This got me thinking, can we just make this whole file `include::7.10.asciidoc[]`?

gtback: Ignore this, `branch` is different

Issue #220, opened 2017-08-03
Doc build should show error when asciidoc attributes are not resolved


dedemorton:
Right now, if the doc build cannot resolve an asciidoc attribute, the attribute and the text that follows it are dropped from the docs. We've had multiple situations where content was missing from the doc, but we didn't know until a user or someone internal noticed the problem.

clintongormley:
This is just how asciidoc works, unfortunately.  If an attribute is undefined, then the line that contains it is ignored.

I don't know of a way around this.

dedemorton:
@clintongormley Then maybe we need to find a solution outside of asciidoc.  Even a rudimentary script that does something like the following would uncover problems.

For each book:

1. search for all `:attributes:` and compile a list
2. search for all `{attributes}` and compile a list
3. For each item in the second list, make sure a corresponding definition exists in the first list

It wouldn't be 100% accurate because it wouldn't take order into consideration, but it would pick up most, if not all, unresolved attribute references.

Is this feasible? If this is too heavy-weight to run for every build, I'd even be happy to have a script that I can run locally.

debadair:
Another check to add to the attributes preflighting: make sure there are no attribute references within a bulleted list embedded in a table. They do not resolve & the list items will be blank.

gtback:
Asciidoctor has an [`attribute-missing` option](https://asciidoctor.org/docs/user-manual/#handle-missing-or-undefined-attributes) to control what we want to do. It seems like the Asciidoctor default is to leave the reference intact (`skip`), so it should be easier to notice them. The default `asciidoc.py` behavior was to drop the entire line, which of course is not desirable. 

There is [code to warn on missing attributes](https://github.com/elastic/docs/blob/5e9e844142c9b62f162d7dd12321342a49a817e6/lib/ES/Util.pm#L89), but it's currently commented out. I recommend re-enabling that, at least for branches that we're still actively maintaining, and find and fix any warnings in those branches.

lcawl:
I can confirm that unresolved attributes are now left in the docs (e.g. https://github.com/elastic/elasticsearch/pull/44249).  This is better than the previous behaviour but anything we can do to make it even more obvious when this happens would be great!

gtback:
This popped up again today in some old branches. I'll prioritize getting this fixed, at least for live branches.

dedemorton:
I'm guessing this won't be fixed in the legacy docs. As it's quite old and I'm tired of seeing it in my list, I'm closing it.

PR #1911, opened 2020-07-21
Remove unused docs constants

przemekwitek:
This PR removes constants that are not used anywhere.

Relates https://github.com/elastic/elasticsearch/issues/59947

### DIFF
--- a/shared/attributes.asciidoc
+++ b/shared/attributes.asciidoc
@@ -258,8 +258,6 @@ Common words and phrases
 :fiscore:                    feature influence score
 :evaluatedf-api:             evaluate {dfanalytics} API
 :evaluatedf-api-cap:         Evaluate {dfanalytics} API
-:binarysc:                   binary soft classification
-:binarysc-cap:               Binary soft classification
 :regression:                 regression
 :regression-cap:             Regression
 :reganalysis:                regression analysis


droberts195:
I am not sure you can merge this.  Old versions of the docs will reference the constant you've removed, and it looks like this repo doesn't have a separate branch for each stack minor release.  I think merging this would break the docs builds for 7.6, 7.7, 7.8 and 7.9.

przemekwitek:
> I am not sure you can merge this. Old versions of the docs will reference the constant you've removed, and it looks like this repo doesn't have a separate branch for each stack minor release. I think merging this would break the docs builds for 7.6, 7.7, 7.8 and 7.9.

Actually these constants were *not* used in out evaluate `asciidoc` before my renaming PR. Still, you may be right that they were used at some point (for at least one minor release) and it's better not to touch them.
Maybe @lcawl will be able to confirm.

gtback:
@lcawl would know for sure. We also have an open issue (#220) to make the build fail for undefined attributes.

lcawl:
@droberts195 is right.  It's used in pages like this: https://www.elastic.co/guide/en/machine-learning/7.6/ml-dfa-overview.html and https://www.elastic.co/guide/en/machine-learning/7.6/ml-dfanalytics-evaluate.html

lcawl:
You're also right that this repo isn't versioned. When there has been a significant change in the shared attributes, we've split a new version of that file (e.g. around 6.2 we split off https://github.com/elastic/docs/blob/master/shared/attributes62.asciidoc), but I don't think this particular change warrants that.

gtback:
At most, I think we can add a comment that these attributes aren't in use after 7.6. We don't regularly rebuild the 7.6 docs, and aren't making changes to them anymore, but we will occasionally rebuild all the docs, and the 7.6 docs will be "broken" (at least, these strings will be missing) once that happens.

przemekwitek:
> We don't regularly rebuild the 7.6 docs, and aren't making changes to them anymore, but we will occasionally rebuild all the docs, and the 7.6 docs will be "broken" (at least, these strings will be missing) once that happens.

Thanks for explanation. Of course, I'm not going to risk breaking the docs. Let's just drop this PR altogether.

Issue #1850, opened 2020-05-21
Support mapping branch names to different versions


gtback:
Currently, there's an assumption that every branch that gets built is the same as the version number we want to show in the built documentation. We'd like to break that dependency.

Introduce a `branch_version_mapping` key in each book in conf.yaml, and add support to the Perl script to look at this key. For example:

```yaml
branch_version_mapping:
  branch-1: 1.1
  branch-2: 2.0
  legacy-0.7: 0.7
```

If this isn't included, then we'll default back to the current behavior of having the version and branch name match.

gtback:
This can be closed now that #1918 is merged! 🎉

PR #1912, opened 2020-07-22
Add yamale to enforce (and document) a schema for conf.yaml

gtback:
schema.yaml lists the keys and their expected values, along with
comments to describe many of the settings in conf.yaml.

This is incorporated into `make check`.


gtback:
I do want to add more detail to the schema, but this is a good start and will give me a basis to work on adding a new setting in #1850 .

PR #1914, opened 2020-07-23
[DOCS] Adding shared checkmark and crossmark icons.

lockewritesdocs:
Adds a green checkmark and gray crossmark icon that we can reuse across teams.


lockewritesdocs:
@bmorelli25, I think `yes` and `no` still works. I'll make that change 👍

shared/attributes.asciidoc
```diff
@@ -318,6 +318,12 @@ Common words and phrases
 :multi-arg:     †footnoteref:[multi-arg,This parameter accepts multiple arguments.]
 :multi-arg-ref: †footnoteref:[multi-arg]
 
+//////////
+Reusable images
+//////////
+:y:          image:images/icon-yes.png[Yes,20,15]
```
gtback, line 324: Would it be possible to make these `:yes-icon:` and `:no-icon:`? Just a bit easier to find across the docs than single-letter variables.

lockewritesdocs: Probably, yes. I'll make that change as well. Good suggestion 👍

PR #1919, opened 2020-07-27
Add attributes for legacy books

gtback:
We're going to change the URLs of these books, so it helps to have them in one place.


lcawl:
I think it's likely a good idea to add those attributes in https://github.com/elastic/docs/blob/master/shared/attributes62.asciidoc as well, since that was the predecessor to attributes.asciidoc

lcawl:
I think it also might be worthwhile adding an attributes1x.asciidoc (or some such name) for use in the early branches that don't currently currently use shared attributes--and only include these two new attributes in that file.  Otherwise, I worry that we'll override existing behaviour in the old branches.

gtback:
Thanks, @lcawl. I moved them to a separate file.

Issue #1917, opened 2020-07-27
Add paths to Kibana docs


gtback:
This was originally done in #1904 but reverted in 415efb1fc0fe5c9ce75242fc86b518132df31560. See if we can reproduce the issue and get this merged in.

stacey-gammon:
Blocking https://github.com/elastic/kibana/pull/72292

gtback:
Since elastic/kibana#72292 is now passing the docs build, and I haven't the [Kibana PR Docs builds](https://elasticsearch-ci.elastic.co/job/elastic+kibana+pull-request+build-docs/) all seem to be passing recently (when they were failing before), I'm going to close this issue. I'll reopen if there's more weirdness.

PR #1924, opened 2020-07-28
Fix issue with `:(glob)` in sparse-checkout files

gtback:
These were added in 69e747baa8653201b3589869bb2ab9c9dc2e7b87, and are
necessary to detect all the right files, but we don't want to pass that
flag in the sparse-checkout file.

Related to #1917


gtback:
cc: @stacey-gammon this should fix the Kibana PRs once merged.

gtback:
>  I miss the =~ operator 😄

It's really growing on me. I don't know what that says about my future.

PR #1939, opened 2020-08-16
Adjust CSS to make the heading hierarchy clearer and fix spacing issues

debadair:


dedemorton:
Found some things that didn't look quite right. CSS inheritance is such a big PITA....get out your rubber mallet for some whack-a-mole:

- The relative size of headings looks good down to  good down to H4, but the difference between H4 and H5 is dramatic. It especially stands out in the beats config content where the formatting is, shall we say, rather eccentric (was written before we could add anchors in definition lists).

- There's too much space between items in callout lists:
image

- Having an example after a numbered list items adds extra space:
image

- Looks like there's also some extra space in some bulleted lists (again when there is an example):
image

dedemorton:
Oh found another one while I was working on other docs. Alignment of the icon in the admonition blocks isn't right:

image

debadair:
Thanks, Dede! There shouldn't be that dramatic an effect between h4 & h5s--that sounds like I didn't save a change I thought I saved. The last one is puzzling, as the only change I made there was to the H3 style if a heading is used within an admonition. The list one is related to the spacing around code blocks I think, which is really...interesting. That's probably also related to the callouts. CSS inheritance is always a pain, but the way things are done for "special" cases tends to impact normal things in unexpected ways.

gtback:
@debadair is there anything I can do to help with this? I'm guessing we don't want to put a ton of work into refreshing the current design, but small-ish changes like this are definitely welcome 😁 .

dedemorton:
@debadair I'm going to remove my name as reviewer so this doesn't show up on my list of pending reviews. Feel free to add me back if you decide to move forward with this. thanks!

gtback:
@debadair is this still something we want to try to get merged in?

PR #1940, opened 2020-08-17
Update ecctl doc branches and title

marclop:
## Description

Updates the current branch to `1.0`, adds `1.0` in the branches array,
and updates the title to include Elasticsearch Service.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -794,12 +794,12 @@ contents:
               -
                 repo:   docs
                 path:   shared/attributes.asciidoc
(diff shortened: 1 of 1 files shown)


marclop:
The build will fail since https://github.com/elastic/ecctl/pull/364 needs to be merged and the `1.0` branch cut.

marclop:
Yeah I thought about about it, it'd probably make sense to remove the pre-GA after the next major as well, giving people time to upgrade although I can't see many people holding to old versions to be honest.

PR #1944, opened 2020-08-20
Make ECS 1.6 current

webmat:
Do not merge until ECS release day :-)

cc @ebeahan


webmat:
Thanks @gtback! Yes we always kick off a full rebuild after these :-)

PR #1957, opened 2020-09-16
Prefer remote repos that match the local dir name

tvernum:
If the local git project has multiple remotes from preferred
organizations, we should attempt to use one that matches the basename
of the working (mounted) directory. This repo is more likely to be the
preferred upstream repo

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -158,6 +158,8 @@ sub _guess_opts {
     my $repo_name = _guess_repo_name( $remote );
     # We couldn't find the top level so lets make a wild guess.
     $toplevel = $index->parent unless $toplevel;
+    printf "Guessed toplevel=[%s] remote=[%s] branch=[%s] repo=[%s]\n", $toplevel, $remote, $branch, $repo_name;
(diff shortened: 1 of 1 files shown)


tvernum:
This is the fix I need for the issue I ran into (reported offline).
It turns out that I had 2 remotes referencing `github.com/elastic/` repositories, and the script was guessing the wrong one.

It needed to guess `github.com/elastic/elasticsearch`, otherwise it wouldn't define `elasticsearch-root`, and includes would fail.

This fix isn't perfect, it probably won't solve the problem if the docs are being generated against a worktree like `elasticsearch-7.x` _and_ there's multiple matching remotes, but it covers the most likely case.

I also added a small debug logging for the guesses, since that would have explained the problem I ran into.

gtback (approved):
I don't completely understand how all this guessing works, and at first I was worried this might affect how the main docs are built. But looking closer, it seems like it's more of an edge case, specific to some developer-local workflows, and unlikely to be an issue in the main docs builds.

PR #1965, opened 2020-09-22
Adds chunk attribute to GO API book

szabosteve:
## Overview

This PR adds the `chunk: 1` attribute to the attribute list of the GO API book in the conf.yml and adds `--chunk 1` to the GO API book alias in `doc_build_aliases.sh`.

These changes are required because of upcoming structural changes in the GO client book.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -428,7 +428,7 @@ contents:
                 current:    7.x
                 branches:   [ master, 7.x ]
                 index:      .doc/index.asciidoc
-                single:     1
+                chunk:      1
                 tags:       Clients/Go
                 subject:    Clients
                 sources:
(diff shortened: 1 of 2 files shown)


szabosteve:
@gtback I'm not sure whether it is okay to replace `single` with `chunk`. I'd like to reorganize the Go Client Docs which is a one-pager now. I'd like to have a TOC front page generated from the index.asciidoc file and add sub-sections to the structure.   Should I do any other infrastructural changes in this repo?

gtback (approved):
This seems OK to me. You probably want to also update the alias to point to the file in the `go-elasticsearch` repository. 

Right now, each chunk is **tiny**. By default (with `--chunk 0` or no `chunk` at all, since 0 is the default when we don't pass `--single`), we split pages at every level 2 heading, which ends up looking like this: 

image

With `--chunk 1` (like in this PR), we actually split at every level 3 heading, which looks like this:

image

So depending on what you want and how much you plan on adding, you may want to remove the `--chunk` attribute (or set it explicitly to 0), rather than set it to 1. But that's up to you.

lcawl (approved):
LGTM too!

PR #1967, opened 2020-09-23
Move the edit links outside of the <h#> elements

### LINKED ISSUE #1966: Move Edit links out of headings
Right now the edit links are located within the heading (for example `<h1>`) elements. Besides being semantically incorrect, this leads to to "Edit" occasionally being considered part of the heading.

gtback:
- Use flexbox rather than absolute positioning.
- Give headers a consistent top and bottom margin.

Fix #1966

### DIFF
--- a/resources/web/style/heading.pcss
+++ b/resources/web/style/heading.pcss
@@ -14,8 +14,8 @@
   }
   
   /* For *JUST* the first heading on the page, shrink the 40px margin a bit */
-  titlepage div {
-    margin-top: -30px;
+  .titlepage div {
+    margin-top: -40px;
   }
 
   h1, h2, h3, h4, h5, h6 {
--- a/resources/web/style/heading.pcss
+++ b/resources/web/style/heading.pcss
@@ -11,10 +11,6 @@
     /* Give all headings plenty of space above them */
     margin-top: 40px;
     margin-bottom: 10px;
-
-    a, h1, h2, h3, h4, h5, h6 {
-      display: flex
-    }
   }
 
   h1, h2, h3, h4, h5, h6 {
@@ -26,7 +22,7 @@
     &:hover a[id][href] {
       display: block;
       position: absolute;
-      bottom: 3px;
+      top: 3px;
       margin: 0;
       padding: 0;
       left: -20px;
--- a/resources/web/style/heading.pcss
+++ b/resources/web/style/heading.pcss
@@ -14,9 +14,9 @@
   }
   
   /* For *JUST* the first heading on the page, shrink the 40px margin a bit */
-  .titlepage div {
+  /* .titlepage div {
     margin-top: -40px;
-  }
+  } */
 
   h1, h2, h3, h4, h5, h6 {
     /* Relative position allows laying out the edit me links inside the header. */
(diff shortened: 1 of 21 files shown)


gtback:
Once the docs-build runs, I'll put some before-and-after screenshots. I don't think there will be much significant visual difference.

resources/web/style/heading.pcss
```diff
@@ -1,4 +1,23 @@
...
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+    /* Give all headings plenty of space above them */
+    margin-top: 40px;
+    margin-bottom: 10px;
+
+    a, h1, h2, h3, h4, h5, h6 {
+      display: flex
+    }
```
JoshMock, line 13: What is the benefit of employing flexbox here? It's within a pretty small scope so I guess introducing accidental flexbox bugs is probably unlikely but I'm just curious. I've learned just enough flexbox to be equal parts dangerous and wary of its ways. 😄

gtback: It's the best way **I know** to have two elements center-aligned vertically, while having the second one pushed all the way to the right. That's not the same as it being the _actual_ best way, though. 😁 

I started trying by trying to set the `.edit_me` icons as `float: right`, but the vertical alignment was a mess, and I was trying to minimize the changes to HTML (I  basically just added the `heading` class to the enclosing `<div>` to make it easier to target with this rule.

JoshMock: Totally fair. `display: flex` is the new `float:` 😄

JoshMock (approved):

gtback:
I still need to do some work on this. The `-40px` doesn't work correctly in all cases (on some pages, it's not needed).

gtback:
I spot-checked a bunch of pages, and they are all looking pretty consistent to me. I'd appreciate if some @elastic/docs writers could take a look at pages in [this preview](https://docs_1967.docs-preview.app.elstc.co/guide/index.html) to make sure you all don't notice anything weird. 

I did notice that the headline on [this page](https://docs_1967.docs-preview.app.elstc.co/guide/en/beats/libbeat/7.9/whats-new.html) isn't getting wrapped correctly. I'll make sure to fix that before merging.

bmorelli25:
I wonder if a top margin or padding should be added? For example, the H3s on [this page](https://docs_1967.docs-preview.app.elstc.co/guide/en/observability/current/ingest-logs.html) seem tight:

---

Screen Shot 2020-11-05 at 11 39 08 AM

---

Screen Shot 2020-11-05 at 11 39 18 AM

---

Screen Shot 2020-11-05 at 11 39 27 AM

lcawl:
There seems to be something strange happening here:
https://docs_1967.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/index.html

image

gtback:
Thanks for these examples, @bmorelli25 and @lcawl. I'll take a closer look. I'm curious why the headings are so inconsistent throughout the site 🤷

gtback:
It looks like the issue is that the `titlepage` div is used in multiple places on the page :-( I'll see if I can make that CSS exception more specific.

gtback:
I've spent too long on this already, and the HTML that gets generated is inconsistent enough that there's not an elegant CSS solution I've found without a ton of edge cases which I'd like to avoid. I'm just going to close this, but leave the branch in case this becomes a priority again.

PR #1968, opened 2020-09-23
Add live branches for client docs

lcawl:
Some very old client versions are appearing high in the search results. For example:

image

This PR adds the "live" option to the client docs in the conf.yaml so that their active branches are prioritized. In a couple of cases there is only one branch, but I didn't think it hurt to add it there either in case more branches are added in the near future.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -404,6 +404,7 @@ contents:
                 prefix:     javascript-api
                 current:    7.x
                 branches:   [ master, 7.x, 6.x, 5.x, 16.x ]
+                live:       [ master, 7.x ]
                 index:      docs/index.asciidoc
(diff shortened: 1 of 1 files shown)


gtback (approved):
Live branches [will be the same as](https://github.com/elastic/docs/blob/bb77e0101cae6bcd7a4485954477d69a9ca3d518/lib/ES/Book.pm#L152) the full list of current branches if not otherwise specified, so in some of these cases it's not needed, but in practice it doesn't hurt and prevents forgetting to add it later.

There are some other changes I need to make related to this, but I can probably make them in a separate PR.

szabosteve (approved):
Thanks for this improvement! LGTM!

PR #1970, opened 2020-09-30
Change ECS chunking level

ebeahan:
Updating the ECS book chunking config from `1` to `2`. This change will enable adding additional usage sub-sections per fieldset in the ECS field reference.

Example:

Screen Shot 2020-09-30 at 1 21 20 PM

The full details of these ECS documentation changes are captured in https://github.com/elastic/ecs/issues/943.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -172,7 +172,7 @@ contents:
             current:    1.6
             branches:   [ master, 1.x, 1.6, 1.5, 1.4, 1.3, 1.2, 1.1, 1.0 ]
             index:      docs/index.asciidoc
-            chunk:      1
+            chunk:      2
             tags:       Elastic Common Schema (ECS)/Reference
             subject:    Elastic Common Schema (ECS)
             sources:
--- a/doc_build_aliases.sh
+++ b/doc_build_aliases.sh
@@ -185,7 +185,7 @@ alias docbldela='$GIT_HOME/docs/build_docs --doc $GIT_HOME/elasticsearch-eland-d
 alias docbldx='$GIT_HOME/docs/build_docs --doc $GIT_HOME/x-pack/docs/en/index.asciidoc --resource=$GIT_HOME/kibana-extra/x-pack-kibana/docs --resource=$GIT_HOME/elasticsearch-extra/x-pack-elasticsearch/docs --chunk 1'
 
 # ECS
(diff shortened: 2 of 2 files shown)


gtback (approved):
This is fine and does exactly what it claims to do 😁 .

I'll note that in a lot of the docs, writers are moving to *longer* pages, to make it easier on developers to CTRL-F to find content on a page. Additionally, the pages that currently hold the bulk of the information, like https://www.elastic.co/guide/en/ecs/current/ecs-as.html, are changing to become stubs like https://docs_1970.docs-preview.app.elstc.co/guide/en/ecs/current/ecs-as.html. At least we don't need to add redirects, but I think it's still a suboptimal user experience.

Finally, all of the "Field Reuse" pages have URLs like https://docs_1970.docs-preview.app.elstc.co/guide/en/ecs/current/_field_reuse_6.html. It might be good to update the IDs in the source files before merging this in. That will break existing anchors like `https://www.elastic.co/guide/en/ecs/current/ecs-file.html#_field_reuse_6`, but those will be changing anyway as this content gets moved onto a new page.

ebeahan:

ebeahan:
`master` and `1.x` are now correct in the docs preview, but I have some additional backporting for the other versions (including `current`/`1.6`) to perform.

webmat:
Quick question @gtback, is it possible for us to set the chunking per branch of the ECS docs?

Setting chunking to 2 should only be relevant to our next release, 1.7. Since the pages that require an adjustment in adding `[discrete]` are generated pages, we need to go modify different versions of the script over the years to make all past versions look right, at this new setting.

If instead we can easily force chunking to `1` in branches 1.0, 1.1, 1.2, 1.3, 1.4, 1.5 and 1.6 from here, I think that would be easier.

gtback:
Hey @webmat. It's not currently possible to change chunking level per-branch. If it's important, I can look into what level of effort would be needed to add this ability.

An alternative would be to create a new book and move the current one to the "legacy" section, but then you lose some of the ability to switch between the "same" page in different versions, and we might have to choose a different URL. I'd say the cons of that approach probably outweigh the pros, but wanted to at least suggest it.

webmat:
Alright, thanks @gtback! We'll handle it on our side, then.

ebeahan:

PR #1972, opened 2020-10-01
Add 7.10 branches to conf.yaml

lcawl:
This PR adds new 7.10 branches to the appropriate books.

NOTE: These build checks will fail until all the branches exist.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -975,7 +975,7 @@ contents:
                 prefix:     get-started
                 index:      docs/guide/index.asciidoc
                 current:    *stackcurrent
-                branches:   [ master, 7.9, 7.8, 7.7, 7.6, 7.5, 7.4, 7.3, 7.2, 7.1, 7.0, 6.8, 6.7, 6.6, 6.5, 6.4, 6.3, 6.2, 6.1, 6.0 ]
+                branches:   [ master, 7.10, 7.9, 7.8, 7.7, 7.6, 7.5, 7.4, 7.3, 7.2, 7.1, 7.0, 6.8, 6.7, 6.6, 6.5, 6.4, 6.3, 6.2, 6.1, 6.0 ]
                 live:       *stacklive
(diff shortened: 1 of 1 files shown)


conf.yaml
```diff
@@ -1768,8 +1768,8 @@ contents:
           - title:      Ingest Management Guide
             prefix:     en/ingest-management
             current:    *stackcurrent
-            branches:   [ master, 7.x, 7.9, 7.8 ]
-            live:       [ master, 7.x, 7.9, 7.8 ]
+            branches:   [ master, 7.x, 7.10, 7.9, 7.8 ]
+            live:       [ master, 7.x, 7.10, 7.9, 7.8 ]
```
gtback, line 1772: Should we take 7.8 out of this list? Or can we change it to `*stacklive`?

gtback (commented):
I also noticed we didn't add 7.10 to the Enterprise Search or some of the APM books. Was that intentional? Did you just add 7.10 whenever there was a 7.x?

conf.yaml
lcawl: I think this is a question for @dedemorton.  Can happen separately from this PR (before or after) as far as I'm concerned.

lcawl:
> Did you just add 7.10 whenever there was a 7.x?

Yup, if there are other exceptional cases that don't have a 7.x, we can follow up after this PR if necessary. If they don't have a 7.x there must not be links to those books from other books in the library.

conf.yaml
dedemorton: @lcawl I think I did that before because I didn't have all the branches in the list. You should be able to use *stacklive now.

lcawl:

lcawl:

lcawl:

lcawl:

jmikell821 (approved):

PR #1973, opened 2020-10-05
[DOCS] Adds chunking to Py, Rust, and eland client books

szabosteve:
## Overview

This PR adds the `chunk: 1` attribute to the attribute list of the Python, Rust, and eland client books in the conf.yml and adds `--chunk 1` to the respective book aliases in `doc_build_aliases.sh`.

These changes are required because of upcoming structural changes in the above-mentioned client books. This PR covers the last bits of infrastructural changes required to the Clients docs restructuring.

### Note

**Merge only the following PRs are merged:**

* py: https://github.com/elastic/elasticsearch-py/pull/1393
* eland: https://github.com/elastic/elasticsearch-eland-docs/pull/9
* rust: https://github.com/elastic/elasticsearch-rs/pull/150

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -488,7 +488,7 @@ contents:
                 branches:   [ master, 7.x ]
                 live:       [ master, 7.x ]
                 index:      docs/guide/index.asciidoc
-                single:     1
+                chunk:     1
                 tags:       Clients/Python
                 subject:    Clients
                 sources:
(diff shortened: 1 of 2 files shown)


gtback (approved):
LGTM!

We realized in #1970 that the chunking level is used by all versions of a book. In this case it's OK, since most of the client books only build one version, but wanted to explicitly call that out in case it matters.

PR #1981, opened 2020-10-16
Update Observability sources

bmorelli25:
How do sources work for documentation books? We have files in `apm-server`'s `doc` folder that the Observability guide needs to access. Is the fix as simple as removing the `/guide` from the source?

For the failures in https://github.com/elastic/observability-docs/pull/188:

```
08:57:19 INFO:build_docs:
08:57:19 INFO:build_docs:asciidoctor: ERROR: instrument-apps.asciidoc: line 22: include file not found: /tmp/docsbuild/qaSsMmfBdn/apm-server/docs/tab-widgets/install-agents-widget.asciidoc
08:57:19 INFO:build_docs:asciidoctor: ERROR: instrument-apps.asciidoc: line 36: include file not found: /tmp/docsbuild/qaSsMmfBdn/apm-server/docs/tab-widgets/configure-agent-widget.asciidoc
08:57:19 INFO:build_docs:asciidoctor: ERROR: instrument-apps.asciidoc: line 43: include file not found: /tmp/docsbuild/qaSsMmfBdn/apm-server/docs/tab-widgets/configure-server-widget.asciidoc
08:57:19 INFO:build_docs:
```

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -956,7 +956,7 @@ contents:
                 path:   docs/en
               -
                 repo:   apm-server
-                path:   docs/guide
+                path:   docs
                 exclude_branches:   [ 6.8, 6.7, 6.6, 6.5, 6.4, 6.3, 6.2, 6.1, 6.0, 5.6, 5.5, 5.4, 5.3, 5.2, 5.1, 5.0 ]
               -
                 repo:   beats


bmorelli25:
no_idea

gtback:
As I understand it, the `path` under each `sources` entry serves two purposes:

- If anything in that [path has changed](https://github.com/elastic/docs/blob/006f26a186820af2f61d30a153125a73ccd125be/lib/ES/Repo.pm#L97-L102), that's a sign that the book needs to be rebuilt.
- We do a [sparse checkout in git](https://github.com/elastic/docs/blob/006f26a186820af2f61d30a153125a73ccd125be/lib/ES/Repo.pm#L224-L227) to only check out those files (which helps in larger repos). That is what's causing those errors.

In this case, it seems like removing `guide` makes the most sense. 

@nik9000 am I explaining that correctly?

gtback (approved):

gtback:
I was thinking I should document this somewhere, and then I [realized that I already did](https://github.com/elastic/docs/blob/master/schema.yaml#L115-L118) 😂 

Part of why I made that file was to document things like this, but I don't think I really told anyone else about it.

bmorelli25:
That's very helpful! It also explains why the path to that content is correct, but the build claims that it doesn't exist -- it never checked it out.

bmorelli25:
Oh cool! I've never seen that schema file

lcawl:
I'm running the full build locally now to test, but it's worth noting that per Greg's comment, you technically could reduce the amount of churn by adding docs/tab-widgets as a source, instead of adding all folders in docs.  Depends how many of these other folders are likely to be needed in the future and how much they change, I guess.

bmorelli25:
Interesting! At some point in the future we'll want to add the APM Server reference content (also in `/docs`), but that might not happen before the new build system. Are you suggesting something like this?

```
 -
                repo:   apm-server
                path:   docs/guide
 -
                repo:   apm-server
                path:   docs/tab-widgets
```

PR #1987, opened 2020-10-22
Get Fleet User Guide building

dedemorton:
Closes https://github.com/elastic/observability-docs/issues/128

TODO after merging:

- [ ] Move the Ingest Management Guide from Beats section to legacy docs section.
- [ ] Create new path attribute for the Fleet User Guide and update all docs in master, 7.x, and 7.10 (across repos) to use the new attribute.
- [ ] Set up redirects for master, current, and 7.10.
- [ ] Stop building master, 7.x, and 7.10 branches of the Ingest Management Guide

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -1305,7 +1305,7 @@ contents:
                 repo:   docs
                 path:   shared/attributes.asciidoc
 
-    -   title:      "Fleet: Install and manage Elastic Agents"
+    -   title:      "Fleet: Install and Manage Elastic Agents"
         sections:
           - title:      Fleet User Guide
             prefix:     en/fleet
--- a/conf.yaml
+++ b/conf.yaml
@@ -1305,6 +1305,31 @@ contents:
                 repo:   docs
                 path:   shared/attributes.asciidoc
 
+    -   title:      "Fleet: Install and manage Elastic Agents"
+        sections:
+          - title:      Fleet User Guide
+            prefix:     en/fleet
+            current:    master
+            branches:   [ master, 7.x, 7.10 ]
+            live:       *stacklive
+            index:      docs/en/ingest-management/index.asciidoc
+            chunk:      1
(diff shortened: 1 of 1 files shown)


dedemorton:
@andresrc Given today's discussion at the EAH, I wonder if it might make sense to put the Fleet User Guide under the Observability container rather than creating a separate container as I propose in this PR. WDYT?

gtback:
Generally speaking, I'd vote against making a new section just for one book, unless it really doesn't fit into any other sections. We can always split off a new section later if there are more books that fit naturally with this one. 

I'm not sure what makes the most sense here, though. Fleet is *primarily* for Observability integrations, but also used for Security, correct?

dedemorton:
Me too, but there is not a great place in the current doc landing page for this content. After Tanya's talk at the EAH meeting this week, it seemed like Observability might be an OK place, but yah...it's relevant for security, too.

For me, the main thing is getting it out of the Beats section where I think it's hidden.

conf.yaml
```diff
@@ -1788,6 +1788,31 @@ contents:
                 repo:   docs
                 path:   shared/attributes.asciidoc
 
+    -   title:      "Fleet: Manage Integrations and Elastic Agents"
```
dedemorton, line 1791:
```suggestion
    -   title:      "Fleet: Install and manage Elastic Agents"
```

conf.yaml
```diff
@@ -1305,6 +1305,31 @@ contents:
                 repo:   docs
                 path:   shared/attributes.asciidoc
 
+    -   title:      "Fleet: Install and manage Elastic Agents"
```
dedemorton, line 1308:
```suggestion
    -   title:      "Fleet: Install and Manage Elastic Agents"
```

bmorelli25 (approved):
LGTM https://docs_1987.docs-preview.app.elstc.co/guide/index.html

PR #1995, opened 2020-11-03
Add Cloud Terraform provider docs to builds

nrichers:
This PR adds:

* A new `tpec` entry in conf.yaml for the Elastic Cloud Terraform provider book
* A new build alias for building the docs locally (try `docbldtpec --open` once you have the PR mentioned below locally)

Depends on the new docs content in https://github.com/elastic/terraform-provider-ec/pull/163. 

In terms of placement, this is where the new docs will appear:

image

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -824,7 +824,7 @@ contents:
             subject:    TPEC
             current:    master
             branches:   [ master ]
-            index:      docs/index.asciidoc
+            index:      docs-elastic/index.asciidoc
             chunk:      1
             sources:
               -
--- a/conf.yaml
+++ b/conf.yaml
@@ -818,6 +818,18 @@ contents:
               -
                 repo:   ecctl
                 path:   docs/
+          - title:      Elastic Cloud Terraform Provider
+            prefix:     en/tpec
+            tags:       CloudTerraform/Reference
+            subject:    TPEC
+            current:    master
+            branches:   [ master ]
+            index:      docs/index.asciidoc
+            chunk:      1
+            sources:
+              -
+                repo:   terraform-provider-ec
+                path:   docs/
 
     -   title:      "Kibana: Explore, Visualize, and Share"
         sections:
--- a/conf.yaml
+++ b/conf.yaml
@@ -829,7 +829,7 @@ contents:
             sources:
               -
                 repo:   terraform-provider-ec
-                path:   docs/
+                path:   docs-elastic/
 
     -   title:      "Kibana: Explore, Visualize, and Share"
         sections:
--- a/conf.yaml
+++ b/conf.yaml
@@ -823,7 +823,7 @@ contents:
             prefix:     en/tpec
             tags:       CloudTerraform/Reference
             subject:    TPEC
-            current:    master
+            current:    v0.1.0-beta
             branches:   [ master, v0.1.0-beta ]
             index:      docs-elastic/index.asciidoc
             single:     1
--- a/conf.yaml
+++ b/conf.yaml
@@ -825,6 +825,7 @@ contents:
             current:    master
             branches:   [ master ]
             index:      docs-elastic/index.asciidoc
+            single:     1
             chunk:      1
             sources:
               -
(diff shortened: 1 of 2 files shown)


conf.yaml
```diff
@@ -818,6 +818,18 @@ contents:
...
+            tags:       CloudTerraform/Reference
+            subject:    TPEC
+            current:    master
+            branches:   [ master ]
+            index:      docs/index.asciidoc
+            chunk:      1
+            sources:
+              -
+                repo:   terraform-provider-ec
+                path:   docs/
```
karencfv, line 832: Depends on https://github.com/elastic/terraform-provider-ec/pull/163#pullrequestreview-522134387

doc_build_aliases.sh
```diff
@@ -96,6 +96,9 @@ alias docbldecctl='$GIT_HOME/docs/build_docs --doc $GIT_HOME/ecctl/docs/index.as
 # Cloud - Elastic Cloud for K8s
 alias docbldk8s='$GIT_HOME/docs/build_docs --doc $GIT_HOME/cloud-on-k8s/docs/index.asciidoc --chunk 1'
 
+# Cloud - Terraform provider
+alias docbldtpec='$GIT_HOME/docs/build_docs --doc $GIT_HOME/terraform-provider-ec/docs/index.asciidoc --chunk 1'
```
karencfv, line 100: This link also depends on https://github.com/elastic/terraform-provider-ec/pull/163#pullrequestreview-522134387 :)

conf.yaml
```diff
@@ -818,6 +818,18 @@ contents:
...
+            prefix:     en/tpec
+            tags:       CloudTerraform/Reference
+            subject:    TPEC
+            current:    master
+            branches:   [ master ]
+            index:      docs-elastic/index.asciidoc
+            chunk:      1
+            sources:
+              -
+                repo:   terraform-provider-ec
```
gtback, line 831: I think you need to add this at the top under `repos` as well. This is what's causing the build failure.

gtback (commented):
I'm not a huge fan of the acronym `tpec` (It took me too long to realize what it meant). For similarity with the other cloud books, maybe `cloud-terraform-provider` or `cloud-tf-provider` or `cloud-terraform`?

conf.yaml
nrichers: Yup, I updated, content now lives in https://github.com/elastic/terraform-provider-ec/pull/163/files#diff-549ff6e8034edcde5d1dbd30d3546302ffccfad46d8f0d0845133f39e4521cd5.

doc_build_aliases.sh
nrichers: Same here, I updated the build alias to use the new `docs-elastic` location.

karencfv (commented):
Thanks for addressing my comments :) Should we hold off on merging until we release the provider though?

nrichers:
> Thanks for addressing my comments :) Should we hold off on merging until we release the provider though?

@karencfv Yes, we should not merge this until we're ready. 

What version is being released around November 15? If it's a beta, we might want to cut a different branch for that. As is, `master` will get a version disclaimer at the top that says this content is for a future release.

karencfv:
Yes it's a beta, so we'll do that! Just a heads up that we'll probably release a bit later do to marketing things, but I don't have an exact date yet. 

cc @Kushmaro

nrichers:
> Yes it's a beta, so we'll do that!

You'll also need a page_header.html file that needs to be added to the branch and you will need to add the beta branch to this PR. Happy to help further when you're ready!

karencfv:
Thanks @nrichers! We'll most likely need the help, so we'll ping you when the time comes :)

cc @marclop

conf.yaml
```diff
@@ -818,6 +818,18 @@ contents:
               -
                 repo:   ecctl
                 path:   docs/
+          - title:      Elastic Cloud Terraform Provider
+            prefix:     en/tpec
+            tags:       CloudTerraform/Reference
+            subject:    TPEC
+            current:    master
+            branches:   [ master ]
```
marclop, line 826: we should add `v0.1.0-beta` as a branch too. It'll be created shortly

karencfv (approved):

conf.yaml
nrichers: Yup, good catch! I added the missing terraform-provider-ec repo at the top. Hopefully this will fix the failures.

conf.yaml
nrichers: OK, I've added `v0.1.0-beta` to the build after creating the branch. That should do it.

PR #1998, opened 2020-11-05
Add Observability repo source to stack docs

bmorelli25:
## Summary

This PR adds observability-docs as a source to the stack-docs repo. This will allow Observability "What's new" content to be included in the installation and upgrade guide.

## Related 

https://github.com/elastic/observability-docs/pull/216

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -91,7 +91,7 @@ contents:
               -
                 repo:   observability-docs
                 path:   docs/en/observability
-                exclude_branches:   [ 7.9 ]
+                exclude_branches:   [ 7.9, 7.8, 7.7, 7.6, 7.5, 7.4, 7.3, 7.2, 7.1, 7.0, 6.8, 6.7, 6.6, 6.5, 6.4, 6.3, 6.2, 6.1, 6.0, 5.6, 5.5, 5.4, 5.3, 5.2, 5.1, 5.0 ]
               -
                 repo:   docs
                 path:   shared/versions/stack/{version}.asciidoc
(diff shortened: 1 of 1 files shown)


conf.yaml
```diff
@@ -88,6 +88,10 @@ contents:
                 repo:   logstash
                 path:   docs/
                 exclude_branches:   [ 6.8, 6.7, 6.6, 6.5, 6.4, 6.3, 6.2, 6.1, 6.0, 5.6, 5.5, 5.4, 5.3, 5.2, 5.1, 5.0 ]
+              -
+                repo:   observability-docs
+                path:   docs/en/observability
+                exclude_branches:   [ 7.9 ]
```
gtback, line 94: If I understand this correctly, this should be a list of the branches of the Installation and Upgrade guide that shouldn't use this as a source. @nik9000, can you confirm that?

bmorelli25: That's how I understand it as well. It only includes `7.9` because the obs-docs guide is only built in `master`, `7.x`, `7.10`, and `7.9`

bmorelli25: I see what you're saying. You're right, it should be based on IU guide, not Obs. I'll fix!

bmorelli25:
```suggestion
                exclude_branches:   [ 7.9, 7.8, 7.7, 7.6, 7.5, 7.4, 7.3, 7.2, 7.1, 7.0, 6.8, 6.7, 6.6, 6.5, 6.4, 6.3, 6.2, 6.1, 6.0, 5.6, 5.5, 5.4, 5.3, 5.2, 5.1, 5.0 ]
```

gtback (approved):
LGTM!

PR #1999, opened 2020-11-07
Add observability docs resource to the doc build aliases

dedemorton:
The following resource is needed to support changes to the tagged regions in the highlights section:

`--resource=$GIT_HOME/observability-docs/docs/en/observability`


lcawl (approved):
LGTM, thanks!

PR #2003, opened 2020-11-12
Add attribute and new alias for Fleet User Guide

dedemorton:
Adds a new attribute for resolving the path to the Fleet User Guide. Also adds the build alias.

Note that I plan to rename the `ingest-management` docs folder to `fleet`, but leaving it as-is for now to get these changes in place first.


dedemorton:
Closing because we are going to do this differently.

PR #2013, opened 2020-11-30
ecctl: Remove beta suffixed branches from listing

marclop:
Removes the `beta` suffixed branches from the doc building.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -818,7 +818,7 @@ contents:
             tags:       CloudControl/Reference
             subject:    ECCTL
             current:    1.0
-            branches:   [ master, 1.0, 1.0.0-beta3, 1.0.0-beta2 ]
+            branches:   [ master, 1.0 ]
             index:      docs/index.asciidoc
             chunk:      1
             sources:


gtback:
It generally doesn't hurt to leave the existing branches, but we can de-emphasize them by creating a `live` key with just the two branches you've kept in this PR. If we really want to delete these branches, I'd want to check the analytics to make sure they aren't getting much traffic, and probably set up a redirect (I can help with this if that's what we decide to do).

nrichers (approved):
LGTM. We'd talked about removing the old beta branches sometime after GA to make sure people have a chance to move to over to a current version. It feels like the time is right. @gtback if you want to set up a redirect, that would certainly help any stragglers still on an old beta version.

gtback:
I was hoping that 1.0 would be a superset of the pages in 1.0.0-beta2 and 1.0.0-beta3, but it doesn't seem like that's the case. Do you think it would be OK to just redirect everything under `https://www.elastic.co/guide/en/ecctl/1.0.0-beta3/*` and `https://www.elastic.co/guide/en/ecctl/1.0.0-beta3/*` to https://www.elastic.co/guide/en/ecctl/current/index.html? Otherwise I can go through the work of redirecting each page to the equivalent page and for any that don't exist just redirecting those to the index page.

nrichers:
> Do you think it would be OK to just redirect everything under `https://www.elastic.co/guide/en/ecctl/1.0.0-beta2/*` [_sic_] and `https://www.elastic.co/guide/en/ecctl/1.0.0-beta3/*` to https://www.elastic.co/guide/en/ecctl/current/index.html?

I think that would be totally OK! The redirect is just to be nice to our users; if they end up on the front page of the current docs instead of getting a 404, that would do it. Thank you so much!

gtback:
@nrichers, since 1.1 is coming out soon (#2016), I think I'm going to set 1.0.0-beta2 and 1.0.0-beta3 to instead redirect to `1.0` rather than `current`. The 1.0 docs will have a banner pointing to the new current version anyway.

Does that sound OK?

nrichers:
> Does that sound OK?

Yup! Thank you so much for adding these redirects, @gtback.

gtback:
I think the conflict is due to #2016 getting merged. It should be straightforward to resolve, but let me know if I can help with anything.

nrichers:
Conflict should be fixed now. Let's see if we can get this thing to turn green.

gtback:
The redirects are live, so I think this is good to merge whenever!

PR #2018, opened 2020-12-08
Add ECS .NET & Ruby documentation

bmorelli25:
This PR adds ECS Logging .NET documentation to our build system. It sets `master` as the `current` branch and sets up the required include parameters for including content from `ecs-logging`.

Blocked by:
- [x] https://github.com/elastic/ecs-dotnet/pull/120
- [x] elastic/ecs-logging#40
- [x] https://github.com/elastic/docs/pull/2017
- [x] Update ecs-dotnet docs to use shared content elastic/ecs-dotnet#122

PR #2017, opened 2020-12-08
docs: add ecs-logging

bmorelli25:
This PR adds the `ecs-logging` repository as a source repo for the `ecs-logging-java` documentation. It also maps the `master` branch of `ecs-logging` to both the `1.x` and `0.x` branches of `ecs-logging-java`.

Blocked by:
- [x] https://github.com/elastic/ecs-logging/pull/40
~https://github.com/elastic/ecs-logging-java/pull/117.~

PR #2019, opened 2020-12-09
Add ecs-logging and ecs-dotnet repos

gtback:
These were added to the cached worker images used for CI, and will soon be needed by docs builds. I have a periodic job that keeps these two lists in sync, so I'm adding them in here, even though they aren't used by any published docs yet.


bmorelli25 (approved):
FYI these will be added in #2017 and #2018

gtback:
Thanks @bmorelli25 

> FYI these will be added in #2017 and #2018

Ah, I hadn't looked at those yet 🤦 . In any event, merging this in and them merging back to those branches will prevent the conflicts.

PR #2020, opened 2020-12-10
Add missing resource to filebeat build alias

dedemorton:
Fixes doc build alias for Filebeat so that local build does not fail.

@gtback The Beats aliases were defined before we made the file structure of the books more complex. Should we try to sync up the aliases with all potential resource paths defined in the conf.yaml, or just deal with them when they break? I'm inclined to wait until they break.

### DIFF
--- a/doc_build_aliases.sh
+++ b/doc_build_aliases.sh
@@ -106,7 +106,7 @@ alias docbldbdg='$GIT_HOME/docs/build_docs --respect_edit_url_overrides --doc $G
 
 alias docbldpb='$GIT_HOME/docs/build_docs --respect_edit_url_overrides --doc $GIT_HOME/beats/packetbeat/docs/index.asciidoc --chunk 1'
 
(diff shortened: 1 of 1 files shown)


dedemorton:
@EamonnTP Don't forget to run `source ~/.bash_profile` (assuming you use bash).

gtback:
> @gtback The Beats aliases were defined before we made the file structure of the books more complex. Should we try to sync up the aliases with all potential resource paths defined in the conf.yaml, or just deal with them when they break? I'm inclined to wait until they break.

I'm tempted to wait until they break, but I'm fine with whatever works better for you. The aliases don't have the same "exclude_branches" logic as the entries in conf.yaml, so its possible things might break if you add resources that aren't used in the branch you're trying to build. You can also add any extra arguments to a build, right after the alias:
```
docbldfb --resource=<SOME OTHER RESOURCE>`
```
(I do this with `--open` all the time)

gtback (approved):

PR #2042, opened 2021-01-20
Change current to MS 51

nrichers:
This PR updates our docs to use MS 51 as `current` for ESS and the Heroku docs built partially off the same source.


gtback (approved):
🍔

beiske (approved):

PR #2044, opened 2021-01-26
Add is-current-version attribute

bmorelli25:
## Summary

This PR proposes the addition of a new shared attribute, `is-current-version`. `is-current-version` will be set to `true` in the currently released stack version, and `false` in all other versions.

## Release

At each minor and major release, a two-line update will be required. The previously current version will need to be set to `false`, and the newly current version will need to be set to `true`.

For example, in the 7.11 release, `shared/versions/stack/7.10.asciidoc` will set `:is-current-version:` to `false` and `shared/versions/stack/7.11.asciidoc:` will set `:is-current-version:` to `true`.

If Greg and Lisa are cool with this, I'll update the release documentation as well.

## Related

Required for https://github.com/elastic/observability-docs/pull/350.

### DIFF
--- a/shared/versions/stack/7.10.asciidoc
+++ b/shared/versions/stack/7.10.asciidoc
@@ -17,8 +17,12 @@ bare_version never includes -alpha or -beta
 //////////
 release-state can be: released | prerelease | unreleased
 //////////
-
 :release-state:          released
+
+//////////
+is-current-version can be: true | false
+//////////
+:is-current-version:    true
 
 ////
 APM Agent versions
--- a/shared/versions/stack/7.11.asciidoc
+++ b/shared/versions/stack/7.11.asciidoc
@@ -17,8 +17,12 @@ bare_version never includes -alpha or -beta
 //////////
 release-state can be: released | prerelease | unreleased
 //////////
-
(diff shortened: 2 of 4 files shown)


dedemorton (approved):
LGTM.

I think the value of putting this in a central location outweighs the minor inconvenience of having an extra setting to change at release time. Plus if someone forgets, that's OK because it just means that the video will display in older versions (not a deal breaker).

It's been awhile since I've bumped doc versions, but beats has a script they use. If we don't have that for the docs build, might be time to add it (general comment, not intended for this PR).

gtback (approved):
I'm 👍🏻 with this approach. It's not the most elegant, but practicality beats complexity 😁

lcawl:
This reminds me of the work @nik9000 did to create a {source_branch} attribute seamlessly for us to use via https://github.com/elastic/docs/pull/1147.  Since "current" is currently defined in the conf.yaml, it would have been nice to have the build therefore set an "isCurrent" attribute (or whatever) from that info too. But I don't think it's worth investing time in a fancy solution like that in the current system.

lcawl (approved):
Added comments and link for future vision, but LGTM in current world

PR #2045, opened 2021-01-26
Add buildSrc resource for ES Ref to conf.yaml

debadair:
To pull the version.properties for Elasticsearch dependencies into an appendix, we need to add the directory as a resource.


gtback (approved):
LGTM!

PR #2058, opened 2021-02-11
Change current to MS 52 for ESS

alaudazzi:
This PR updates our docs to use MS 52 as `current` for ESS and the Heroku docs.


szabosteve (approved):
LGTM!

gtback:
(After the fact) Looks good! 👍🏻

PR #2060, opened 2021-02-11
Update attributes.asciidoc

gchaps:
This PR updates the searchprofiler attribute to use Search Profiler (instead of Query Profiler).


lcawl:
Depending on how far back that change goes, you might also want to update the attribute here: https://github.com/elastic/docs/blob/master/shared/attributes62.asciidoc

lcawl (approved):
One comment, otherwise LGTM

PR #2065, opened 2021-02-16
Capitalize subject for ML Book

gtback:
This value is used as the `DC.subject` meta value for the ML book. The search facets on www.elastic.co use the capitalized version of the name, so we should match that in order for results to appear.

cc: @laurent-qa @brianjolly


lcawl (approved):
LGTM

PR #2069, opened 2021-02-18
Add build alias for logstash plugin reference

gtback:
It's actually a function, though ;-)


karenzone (approved):
Helpful! Thank you <3
LGTM

PR #2079, opened 2021-03-15
Add zshrc to doc_build_aliases instructions

lcawl:
On newer machines, it seems the default is zsh not bash, so the instructions in the doc_build_aliases can be misleading as-is. This PR adds a reference to .zshrc.


debadair (approved):
Thanks!

gtback:
Thanks @lcawl !

Issue #1805, opened 2020-04-10
Kibana link checking is broken


bmorelli25:
The Kibana link checker doesn't seem to be catching broken links.

https://github.com/elastic/docs/blob/0cee4e4715554515b2e297fe1113708ac03c9cda/build_docs.pl#L334-L399

Here's an example of a broken link: https://github.com/elastic/kibana/pull/63295.

gtback:
I'm only starting to understand how Kibana link checking works. @nik9000 gave me an overview this morning. It might take me a little while to wrap my head around it all.

gtback:
I figured out why this was broken (at least for elastic/kibana#63295):

The only links that are checked are those that are located in the root Doc Links file (which has moved and been renamed between versions of Kibana)

If you do something like this in a different file, and then build URLs in those files, those URLs won't be checked. 

```
import { ELASTIC_WEBSITE_URL, DOC_LINK_VERSION } from 'ui/documentation_links';
```

This is what @lcawl is trying to fix in elastic/kibana#88107.

gtback:
I'm tempted to close this, because I don't think looking through all of the Kibana code for links is feasible, but I'd like to track down why elastic/kibana#87721 *wasn't* failing as well.

gtback:
I believe elastic/kibana#94274 (which prevents the `doc_links_service.ts` from being checked) is the ultimate root cause of this (along with elastic/kibana#88107: some of the links aren't in that file so aren't being checked). 

Order of operations for resolving this:

- [ ] Fix #2081.
- [ ] Ensure elastic/kibana#94274 passes, then merge it in.
- [ ] Confirm elastic/kibana#87721 still fails. I merged in some of the changes from elastic/kibana#87721, so it's already failing, but merging in the latest should reduce the diff and it should still fail.
- [ ] Fix the link in elastic/kibana#87721 to confirm it passes.
- [ ] Close this issue

gtback:
elastic/kibana#87721 is passing! 🙌🏻 So I'm confident we have this fixed. The actual link will get fixed once that's merged, but that's not a blocker here.

PR #2085, opened 2021-03-16
Improve link extraction for Kibana link checking

### LINKED ISSUE #2081: Skip checking Kibana links that are on www.elastic.co but not docs links
The Kibana link checking process looks for URLs containing `${ELASTIC_WEBSITE_URL}` and tries to check those links. For pages outside the docs (blog pages, marketing pages), these will result in false-positives for the link-checker, which is preventing elastic/kibana#94274 from passing. That PR should be the ultimate fix for #1805.

gtback:
- Ignore links to web content that's outside the docs
- Ensure Elasticsearch Reference content is checked

Fixes #2081

Related: https://github.com/elastic/kibana/pull/94274, #1805

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -356,12 +356,21 @@ sub check_kibana_links {
     my $extractor = sub {
         my $contents = shift;
         return sub {
-            while ( $contents =~ m!`(\$\{(?:baseUrl|ELASTIC_.+)\}[^`]+)`!g ) {
+            while ( $contents =~ m!`(\$\{(?:baseUrl|ELASTIC.+)\}[^`]+)`!g ) {
                 my $path = $1;
                 $path =~ s/\$\{(?:DOC_LINK_VERSION|urlVersion)\}/$branch/;
-                $path
-                    =~ s!\$\{ELASTIC_DOCS\}!en/elasticsearch/reference/$branch/!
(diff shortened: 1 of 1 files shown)


gtback:
I don't trust my Perl skills, so there's probably a better way to do these regular expressions. I tested this locally with a copy of `doc_links_service.ts` from https://github.com/elastic/kibana/pull/94274, and it fixed the three broken `${ELASTIC_WEBSITE_URL}` links, and only added one broken Elasticsearch link `en/elasticsearch/reference/master/search-aggregations.html#_values_source`, which should probably be fixed in the Kibana PR.

nik9000 (approved):
LGTM. I'd always hoped repos could standardize on this sort of thing but I think that's not really a thing we can expect a ton of at this point.

I know that there are tests for this sort of thing off in ruby land. It'd probably be worth adding one just in case.

tylerjl (approved):
Oh my, perl is very much out of my wheelhouse. But I have implicit confidence in @nik9000's confidence `:recursive-approval:`

nik9000:
It's not the perl I'm confident in. It's that Greg tested it. It looks
idiomatic enough for me though.

olksdr (approved):
:shipit:

PR #2097, opened 2021-03-31
Change chunk level for Fleet User Guide

dedemorton:
DO NOT MERGE UNTIL ALL VERSIONS OF THE FLEET DOCS HAVE BEEN MODIFIED TO SUPPORT THIS CHANGE

This change is required in the current doc system to accommodate the large number of policy settings that will be documented in the agent docs. 

I know this goes contrary to the new design, but until we have the docs in the new docs system, having a flat structure is too overwhelming.

Incidentally, when we merge changes like this, we should let contributors know so they can pull down the latest aliases. We should also tell them when we add new build dependencies.


bmorelli25 (approved):
👍

dedemorton:

PR #1918, opened 2020-07-27
Use the "version" rather than "branch" in file paths

gtback:
There's an existing method to give each "branch" a "title". The title corresponds to the "name" of the version, in cases where it doesn't match the name of the branch itself, and the "title" already appears in the version dropdown. This change extends that to use the version name ("title") in the URL.

Future work will better distinguish between "branches" and "versions" in cases where they aren't the same.

---

To use this, put a `{ branch-name: version-name }` into the `branches` list of a book in conf.yaml. All other places where you would put a branch (like `live` and `current`) should continue to use the branch name (not the version name), but I plan to make further enhancements to make this more straighforward.

---

Related to https://github.com/elastic/docs/issues/1850

PR #2102, opened 2021-04-12
Give the single-version cloud books a version name of "current"

gtback:
This should prevent the milestone name from being used as the
"DC.identifier" used for search facets

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -633,7 +633,7 @@ contents:
             tags:       Cloud/Reference
             subject:    Elastic Cloud
             current:    ms-54
-            branches:   [ ms-54 ]
+            branches:   [ {'ms-54': current} ]
             index:      docs/saas/index.asciidoc
             chunk:      1
             private:    1
@@ -685,7 +685,7 @@ contents:
             tags:       Cloud-Heroku/Reference
             subject:    Elasticsearch Add-On for Heroku
             current:    ms-54
-            branches:   [ ms-54 ]
+            branches:   [ {'ms-54': current} ]
             index:      docs/heroku/index.asciidoc
             chunk:      1
             noindex:    1


gtback:
This might break everything, if it tries to "copy" the files from "current" (the new version name in this PR) to "current" (the hard-coded name for the current version). If so, we'll need a different name, for what we want to show up under "Product Version":

image

kilfoyle (approved):
LGTM. 🚀 
I'm not sure what the "DC.identifier" is but if this can pass the CI checks it works for me. Thanks for fixing this @gtback!

alaudazzi (approved):
LGTM. Thanks Greg!

gtback:
> I'm not sure what the "DC.identifier" is..

This is page metadata that is used to pre-populate the search facets when you click on the Search:

image

Untitled

```
<meta name="DC.subject" content="Elastic Cloud"/>
<meta name="DC.identifier" content="ms-54"/>
```

The value of the `DC.identifier` meta tag is the book's "version". The `{'ms-54': current}` syntax in `branches: [ {'ms-54': current} ]` was added om #1918 and means "create a version called 'current' from the 'ms-54' branch". If we still want to call this version "current", I can make some tweaks to the build to allow this, but otherwise we just need to find another work, because of the build error:

`
15:57:02 INFO:build_docs:Couldn't copy </tmp/docsbuild/target_repo/html/en/cloud/current> to </tmp/docsbuild/target_repo/html/en/cloud/current>: No such file or directory at lib/ES/Book.pm line 413.
`

kilfoyle:
Nice. Thanks for explaining, @gtback!

nrichers (approved):
LGTM with the proviso that this not break for the new docs system, but based on our convo today this should be peachy. 🍑 🍑 🍑

gtback:

PR #2103, opened 2021-04-12
Change Cloud Terraform provider version

nrichers:
This PR updates the version for our Terraform provider docs to be in step with https://github.com/elastic/terraform-provider-ec/blob/master/CHANGELOG.md. Need to do a full rebuild after the PR gets merged.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -850,8 +850,8 @@ contents:
             prefix:     en/tpec
             tags:       CloudTerraform/Reference
             subject:    TPEC
-            current:    v0.1.0-beta
-            branches:   [ master, v0.1.0-beta ]
+            current:    0.1
+            branches:   [ master, 0.1 ]
             index:      docs-elastic/index.asciidoc
             single:     1
             chunk:      1


karencfv (approved):

gtback (approved):
Looks fine to me. Are we okay with the pages at the old URLs going away, or are we planning to set up forwards?

karencfv:
Thanks @gtback! These docs are al little different than the rest of the elastic docs since they have to live in the official [terraform provider registry](https://registry.terraform.io/providers/elastic/ec/latest/docs) instead of the Elastic docs. Essentially all we have is a page with a [link](https://www.elastic.co/guide/en/tpec/current/index.html) to the provider docs. This change is only so this [page](https://www.elastic.co/guide/index.html) says 1.1 instead of 1.0-beta :)

gtback:
> Thanks @gtback! These docs are al little different than the rest of the elastic docs since they have to live in the official [terraform provider registry](https://registry.terraform.io/providers/elastic/ec/latest/docs) instead of the Elastic docs. Essentially all we have is a page with a [link](https://www.elastic.co/guide/en/tpec/current/index.html) to the provider docs. This change is only so this [page](https://www.elastic.co/guide/index.html) says 1.1 instead of 1.0-beta :)

Thanks, @karencfv . I still think it's probably a good idea to set up a redirect. @nrichers should be able to handle that, but I'm happy to help as well.

gtback:
@nrichers @karencfv is there anything still needed on this PR? I wanted to make sure you weren't waiting on me for anything.

karencfv:
Heya @gtback not from my side, I'm not sure what @nrichers has to say about the previous comment, so just waiting on him I think 😄

nrichers:
> @nrichers should be able to handle that, but I'm happy to help as well.

@gtback catching up with GH notifications ... I've usually requested redirects via Sylvie. Is there some magic I'm missing here? I've never had access to wherever these get set up. 

Given that these docs are nothing but a pointer to the real docs, I'd be OK with just merging as is, without a redirect.

gtback:
Yep, I was referring to the normal process of requesting redirects via the web team. I don't feel strongly between creating a redirect or just merging this as-is.

PR #2105, opened 2021-04-14
Change current to MS 55 for ESS

obierlaire:
Related to https://github.com/elastic/dev/issues/1659

As MS-55 has now been deployed in ESS production, this PR bumps the docs to use MS 55 as current for ESS and the Heroku docs


kellyemurphy (approved):
LGTM

gtback (approved):
This will conflict with #2102 , but it should be easy to resolve depending on which gets merged first.

PR #2108, opened 2021-04-19
Add Fleet Server to shared attributes list

dedemorton:
Updated for new Fleet Server that will be added in 7.13.


bmorelli25 (approved):
🏅

PR #2129, opened 2021-06-04
added ios agent

bryce-b:
updating conf.yml to include iOS agent docs. 
This PR is intended for a 7.14 release.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -8,6 +8,7 @@ repos:
     apm-agent-java:       https://github.com/elastic/apm-agent-java.git
     apm-agent-dotnet:     https://github.com/elastic/apm-agent-dotnet.git
     apm-agent-php:        https://github.com/elastic/apm-agent-php.git
+    apm-agent-ios:        https://github.com/elastic/apm-agent-ios.git
     azure-marketplace:    https://github.com/elastic/azure-marketplace.git
     beats:                https://github.com/elastic/beats.git
     clients-team:         https://github.com/elastic/clients-team.git
@@ -1223,6 +1224,22 @@ contents:
                         repo:   apm-agent-rum-js
                         path:   CHANGELOG.asciidoc
                         exclude_branches:   [ 3.x, 2.x, 1.x, 0.x ]
+                  - title: APM iOS Agent
+                    prefix: swift
+                    current: 0.x
+                    branches: [ main ]
+                    live: [ main ]
+                    index: docs/index.asciidoc
+                    tags: APM iOS Agent/Reference
+                    subject: APM
+                    chunk: 1
+                    sources:
+                      -
+                        repo: apm-agent-ios
+                        path: docs
+                      -
+                        repo: apm-agent-ios
+                        path: CHANGELOG.asciidoc
           - title:      ECS logging
             base_dir:   en/ecs-logging
             sections:
--- a/conf.yaml
+++ b/conf.yaml
@@ -1107,6 +1107,22 @@ contents:
                         repo:   apm-agent-go
                         path:   CHANGELOG.asciidoc
                         exclude_branches:   [ 0.5 ]
+                  - title: APM iOS Agent
+                    prefix: swift
+                    current: main
+                    branches: [ main ]
+                    live: [ main ]
+                    index: docs/index.asciidoc
+                    tags: APM iOS Agent/Reference
+                    subject: APM
+                    chunk: 1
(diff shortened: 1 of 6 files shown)


bmorelli25:
Thanks, Bryce! I pushed a few more changes. We'll see if the doc build comes back green.

Tagging @gtback to take a look at this as well. 👋 Greg. As far as I can tell, this is the first Elastic repository to be added to the doc build with a `main` branch instead of `master`. I imagine this might break a few things. Has a conversation been started around how to handle this difference?

conf.yaml
```diff
@@ -1106,6 +1107,22 @@ contents:
                         repo:   apm-agent-go
                         path:   CHANGELOG.asciidoc
                         exclude_branches:   [ 0.5 ]
+                  - title: APM iOS Agent
+                    prefix: swift
+                    current: main
```
bmorelli25, line 1112: @bryce-b I changed this to `main` because there is no `0.x` branch in the apm-ios-agent repo yet. The `current` variable, `branches` array, and `live` array, must match branch names in the associated repo

bmorelli25: I also moved this entire section up to maintain alphabetical ordering for the APM agents

shared/versions/stack/7.13.asciidoc
```diff
@@ -28,6 +28,7 @@ is-current-version can be: true | false
 APM Agent versions
 ////
 :apm-go-branch:         1.x
+:apm-ios-branch:        main
```
bmorelli25, line 31: @bryce-b these files are used to map links from stack-versioned docs to the agent documentation. These branches represent the version of the documentation linked to from the stack.

As an example, this file is named 7.13.asciidoc. So any links from 7.13 stack docs will go to the main branch of ios, or the 1.x branch of the go agent

shared/attributes.asciidoc
```diff
@@ -55,6 +55,7 @@ endif::[]
 :apm-go-ref:           https://www.elastic.co/guide/en/apm/agent/go/current
 :apm-dotnet-ref:       https://www.elastic.co/guide/en/apm/agent/dotnet/current
 :apm-php-ref:          https://www.elastic.co/guide/en/apm/agent/php/current
+:apm-ios-ref:          https://www.elastic.co/guide/en/apm/agent/ios/current
```
bmorelli25, line 58: These attributes are used for cross-doc linking. For example, when linking from the stack or another agent to the ios docs, instead of copy/pasting the URL, you'd use this attribute + the id.

```
{apm-ios-ref}/setup.html[link text]
```

This enables link checking in our docs.

gtback: We define `prefix` as `swift` in `conf.yaml. That means this should be: 
```suggestion
:apm-ios-ref:          https://www.elastic.co/guide/en/apm/agent/swift/current
```
Unless you want to make it `ios` everwhere.

gtback:
Thanks @bmorelli25! And hi, @bryce-b!

> As far as I can tell, this is the first Elastic repository to be added to the doc build with a `main` branch instead of `master`. I imagine this might break a few things. Has a conversation been started around how to handle this difference?

Yes, I believe it is. We've started talking about it, but I haven't directly addressed anything in this repo yet. It seems to have built just fine, but I'll look through the code more closely next week to see if there's anything I need to fix, and add some tests just to be on the safe side.

bmorelli25:
> It seems to have built just fine

Right now there are no cross-doc links to or from this book. I'm wondering if we should add a few to test things out?

gtback:
I can't think of any reason why cross-book links wouldn't work (famous last words!) in a new book since they just point to whatever URL is defined. It's only when we get to some of the existing books whose outbound links substitute "the same branch/version I'm currently building" that could cause problems when one repo uses `master` and the other uses `main`.

shared/versions/stack/7.13.asciidoc
bryce-b: gotcha! thanks for the clarification.

shared/attributes.asciidoc
bryce-b: I think swift is more appropriate

bmorelli25: Good catch

bmorelli25 (approved):
LGTM! Let's wait for @gtback to 👍 before merging.

gtback (approved):
LGTM!

Issue #2039, opened 2021-01-14
Building ES docs against git worktree fails


henningandersen:
Using git worktree for different branches of ES is common. Running the docs build in a worktree (or perhaps just a folder not called elasticsearch) does not seem to work.

My setup is that I have a GIT_HOME with:
```
docs
elasticsearch (master)
elasticsearch-7.x (worktree)
elasticsearch-7.11 (worktree)
...
```

I can successfully run the docs build from the master git directory. But if I run it from any of my worktrees it fails:

```
cd $GIT_HOME/elasticsearch
git worktree add ../elasticsearch-7.x 7.x
cd $GIT_HOME/elasticsearch-7.x
$GIT_HOME/docs/build_docs --doc $GIT_HOME/elasticsearch-7.x/docs/reference/index.asciidoc --resource=$GIT_HOME/elasticsearch-7.x/x-pack/docs/ --chunk 1
INFO:build_docs:Building HTML from /doc/elasticsearch-7.x/docs/reference/index.asciidoc
INFO:build_docs:Couldn't find repo toplevel for /doc/elasticsearch-7.x/docs/reference
INFO:build_docs:Guessed toplevel=[/doc/elasticsearch-7.x/docs/reference] remote=[0] branch=[master] repo=[repo]
INFO:build_docs:Couldn't find repo toplevel for /doc/elasticsearch-7.x/x-pack/docs
INFO:build_docs:
INFO:build_docs:asciidoctor: ERROR: modules/cluster/misc.asciidoc: line 123: include file not found: /doc/elasticsearch-7.x/docs/reference/modules/cluster/{elasticsearch-root}/docs/reference/indices/dangling-indices-list.asciidoc
...
(lots of errors)
```

I can workaround this by checking out the branch in my `master` `elasticsearch` dir, but it is a bit inconvenient. Thus not urgent in any way.

It seems worktree support was added in #860.

gtback:
Thanks, @henningandersen!

> perhaps just a folder not called elasticsearch

I think this is the important factor. 

As I understand it, #860 allowed `$GIT_HOME/elasticsearch/` to be a worktree, but there's a lot of code (and references within AsciiDoc files) that still depends on it being named exactly `elasticsearch`. If you made the "main working tree" at `elasticsearch-master` and then a "linked working tree" with 7.x at `elasticsearch`, it should still work. 

[...]

PR #2130, opened 2021-06-07
Support running docs builds against worktrees

tvernum:
The Elasticsearch convention is to use git worktrees to support
multiple lines of development across different branches.

A typical layout would be:
 - A regular checkout at `$GIT_HOME/elasticsearch` for the `master` branch
 - A worktree at `$GIT_HOME/elasticsearch-7.x` for the `7.x` branch
 - A worktree at `$GIT_HOME/elasticsearch-6.x` for the `6.x` branch

This commit changes the docs build to support this structure.

### DIFF
--- a/build_docs
+++ b/build_docs
@@ -169,7 +169,13 @@ def run_build_docs(args):
             extra_name = repo_name.replace('x-pack-', '')
             repo_mount = '/doc/%s-extra/%s' % (extra_name, repo_name)
         else:
-            repo_mount = '/doc/' + repo_name
+            match = re.search(r'(.*)-[0-9]+\.[0-9x]+$', repo_name)
+            if match is None:
+                repo_mount = '/doc/' + repo_name
+            else:
+                # Repo name ends with what looks like a version string (e.g. elasticsearch-7.x)
+                # Mount this under the unversioned project name
+                repo_mount = '/doc/' + match.group(1)
         if repo_root not in mounted_doc_repo_roots:
             if repo_name in mounted_doc_repo_names:
                 raise ArgError("Can't mount two repos with the same " +
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -155,7 +155,7 @@ sub _guess_opts {
     my $toplevel = _find_toplevel( $index->parent );
     my $remote = _pick_best_remote( $toplevel );
     my $branch = _guess_branch( $toplevel );
-    my $repo_name = _guess_repo_name( $remote );
+    my $repo_name = _guess_repo_name( $remote, $toplevel );
     # We couldn't find the top level so lets make a wild guess.
     $toplevel = $index->parent unless $toplevel;
     printf "Guessed toplevel=[%s] remote=[%s] branch=[%s] repo=[%s]\n", $toplevel, $remote, $branch, $repo_name;
@@ -169,7 +169,7 @@ sub _guess_opts {
         $toplevel = _find_toplevel( $resource );
         $remote = _pick_best_remote( $toplevel );
         $branch = _guess_branch( $toplevel );
-        $repo_name = _guess_repo_name( $remote );
+        $repo_name = _guess_repo_name( $remote, $toplevel );
         # We couldn't find the top level so lets make a wild guess.
         $toplevel = $resource unless $toplevel;
         $Opts->{roots}{ $repo_name } = $toplevel;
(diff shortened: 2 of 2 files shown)


tvernum:
It's been more than 10 years since I was a perl engineer, so my perl code probably looks more like Java than idiomatic perl.
My python code probably does too :)

There's no tests, because I really wanted to check whether this was useful or not before I put any more time into it.

I originally intended to just detect the worktree style, and fail the build quickly so that ES engineers don't spend time trying to debug it themselves, but then it looked like it might be possible to fix and I went down that rabbit hole.

I don't know what other development practices other projects use, so it's possible this breaks something for another team. I'm happy to iterate on it if needed - or just hand it over if you'd like to run with it yourself.

build_docs.pl
```diff
@@ -188,6 +188,23 @@ sub _find_toplevel {
...
+        my $git_stat = $git->stat();
+        return $git->parent if $git_stat;
+        my $p = $d->parent;
+        if ($p eq $d) {
+            # Root of this filesystem
+            $d = undef;
+        } else {
+            $d = $p;
+        }
+    }
```
gtback, line 206: Can we get rid of the `unless $toplevel` below, and then just `return 0`?

build_docs.pl
```diff
@@ -188,6 +188,23 @@ sub _find_toplevel {
...
+    my $d = $docpath;
+    while ($d) {
+        printf "Checking for .git directory (or file) in '%s'\n", $d;
+        my $git = $d->file(".git");
+        my $git_stat = $git->stat();
+        return $git->parent if $git_stat;
+        my $p = $d->parent;
+        if ($p eq $d) {
+            # Root of this filesystem
+            $d = undef;
```
gtback, line 202: Can we just break here? Or print an error and return immediately? Is there any other way `$d` could become undefined in this loop?

build_docs
```diff
@@ -169,7 +169,13 @@ def run_build_docs(args):
             extra_name = repo_name.replace('x-pack-', '')
             repo_mount = '/doc/%s-extra/%s' % (extra_name, repo_name)
         else:
-            repo_mount = '/doc/' + repo_name
+            match = re.search(r'(.*)-[0-9]+\.[0-9x]+$', repo_name)
```
gtback, line 172: This seems like a pretty safe way to detect "repo names" that include a `-<version branch>` suffix, and is unlikely to inadvertently catch a non-worktree repo name; the only repo that contains a digit right now is `cloud-on-k8s`.

Is this pattern for worktree names (`elasticsearch`, `elasticsearch-7.x`, `elasticsearch-6.x`) pretty universal and documented anywhere? I wouldn't want to have a partial solution that only works for this specific setup if it's not universal; if I understand correctly, any other worktrees that don't match this pattern would still fail, right?

Would it be possible to parse the `.git` file (which on my machine seems to contain something like `gitdir: /absolute/path/to/elasticsearch/.git/worktrees/elasticsearch-7.x` and find the repo name by looking at the path component before `.git`?

gtback:
Related: #2039

gtback (commented):
Thanks for taking the time to look at this, @tvernum. I think this is valuable and worth including; I just had a couple questions about the implementation. I'm not a Perl expert, nor a frequent user of Git worktrees. I don't think it's likely to break anyone's workflows either. 

We do have an integration test suite that could probably be expanded to include this as well. 

Is support for worktrees only needed for the `--resource` argument, or also for `--sub_dir`?

gtback:
@tvernum I haven't looked into why the tests are failing. Are you able to do that?

tvernum:
Yes, now that we've branched for 7.15 I should have some cycles to look at this again.

build_docs
jakelandis: I don't think this would match the pattern I am using for my work trees. I am not aware of any conventions/documentation for how to create these (I have a tiny bash script for personal use)
```
~/workspace/elasticsearch $ ls
7.0  7.1  7.10  7.11  7.12  7.13  7.14  7.15  7.16  7.17  7.2  7.3  7.4  7.5  7.6  7.7  7.8  7.9  8.0  8.1  8.2  main
```

tvernum: This is the format that we recommend in the ES development docs
```
$ git fetch origin 7.x
$ git checkout -b 7.x origin/7.x && git checkout master
$ git worktree add ../elasticsearch-7.x 7.x
```

I'll try and pick this PR up again and see if I can make it agnostic to the worktree name.

jakelandis: Ahh.. i had missed those.  Not a big fan of that layout and the instructions needed updating anyway ... so I am taking a swing at changing the recommendation : https://github.com/elastic/elasticsearch-team/pull/882

PR #2133, opened 2021-06-09
Add shared attribute for Rules and Connectors

lcawl:
Per https://github.com/elastic/kibana/pull/93597, in 7.13.0 and later releases, the Stack Management > "Alerts and Actions" app in Kibana is renamed to "Rules and Connectors". Since we cannot change the shared attribute "alerts-ui" across all the versions that use this file, I've added a new "rules-ui".


gtback (approved):

gchaps (approved):
LGTM

PR #2142, opened 2021-06-29
Add attributes for units sections

debadair:
Adding attributes for the time units & byte units sections that are cross-linked from outside of the ES guide.


lcawl (approved):
LGTM

PR #2153, opened 2021-07-14
Remove extra padding around images

dedemorton:
Link to preview: https://docs_2153.docs-preview.app.elstc.co/guide/index.html

About a year ago, a PR was merged that added padding around images. The change reduced the size of our images to 85% of the original size, resulting in many screen captures that are difficult to read. Most teams did not have an opportunity to review the PR and assess the impact on their published docs.

This PR proposes that we remove the extra padding to maximize the space available for screen captures. (Manipulating the viewport size to get a larger font size in the screen capture is not a good solution to this problem because it forces odd text wraps in the captured screen.)

I've requested a review from @gtback, but I'd like for all the writing teams to weigh in.

Here are some screens that show the difference.

Before:
image

After:
image

### DIFF
--- a/resources/web/style/img.pcss
+++ b/resources/web/style/img.pcss
@@ -15,10 +15,9 @@
   }
   .imageblock {
     /* Asciidoctor emits the title *after* the image but we want it above. */
-    margin: 32px 0 32px 20px;
+    margin: 32px 0 32px 0;
     display: flex;
     flex-direction: column-reverse;
-    width: 85%;
   }
   .screenshot img {
     box-shadow:  0 4px 8px 0 rgba(0, 0, 0, 0.2), 0 6px 20px 0 rgba(0, 0, 0, 0.19);


szabosteve:
I'm in favor of this change.
Sometimes large parts of the UI need to be captured in a single screenshot otherwise users will miss the visual context or the documented UI functionality cannot be presented effectively. In these cases, it is really hard to produce legible screenshots. The [classification example](https://www.elastic.co/guide/en/machine-learning/current/flightdata-classification.html) in the ML book illustrates this case well. Only the last screenshot on the page is truly legible which captures only a small proportion of the screen. Removing the extra padding would make it easier to produce more useful screenshots.
Thank you for drafting this change!

gtback (approved):
I'm fine with this change from the technical side, but like @dedemorton said, I'd like to get feedback from the writers as well.

debadair:
+1 for removing the extra padding. In the trade-off between esthetics and usability, usability wins.

gchaps:
++ to this change.  It makes the screenshots easier to read and the page easier to scan.

PR #2169, opened 2021-08-03
Add shared attributes for kibana-ref-all and security-guide-all

lcawl:
These attributes are useful in the What's New "other versions" links.


jmikell821 (approved):
Looks good!

PR #2171, opened 2021-08-04
Add attribute for index pattern management

lcawl:
The "index pattern management" feature is referenced from places like https://www.elastic.co/guide/en/kibana/master/index-patterns.html and https://www.elastic.co/guide/en/machine-learning/master/setup.html, so this pr adds an attribute as a shortcut


gchaps (approved):
LGTM

PR #2175, opened 2021-08-10
Adds 8.0.0-alpha1

KOTungseth:


lcawl (approved):
LGTM!

lcawl:
Upon closer inspection, it looks like this change is resulting in the following:

> 14:10:11 INFO:build_docs: -                             Kibana Guide: Building 8.0.0-alpha1...
14:11:02 INFO:build_docs: -                             Kibana Guide: Deleting old branch master

... which I think means we won't have https://www.elastic.co/guide/en/kibana/master/index.html anymore, for example.

If we really want a "8.0.0-alpha1" copy of the docs, I think we'd want it _in addition to_ the "master" docs.  However, now I'm wondering why we want alpha docs at all. Maybe @debadair knows the history for this?

The log also shows this error:

> 14:20:01 INFO:build_docs:failed to find kibana links file;
14:20:01 INFO:build_docs:Error executing: git show 8.0.0-alpha1:src/core/public/doc_links/doc_links_service.ts in GIT_DIR /docs_build/.repos/kibana.git

Which means if we *do* want alpha docs, we'll likely need to fix our Kibana link checking code.  I doubt it existed the last time we did alpha/beta docs.

PR #2179, opened 2021-08-12
Make main.asciidoc attributes a symlink

gtback:
I'm raising this as a draft PR for now. If it works, we can talk about whether it will make maintenance easier or if we expect these files to eventually diverge.


lcawl:

lcawl (approved):
They might diverge at some point in the future, but this seems to work and avoids redundancy, so let's go for it!

gtback:
Thanks @lcawl !

Issue #1895, opened 2020-07-14
Add CSS and Styles from linked-tabs widget


gtback:
The CSS and JavaScript in [this file](https://github.com/elastic/beats/blob/master/libbeat/docs/tab-widgets/code.asciidoc) is copied and included directly on several pages in the beats repo. 

Add these to our global styles and scripts to reduce bloated HTML pages, ensure de-duplication, and allow this widget to
be re-used in other books.

gtback:
I didn't do any of this 😂

PR #2180, opened 2021-08-16
Add tabbed widget JS and CSS to build

bmorelli25:
### Summary

We're now using the tabbed widget in at least 10 different documentation books! This PR adds the tabbed widget JS and CSS to the documentation build. I'm confident I've added the CSS correctly, but the JavaScript is another story. It works 😮, but I'm not sure if there's a better way to go about this. @gtback, if this is comically wrong, just let me know and I'll close this PR 😬

### How to test this

1. Checkout this PR: `gh pr checkout 2180`
2. Add a simple tabbed widget to any documentation page. Here's a boilerplate widget:

```
++++
<div class="tabs" data-tab-group="config-agent">
  <div role="tablist" aria-label="Configure agents">
    <button role="tab"
            aria-selected="true"
            aria-controls="cloud-tab-config-agent"
            id="cloud-config-agent">
      Central configuration
    </button>
    <button role="tab"
            aria-selected="false"
            aria-controls="self-managed-tab-config-agent"
            id="self-managed-config-agent"
            tabindex="-1">
      Agent configuration
    </button>
  </div>
  <div tabindex="0"
       role="tabpanel"
       id="cloud-tab-config-agent"
       aria-labelledby="cloud-config-agent">
++++

This is the content for tab #1

++++
  </div>
  <div tabindex="0"
       role="tabpanel"
       id="self-managed-tab-config-agent"
       aria-labelledby="self-managed-config-agent"
       hidden="">
++++

This is the content for tab #2

++++
  </div>
</div>
++++
```
3. Build that documentation book and ensure the widget works as expected

### Related issues

Closes https://github.com/elastic/docs/issues/1895.


gtback (changes requested):
I think we'll want to change the JS so it gets included in `/guide/static/docs.js` (and thus can be loaded once and cached) rather than inlining it on every page (increasing page sizes and slowing down page loads, even if only marginally), even pages that don't use the widgets.

I'll have to refresh my memory on how that all gets built, but if you want to try to figure it out, great!

bmorelli25:
Moved JS to `resources/web/lib/prettify/prettify.js`. That _might_ be right. When running the doc build locally, it looks like JS isn't minified. I'm able to see the new JS when I navigate to http://localhost:8000/guide/static/docs.js, and the JS is applied to the tabbed widgets. But there's a new problem -- this breaks the CSS for some reason. Looking into that now.

bmorelli25:
OK, I think this is ready for review again. I can see the JS at http://localhost:8000/guide/static/docs.js, and the CSS at http://localhost:8000/guide/static/styles.css

resources/web/lib/prettify/prettify.js
```diff
@@ -1731,6 +1731,93 @@ var prettyPrint;
              : (prettyPrint = $prettyPrint)
       };
 
+// BEGIN TABBED WIDGET JS
```
gtback, line 1734: This file is all third-party code from Google that we've copied into the repo. I believe the best way to add this is to add a new file under `resources/web/docs_js/components/` and include an import at the top of `resources/web/docs_js/index.js`.

Sorry for the runaround. I meant to look at the right place to do this before you made any changes, but you're too fast! 🏃🏻

gtback (changes requested):

resources/web/lib/prettify/prettify.js
gtback: It should end up in the same place in the output: `static/docs.js`.

bmorelli25: Awesome! Thank you. No worries, that's my fault for trying to figure it out on my own.

bmorelli25: The code in `docs_js/components` seems to be for functionality that is actually coded into the doc build -- like the console widgets. Are you sure this is where this JS should go?

bmorelli25: To clarify, it does work – I just want to call out the difference.

bmorelli25: Moved in d8b9683.

resources/web/docs_js/index.js
```diff
@@ -281,3 +282,8 @@ $(function() {
 
   // Test comment used to detect unminifed JS in tests
 });
+
+// Tabbed widgets
+switchTabs();
+setActiveTab();
+changeTabs();
```
bmorelli25, line 289: This is a stab in the dark, but it works. I tried making one massive function, but that didn't work. I also tried invoking the functions in `./components/tabbed_widget`, but that didn't work.

resources/web/lib/prettify/prettify.js
gtback: > The code in `docs_js/components` seems to be for functionality that is actually coded into the doc build, like the console widgets. Are you sure this is where this JS should go?

I'm not sure, but I think so. Aren't both this and the console widgets pieces of interactive content in the docs? I guess I'm not sure what you see as the difference.

gtback:
It seems like the test is failing because it's trying to set the active tab on a test page that doesn't have the tab widget?

gtback (approved):
This is looking good! I think we just need to update the functions to not do anything if there are no tab widgets on the page.

resources/web/lib/prettify/prettify.js
bmorelli25: Maybe this doesn't matter, but the difference I see is that the console widget is something that the doc build converts from Asciidoc to HTML. On the other hand, the tabbed widget is raw HTML that we add to an Asciidoc file; the doc build doesn't need to converting anything.

bmorelli25:
> I think we just need to update the functions to not do anything if there are no tab widgets on the page.

Interesting! OK, let me try to figure that out 🤔

bmorelli25:
I have an idea of how to fix this. Is there a way to run `self-test` locally so I don't have to push and wait for it to run?

> `TypeError: Cannot read property 'parentNode' of undefined`

I think this is failing because the `setActiveTab` function [requires](https://github.com/elastic/docs/pull/2180/files#diff-86b9d0a3795b8538b2f06daafb92455f13cb67590425287a4074b0bb8903559dR36) a `target` node parameter. When the page is built, and someone clicks on a tab widget, it receives one. But when the tests run, which invokes the `setActiveTab` function [here](https://github.com/elastic/docs/pull/2180/files#diff-9cf98e27414725688b027f7b277f901546bbaab87519b879e5fcf838efb22cbeR288), no `target` parameter is passed in.
As it turns out, the `setActiveTab` and `changeTabs` functions don't need to be exported and invoked in `index.js`. I _think_ this should fix the problem.

gtback:
> Is there a way to run `self-test` locally so I don't have to push and wait for it to run?

@bmorelli25 You can change into the `resources/web` directory and run `make check` to run just these tests. You can also run `make check` from the root of the repo to run the full test suite (which is pretty slow because of the integration tests) or `make unit_test` for just the "fast" tests. 

Looks like the tests are passing with your latest commit though.

gtback:
Once you get the tests passing, feel free to merge!

bmorelli25:
Will do! It looks like that fixed `self-test` 🎉 . I'm going to add some docs to the readme before merging. Thanks for all of your help!

Issue #1930, opened 2020-08-03
Title TOC Entry and version dropdown do not appear in some cases  


gtback:
To reproduce:
- Find a "current" version where the branch name and version name do not match. (For example ECE 2.6, which comes from the release-ms-41 branch).
- Note that the page `/guide/en/cloud-enterprise/current/index.html` contains a "Elastic Cloud Enterprise - Elastic Cloud on your Infrastructure" TOC entry, and a dropdown for other versions, but `/guide/en/cloud-enterprise/2.6/index.html` (which is otherwise identical) does not.

(Also, investigate whether this affects non-current versions with different branch/version names)

Follow-on to #1850.

gtback:
> (Also, investigate whether this affects non-current versions with different branch/version names)

It does affect any version where the name of the version (directory) doesn't match the name of the branch.

PR #2189, opened 2021-08-19
Add list of live branches that includes main

lcawl:
The books that currently map "main" branches to a "master" output are currently still using "master" in their list of "live" branches, which causes that branch to disappear from the drop-down:

image

This PR updates the conf.yaml to include a "stacklivemain" attribute to address this problem.


gtback:
My gut says this has the same root cause as #1930, but I'm OK with this as a workaround.

gtback (approved):

PR #2204, opened 2021-08-26
Adds data view shared attributes

lcawl:
Related to https://github.com/elastic/kibana/pull/109928 and https://github.com/elastic/kibana/pull/110283

This PR adds shared attributes for data views, so that the terminology can be used appropriately across the library.

This new terminology is specific to 8.0. For earlier branches,  we must either (a) not use the shared attributes, or (b) override them as was done in https://github.com/elastic/kibana/blob/7.x/docs/index.asciidoc


gchaps (approved):
LGTM

PR #2203, opened 2021-08-26
ESS - Change current to MS-62

kilfoyle:
This changes "current" for the Cloud ESS docs to MS-62.

PR #2184, opened 2021-08-17
Change current docs branch to 7.15

chriscressman:
DO NOT MERGE UNTIL RELEASE DAY

PR #2205, opened 2021-08-30
Move shared variables in conf.yaml to the top

gtback:
I also added a script that's used to test changes to ensure YAML refactors don't change the interpreted meaning. I used the script when making this change to ensure that the build would interpret conf.yaml the same way before and after this change.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -2421,7 +2421,6 @@ contents:
             current:    6.2
             subject:    X-Pack
             index:      docs/en/index.asciidoc
-            private:    1
             branches:   [ 6.2, 6.1, 6.0, 5.6, 5.5, 5.4, 5.3, 5.2, 5.1, 5.0 ]
             sources:
               -
@@ -2482,7 +2481,6 @@ contents:
             subject:    Marvel
             current:    2.4
             index:      docs/public/marvel/index.asciidoc
-            private:    1
             branches:   [ 2.4, 2.3, 2.2, 2.1, 2.0, {marvel-1.3: 1.3}]
             sources:
               -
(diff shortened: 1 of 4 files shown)


gtback:
Once this is merged, I'll update:

- [ ] #2184
- [ ] #2203

As well as fixing any PRs with conflicts or new PRs that are trying to use the `mapMasterToMain` and `mapMainToMaster` but needing to move the definitions around.

conf.yaml
```diff
@@ -1345,7 +1369,7 @@ contents:
                   -
                     repo:   ecs-logging
                     path:   docs
-                    map_branches: &mapEcsLoggingJavaToEcsLogging
```
bmorelli25, line 1348: Why did this one disappear instead of moving to the top?

lcawl: I assume because it's only used once and therefore the attribute isn't necessary?

bmorelli25: Ah, that makes sense. Thanks!

conf.yaml
```diff
@@ -2397,7 +2421,6 @@ contents:
             current:    6.2
             subject:    X-Pack
             index:      docs/en/index.asciidoc
-            private:    1
```
lcawl, line 2400: > private: 1

I suspect these "private" settings were used to omit "edit URLs" from these legacy docs.  Is this change necessary?

conf.yaml
gtback: Yep, there were a couple that were only used once, so I just removed them. I thought about including them anyway, but then realized there was no reason to not make all of the lists into variables.

conf.yaml
gtback: `private` is defined twice for these books (see line 2418 in the new version or 2394 in the old version). I've been meaning to clean this up forever (VS Code reports the problem when I'm editing the conf.yaml file)

szabosteve (approved):
LGTM! Thanks! 🚀

lcawl (approved):
LGTM2, thanks!

PR #2211, opened 2021-09-03
Switch from master to main for kibana repo

lcawl:
Relates to https://github.com/elastic/kibana/issues/97827

This PR pulls content from the "main" branch of the kibana repo instead of the "master" branch.


gtback (approved):
I think this is fine, as long as we can quickly revert if anything goes wrong. With this change, we're still building a "master" version, even though it's using the "main" branch (which for now is just automatically updated with all commits to `master`). I believe the Kibana link checking only worries about the name of the files on disk (the "version", not the "branch"), so this change is effectively a no-op as far as that's concerned.

tylersmalley:
Thanks, @lcawl - reach out if you have any questions. The main branch should exist now and is being sync'd.

lcawl:
This change was reverted in https://github.com/elastic/docs/pull/2215

PR #2213, opened 2021-09-08
Hide X-Pack tags on versions 7.11 and later

### LINKED ISSUE #2207: Remove "x-pack" icon on newer versions of the docs
In many places in the docs, we have an "X-Pack" icon that appears next to headings, with a link to https://www.elastic.co/subscriptions. We don't use the term "X-Pack" anymore, so this is at best confusing. 

We'd like to remove this icon from all versions 7.11 and newer.

gtback:
Eventually, we'll want to remove the x-pack role from the Asciidoc for
all sections/headers that include it, but until then this will at least
hide the visual indicator and link. The `xpack` class will be included
on the surrounding divs, since this is the default Asciidoctor behavior
for "role" tags.

Fixes #2207.

### DIFF
--- a/integtest/spec/single_book_spec.rb
+++ b/integtest/spec/single_book_spec.rb
@@ -370,19 +370,28 @@ RSpec.describe 'building a single book' do
       end
     end
 
-    def self.xpack_tag_context(onpart, onchapter, onfloater, onsection)
+    def self.xpack_tag_context(onpart, onchapter, onfloater, onsection,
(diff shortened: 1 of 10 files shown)


gtback:
cc: @debadair

nik9000 (approved):

debadair:
Exciting!

olksdr (approved):
looks like there are some lint issues? 
otherwise LGTM

PR #2218, opened 2021-09-16
Add ECK 1.8 to the list of ECK branches

sebgl:
We are about to release ECK 1.8.0, this commit adds the 1.8 branch
to the list of branches to build docs for.

Note this should be merged slightly before the release.

PR #2219, opened 2021-09-16
Set ECK 1.8 as the current docs branch

sebgl:
This sets up ECK 1.8 as the current ECK release.

Note this should not be merged before the release date.
https://github.com/elastic/docs/pull/2218 should be merged first so we have extra time to double-check everything is correct :) 

(I can handle both merges)

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -877,8 +877,8 @@ contents:
             prefix:     en/cloud-on-k8s
             tags:       Kubernetes/Reference
             subject:    ECK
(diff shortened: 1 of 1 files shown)


gtback (approved):
LGTM! I'm going to mark this as a draft so it doesn't get merged early.

sebgl:
Test failure should be fixed with elastic/cloud-on-k8s#4849.

@elasticmachine test this please

PR #2238, opened 2021-10-04
Rename Fleet User Guide

dedemorton:
Changes the name of the Fleet User Guide as described in https://github.com/elastic/observability-docs/issues/1104.

@gtback I think we should change the following settings, too, but TBH, I don't remember how these settings are used, so I'm unsure what settings make sense.

````
tags:       Fleet/Guide
subject:    Fleet
````

Maybe this?

````
tags:       Fleet/Elastic Agent/Guide
subject:    Fleet and Elastic Agent
````

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -1621,8 +1621,8 @@ contents:
             live:       *stacklive
             index:      docs/en/ingest-management/index.asciidoc
             chunk:      2
-            tags:       Fleet/Guide
-            subject:    Fleet
+            tags:       Fleet/Guide/Elastic Agent
+            subject:    Fleet and Elastic Agent
             sources:
               -
                 repo:   observability-docs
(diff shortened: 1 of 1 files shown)


gtback (approved):

dedemorton:
> Did you mean to link this PR to itself?

Haha...no.

dedemorton:
@gtback Can you take another look? Given your response, I do think it's worth adding Elastic Agent to those tags. Thanks!

gtback (approved):
LGTM!

AnneB-SEO:
@gtback many crawlers, including Google use DC tag data to identify document but prefer schema.org tagging mainly because the latter was developed by and for search engines. Could we consider using both @goodroot ?

AnneB-SEO:
Right @gtback - I was only thinking of the new docs. Appreciate the clarification

PR #2240, opened 2021-10-05
ESS - Change current to MS-63

kilfoyle:
This changes "current" for the Cloud ESS docs to MS-63.


andreaturli (approved):
thank @kilfoyle

gtback (approved):
LGTM! It looks like all the other references to ms-62 are for ECE.

PR #2250, opened 2021-10-19
Build `7.x` branch of APM Guide

bmorelli25:
Adds the 7.x branch of the APM Guide to the build. For https://github.com/elastic/observability-docs/issues/1073.


dedemorton (commented):
I wonder if you keep master as `current` instead of changing it to `7.x`. My thought is that when the 7.x branch gets renamed to 7.16, you'll have a version 7.16 book marked as current, which might cause folks to think that 7.16 has been released.

dedemorton (approved):
Looks groovy

PR #2251, opened 2021-10-19
ESS - Change current to MS-64

kellyemurphy:
This changes "current" for the Cloud ESS docs to MS-64.


gtback (approved):
LGTM!. It's so nice that this is only in one place now 😁

PR #2253, opened 2021-10-20
Update `{ref-7x}` to point to `7.16`

bmorelli25:
7.x doesn't exist. Let's point to 7.16 instead!


gtback:
Admin-merging to fix doc builds

PR #2269, opened 2021-10-29
Fix Kibana link checking

gtback:
We need to look at the 'main' branch when checking links for 'master' in the output.


gtback:
This seems to be passing the actual build, but is failing some of the tests which aren't set up to allow Kibana to have a "master" version build from a "main" branch. I'll see if I can easily update the tests on Monday, otherwise I might just disable them so we can get Kibana link checking back working.

gtback:
Ok, I think I have all the relevant tests passing locally. I pushed to get the full test suite to run.

nik9000 (approved):

PR #2274, opened 2021-11-01
Updated shebang to explicitly state python3 (to work via WSL)

ollyhowell:

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -21,7 +21,7 @@ on any other website without explicit permission.
 
 You'll need the following installed:
 
-* Python (Python 3 preferred, but Python 2 will probably be fine)
+* Python 3
 * Docker
 
 [discrete]
--- a/build_docs
+++ b/build_docs
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 # coding=utf-8
 
 # Build the docs with docker!
--- a/build_docs
+++ b/build_docs
@@ -14,19 +14,6 @@
 # I'm aware that this is a fairly non-standard way to use docker but
 # for the most part all we want docker for is to get us into a consistent
 # environment it'll do the trick. At least for now.
-#
-# The shebang line on this script has a story too! As written it targets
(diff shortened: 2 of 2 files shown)


gtback:
@nik9000 as the author of [this fabulous story](https://github.com/elastic/docs/pull/2274/files#diff-f3730194584f7142c530ccdcec434d78e1d4542138bd78287de9446208dd3df5R18-R27), I'm curious if you have any thoughts about this. I don't have an Arch system to test on, but I'd like to make the docs compatible with WSL for @ollyhowell and anyone else on Windows. 

I suppose an alternative would be to make `./build_docs` into a shell script that detects the Python interpreter and then calls a `./build_docs.py` which is a renamed version of the existing `./build_docs` script.

nik9000:
If it works on Mac and Windows that's fine. Arch certainly has `python3`. I chose `python` because, at least at the time, some Mac didn't have a `python3`. I don't even remember which one. But if Mac is fine with `python3` now that'd be great. We can drop python2 support entirely.

We should probably drop the story of the shebang and the reference to pep-394 too.

gtback:
Thanks @nik9000 . I misread the comment. I didn't realize it was macOS that could be the problem. It looks like there's a builtin `python3` command on macOS (I have homebrew, pyenv, and a bunch of other Pythons installed, but this one looks Apple-provided):

```shell
❯ /usr/bin/python3
Python 3.8.2 (default, Apr  8 2021, 23:19:18)
[Clang 12.0.5 (clang-1205.0.22.9)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
```

nik9000:
LGTM then

gtback:
I ended up just ripping out all the Python 2 support

If that causes problems for anyone, we should fix their environments (I'll help with this) rather than continuing to support Python 2.

gtback:
@justincr-elastic Nope, not documented anywhere as far as I know. @ollyhowell is the only person I know who's using Windows/WSL, so if you're able to work with him and write some documentation, that would be great!

PR #2278, opened 2021-11-03
Adds attribute to data views

szabosteve:
This PR introduces two attributes to the list:
* `data-sources-caps`. Resolves to Data Views. 
* `data-sources-cap`. Resolves to Data views.


lcawl (approved):
LGTM!

PR #2281, opened 2021-11-08
docs: add obs-docs dep to apm-server

bmorelli25:
Add `elastic/observability-docs` as a dependency to the APM Guide.

Required so that the Traces quick start guide can live in both the Observability guide and the APM guide. See https://github.com/elastic/observability-docs/issues/1231.


gtback (approved):
LGTM!

PR #2285, opened 2021-11-15
Update lists that follow paragraphs

lockewritesdocs:
Currently, lists that follow paragraphs get mushed together in tabbed widgets. This change aims to fix that smushing.

image

### DIFF
--- a/resources/web/style/list.pcss
+++ b/resources/web/style/list.pcss
@@ -38,14 +38,4 @@
       }
     }
   }
-  /* Moves lists that directly follow paragraphs inside of dd and li close under
-   * the paragraph so they look more "togeter". */
-  dd, li {
-    & > p + div {
-      &.orderedlist,
-      &.itemizedlist {
-        margin-top: 1em;
-      }
-    }
-  }
 }
--- a/resources/web/style/list.pcss
+++ b/resources/web/style/list.pcss
@@ -44,7 +44,7 @@
     & > p + div {
       &.orderedlist,
       &.itemizedlist {
-        margin-top: -0.9em;
+        margin-top: 1em;
       }
     }
   }


resources/web/style/list.pcss
```diff
@@ -44,8 +44,8 @@
     & > p + div {
       &.orderedlist,
       &.itemizedlist {
-        margin-top: -0.9em;
+        margin-top: 1em;
```
gtback, line 47: I'm curious what would happen if you took this out entirely. I can see why a negative margin could cause problems, but the best CSS rule is the one that isn't even needed.

bmorelli25: Agreed. Removing it entirely in dev tools looks fine (in Chrome at least).

bmorelli25 (commented):
This does impact non-tab widget lists. We might need to get more creative with a fix that only applies to tab widget lists--unless everyone is OK with the larger gap between paragraphs and lists:

Currently:
Screen Shot 2021-11-15 at 12 10 20 PM

After removing the rule entirely:
Screen Shot 2021-11-15 at 12 10 25 PM
:

resources/web/style/list.pcss
lockewritesdocs: Remove the `margin-top` only, or delete everything from lines 43 to 51?

bmorelli25: You'd be able to get rid of L41-L50... but only if we're OK with the larger gap shown in my comment [below](https://github.com/elastic/docs/pull/2285#pullrequestreview-806502427).

gtback: Remove any empty blocks/parent-blocks. So probably lines 41-50 (not the overall closing brace, though). I'd be curious to see how this looks after existing `dd` items, or list items that contain nested paragraphs and divs.

lockewritesdocs:
The extra space isn't terrible, though not ideal. Do either of you have cycles to work on a more intelligent fix? If not, can we merge this for now so that I can get this related PR merged? https://github.com/elastic/elasticsearch/pull/80735

bmorelli25:
So this also works -- and maintains the current spacing outside of the tab widgets. But IDK how ideal this is:

```
  /* Moves lists that directly follow paragraphs inside of dd and li close under
   * the paragraph so they look more "togeter". */
  dd, li {
    & > p + div {
      &.orderedlist,
      &.itemizedlist {
        margin-top: -0.9em;
      }
    }
  }
  /* Same as above, but removes margin when lists follow paragraphs inside of tab widgets */
  .tabs {
    dd, li {
      & > p + div {
        &.orderedlist,
        &.itemizedlist {
          margin-top: 0em;
        }
      }
    }
  }
```

I'm sure there's a way in SASS to exclude `.tabs` in one block, but my SASS isn't good enough.

bmorelli25:
LGTM as is. The extra gap isn't that bad. I tried playing around with the `:not()` selector but couldn't quite get things to work right.

gtback (approved):
Ditto to what @bmorelli25 said. I'm fine with this getting merged as-is, but also if we wanted to set up separate logic for lists inside and outside tabs.

PR #2289, opened 2021-11-17
Remove unused script ref

brianjolly:
Removes unused script reference which is generating many 404s.

### DIFF
--- a/resources/web/template.html
+++ b/resources/web/template.html
@@ -126,10 +126,6 @@
       </section>
     </div>
 
-<script type="text/javascript">
-	var suggestionsUrl = "https://search.elastic.co/suggest";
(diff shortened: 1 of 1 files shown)


brianjolly:
https://docs_2289.docs-preview.app.elstc.co/guide/index.html

gtback (approved):
Looks fine to me. I'm guessing all the search stuff comes through the header now?

I also looked and I'm not sure if the `search.elastic.co` URL above this is doing anything, but we can remove that in a later PR if we want.

brianjolly:
Thanks @gtback, yes, the search all comes through the nav with the Site Search setup. 

I'm happy to drop that `search.elastic.co` script too. It looks like it doesn't even have DNS anymore :)

brianjolly:
Thanks @gtback !

PR #2291, opened 2021-11-18
Add tab-widget code to enable Cloud login text from across stack docs

kilfoyle:
We'd like to a tab widget that contains the login info for Cloud products (ESS and ECE, for now). Storing the widget code in the shared `docs` repo would allow us to a create the widget from anywhere in the stack docs, with two simple include statements.

Screen Shot 2021-11-18 at 1 31 50 PM

I've tested the code locally and it works nicely.


lcawl (approved):
Great idea!

PR #2298, opened 2021-11-25
Add attribute for the integrations developer guide

dedemorton:
Adds attribute to resolve path to the integrations developer guide.

@bmorelli25 Should we set this to `current` since we are not building in multiple branches?


bmorelli25 (approved):
Thanks!

gtback (approved):

PR #2304, opened 2021-12-03
Bump 8.x `prev-major-last` to 7.17

jrodewig:
🚀


gtback (approved):
As long as this is just getting used in text, I think we're fine (and it seems to be that way since the builds are passing). If they're going to be used in links, this probably won't work until there are 7.17 branches/versions for each of the books that are the targets of links.

PR #2308, opened 2021-12-06
Add elasticsearch-java doc code snippets directory

swallez:
https://github.com/elastic/elasticsearch-java/pull/52 pulls code snippets in the document from test source code, which is not checked out by the docs build process.

This PR adds the path to that test source code in the `elasticsearch-java` (`java-api-client`) configuration.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -498,6 +498,9 @@ contents:
                     repo:   elasticsearch-java
                     path:   docs/
                   -
+                    repo:   elasticsearch-java
+                    path:   java-client/src/test/java/co/elastic/clients/documentation
+                  -
                     repo:   elasticsearch
                     path:   docs/java-rest/
                   -


szabosteve:
@gtback Could you please have a look at this PR? If I understand correctly, it fails because the referenced folder is empty. [This PR](https://github.com/elastic/elasticsearch-java/pull/52), however, fails on the docs test because the test files are not found as their path is not referenced in the conf.yml.
In my understanding, the solution would be to add the test files to the path in the java repo in a separate PR, then merge this PR in the docs repository, finally reference the test files in the code snippets of the asciidoc files in the java repo. Am I reading the situation right?
CC @swallez

szabosteve:
@elasticmachine test this please

swallez:
@elasticmachine test this please

(added missing files to the `8.0` branch in addition to `main` and `7.16`

szabosteve (commented):
LGTM!

szabosteve (approved):

gtback:
@szabosteve Yes, there are some times when neither PR will pass without the other, so you'll need to just merge one without CI passing and then the other one should pass. It looks like it got figured out in this case.

> (added missing files to the 8.0 branch in addition to main and 7.16)

You can also use `exclude_branches` to say "don't use these files/paths when building this list of versions" (so in this case `exclude_branches: [ 8.0 ]`)

PR #2309, opened 2021-12-07
Move APM Reference and Overview to Legacy section

bmorelli25:
Moves the APM Server Reference and APM Overview to the Legacy docs section of elastic.co/guide.

Merge after https://github.com/elastic/docs/pull/2299.

PR #2324, opened 2021-12-16
Move legacy APM books back to `/en/apm/*`

bmorelli25:
Amends https://github.com/elastic/docs/pull/2309 by moving legacy APM books back to `/en/apm/*`.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -2296,42 +2296,45 @@ contents:
 
     -   title:      Legacy Documentation
         sections:
-          - title:      APM Overview
-            prefix:     get-started
-            index:      docs/guide/index.asciidoc
-            current:    7.15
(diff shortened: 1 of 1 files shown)


gtback:
We can see what this looks like, but I was thinking it might be easier to just manually add `en/apm` prefix on lines 2300/2303 and 2316/2319. As it is, this is likely to break things, since the page https://www.elastic.co/guide/en/apm/index.html can't come from two different sections.

gtback:
Or, honestly, we could move the books back up into the APM section (on that same page linked above), but have a subsection on that page (maybe?)

bmorelli25:
> Or, honestly, we could move the books back up into the APM section (on that same page linked above), but have a subsection on that page (maybe?)

That's a reaaaaaally interesting idea. A legacy section within that subsection would probably make more sense for our users anyways since they're more likely to look in the APM section for these docs.

gtback (approved):
Looks good, assuming the diff shows what we expect. We can clean up the "new" "legacy" docs at the wrong URL in the built-docs repo once this gets merged.

PR #2327, opened 2021-12-17
Fixes eland-docs attribute

lcawl:
The `eland-docs` attribute in https://github.com/elastic/docs/blob/master/shared/attributes.asciidoc has an incorrect URL, which results in broken links. The correct URL should match https://www.elastic.co/guide/en/elasticsearch/client/eland/master/installation.html


jrodewig (approved):
LGTM.

PR #2330, opened 2021-12-21
Update conf.yaml

bmorelli25:
A follow-up to https://github.com/elastic/docs/pull/2324 that adds `Legacy` titles to the old APM books.

Before:

Screen Shot 2021-12-21 at 1 17 46 PM


dedemorton (approved):
LGTM

PR #2342, opened 2022-01-11
Adds attributes for data views

szabosteve:
## Overview

This PR introduces two attributes to the list:

* `data-source-caps`. Resolves to `Data View`.
* `data-source-cap`. Resolves to `Data view`.


gtback (approved):
LGTM! But why are there so many versions of this variable?

szabosteve:
> LGTM! But why are there so many versions of this variable?

Some of the versions are not correct. The system doesn't respect whether the attribute itself contains capitalization or not. I'll delete the incorrect ones once we added the correct versions and changed the docs to use the new attributes.

PR #2346, opened 2022-01-17
ESS - Change current to MS-69

kilfoyle:
This changes "current" for the Cloud ESS docs to MS-69.
Do not merge until release day.

PR #2355, opened 2022-01-19
ESS - Change current to MS-68

alaudazzi:
This changes "current" for the Cloud ESS docs to MS-68.


gtback (approved):
I'm OK with this, but it will conflict with #2346

kilfoyle:
I can fix up #2346 after this one is merged (MS-69 won't release until Feb.1).

PR #2360, opened 2022-01-27
add new path for kibana links check

pgayvallet:
Required for https://github.com/elastic/kibana/pull/123818 / https://github.com/elastic/kibana/issues/95389

Add the new package path for Kibana links checking

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -422,6 +422,12 @@ sub check_kibana_links {
         say "  Branch: $branch, Version: $version";
         my $links_file;
         my $source = eval {
+            $links_file = "packages/kbn-doc-links/src/get_doc_links.ts";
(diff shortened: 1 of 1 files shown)


gtback (approved):
LGTM! Thanks @pgayvallet

gtback:
It would be great to see the Kibana PR fail without these changes (falling through to the `failed to find kibana links file` error), and then succeed once this is merged. But I'm just paranoid because Kibana link checking was incorrectly succeeding because of a stub file in the wrong place, for a long time (elastic/kibana#94274).

gtback:
> It would be great to see the Kibana PR fail without these changes

Nevermind, this won't happen because we still have a stub file in the old location that delegates to the new package.

PR #2364, opened 2022-02-01
Note that the full path is required for local doc build

matschaffer:
When trying to confirm the issues on https://github.com/elastic/beats/pull/30014 I found this readme, but it wasn't clear that I needed a full path until I started reading through https://github.com/elastic/docs/blob/master/doc_build_aliases.sh

Seemed like we should probably clarify that a full path is expected.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -70,7 +70,7 @@ generate the HTML version of the docs:
 [source,bash]
 ----------------------------
 cd path/to/your/repo
-~/path/to/docs/repo/build_docs --doc path/to/index.asciidoc
+~/path/to/docs/repo/build_docs --doc /full/path/to/index.asciidoc
 ----------------------------
 
 Each Elastic project may need its own documentation book build command.


dedemorton (approved):
LGTM

gtback:
Hey @matschaffer, I'm not sure if this is actually required (it doesn't hurt of course, so I'm fine with this change). 

I was able to run this command from my local `security-docs` repo, in order to build the Security docs (the equivalent of [`docbldsec`](https://github.com/elastic/docs/blob/master/doc_build_aliases.sh#L77):

```
../docs/build_docs --doc docs/index.asciidoc --resource ../stack-docs/docs --chunk 1 --open 
```

All of the path arguments can be weird and confusing, and I haven't tried with the beats docs. Using the aliases is recommended, both for consistency and to make sure you have all the relevant files that are needed to build a specific book..

matschaffer:
Hrm. Yeah, it seems to work for me without the full path now. Not sure why it didn't the first time.

Sorry I didn't copy the error message.

PR #2371, opened 2022-02-08
Include the whole cloud/docs directory for ECE

gtback:
We need this in order to grab the ece-version.asciidoc file that's
included in that directory.


kilfoyle (approved):
LGTM! 🚀 
Thanks a lot for fixing this up.

PR #2379, opened 2022-02-14
Add security-docs as a dependency of stack-docs

joepeeples:
Elastic Security docs need to be a dependency of the Elastic Installation and Upgrade Guide, in order for https://github.com/elastic/stack-docs/pull/2039 to build correctly.


conf.yaml
```diff
@@ -165,6 +165,10 @@ contents:
                 path:   docs/src/reference/asciidoc
                 exclude_branches:   [ 6.8, 6.7, 6.6, 6.5, 6.4, 6.3, 6.2, 6.1, 6.0, 5.6, 5.5, 5.4, 5.3, 5.2, 5.1, 5.0 ]
                 map_branches: *mapMainToMaster
+              -
+                repo:   security-docs
+                path:   docs/
+                exclude_branches:   [ 6.8, 6.7, 6.6, 6.5, 6.4, 6.3, 6.2, 6.1, 6.0, 5.6, 5.5, 5.4, 5.3, 5.2, 5.1, 5.0 ]
```
lcawl, line 171: If you're going to only add a dependency in 8.0 and later branches, you can likely add more versions to this exclude_branches list, otherwise LGTM

lcawl (approved):
One comment, otherwise LGTM

conf.yaml
joepeeples: @lcawl Yes, this is for 8.0.0 and later. So should I add every 7.x version to this list?

lcawl: Yes, might as well clearly exclude them

PR #2385, opened 2022-02-25
Update ES clients to use `stackcurrent` and `stacklivemain` variables

jrodewig:
Updates the following ES clients docs to use the `stackcurrent` and `stacklivemain` variables:

- Java
- JavaScript
- Ruby
- Go
- .NET
- PHP
- Python

This will bump the `current` branch for these docs to 8.0. It'll also ensure the clients stay synced with stack releases in the future.

This **does not** update the current version for the Perl or Rust clients.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -485,7 +485,7 @@ contents:
             sections:
               - title:      Java Client
                 prefix:     java-api-client
-                current:    7.16
+                current:    8.0
                 branches:   [ {main: master}, 8.1, 8.0, 7.17, 7.16 ]
                 live:       [ main, 8.1, 8.0, 7.17, 7.16 ]
                 index:      docs/index.asciidoc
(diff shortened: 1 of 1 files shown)


philkra (approved):
LGTM -- I double checked the repos and the branches are in place. Thanks James for opening the PR

jrodewig:
This change is causing some links from the [legacy elasticsearch.js docs](https://www.elastic.co/guide/en/elasticsearch/client/elasticsearch-js/current/index.html) to the [ES JS client docs](https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/8.0/index.html) to break. To fix those links, I've opened https://github.com/elastic/elasticsearch-js/pull/1634 to add some related redirects. I've also reached out to the clients team about updating the legacy docs.

We should also investigate whether we can use the `stackcurrent` and `stacklivemain` variables for these client docs instead of hardcoding the `current` and `live` branches. I'll reach out to the clients team about this on Monday.

jrodewig:
I spoke with @philkra and confirmed that we can use the `stackcurrent` and `stacklivemain` variables.

I'll merge this in once https://github.com/elastic/elasticsearch-js/pull/1634 is merged and backported.

philkra:
Thanks @jrodewig , the PR https://github.com/elastic/elasticsearch-js/pull/1634 is merged on `main`

gtback (approved):
LGTM once the builds pass.

PR #2393, opened 2022-03-08
Publish 8.1 branch of Welcome docs

debadair:
This has the side effect of removing the master version from the published docs, but we expect the impact to be minimal. They were more a side-effect of having a single version than intentional, and were not surfacing in Google searches.

Going forward, we'll publish in sync with the stack versioning.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -120,9 +120,10 @@ contents:
         sections:
           - title:      "Welcome to Elastic"
             prefix:     en/welcome-to-elastic
-            current:    main
+            current:    *stackcurrent
             index:      welcome-to-elastic/index.asciidoc
-            branches:   [ {main: master} ]
+            branches:   [ 8.1 ]
+            live:       *stacklivemain
             chunk:      1
             tags:       Elastic/Welcome
             subject:    Welcome to Elastic


conf.yaml
```diff
@@ -120,9 +120,10 @@ contents:
         sections:
           - title:      "Welcome to Elastic"
             prefix:     en/welcome-to-elastic
-            current:    main
+            current:    *stackcurrent
             index:      welcome-to-elastic/index.asciidoc
-            branches:   [ {main: master} ]
+            branches:   [ 8.1 ]
+            live:       *stacklivemain
```
gtback, line 126: I'm not sure if this will work, since `stacklivemain` includes branches (7.17) that don't exist for this book. You can leave `live` out entirely and it will default to the full list of `branches`, which I think is fine for now.

goodroot: We can make a 7.17 branch if needed?

goodroot: I made one. :)

debadair:
@elasticmachine retest this please

gtback (approved):

DanRoscigno:
@debadair can this be merged so that we can merge tech-content 12 and 13 and create a live preview?

PR #2394, opened 2022-03-14
Update attributes.asciidoc

goodroot:

### DIFF
--- a/shared/attributes.asciidoc
+++ b/shared/attributes.asciidoc
@@ -182,6 +182,13 @@ Kibana app and UI names
 
 
 //////////
+Enterprise Search
+//////////
+
+:crawler: Enterprise Search Crawler
+:ents: Enterprise Search
+
+//////////
 Ingest terms
 //////////
 


shared/attributes.asciidoc
```diff
@@ -181,6 +181,13 @@ Kibana app and UI names
 :a-data-source:      a data view
 
 
+//////////
+Enterprise Search
+//////////
+
+:crawler: Enterprise Search Crawler
```
chriscressman, line 188:
```suggestion
:crawler: Enterprise Search web crawler
```

chriscressman: We call it "crawler" internally, but the official feature name is "web crawler". And lower case because feature names are lower case.

chriscressman: Looks like I missed that in https://github.com/elastic/enterprise-search-pubs/pull/2168 :grin:

chriscressman (approved):

shared/attributes.asciidoc
goodroot: 😆

richkuz (approved):

PR #2397, opened 2022-03-17
conf.yaml: add apm-agent-go 2.x branch

axw:
Add the apm-agent-go 2.x branch, and mark it as the current branch. v2.0.0 has just been released.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -1273,9 +1273,9 @@ contents:
                 sections:
                   - title:      APM Go Agent
                     prefix:     go
-                    current:    1.x
-                    branches:   [ {main: master}, 1.x, 0.5 ]
-                    live:       [ main, 1.x ]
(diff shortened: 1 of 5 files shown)


gtback (approved):

bmorelli25 (changes requested):
Can you also please update the `shared/versions/stack/*` files. These files map links from the stack-versioned docs to the Go agent. You'll want to update at least the 8.1 file ([link](https://github.com/elastic/docs/blob/0114ab0415056e4da2e212b7c37ac3b1446f4939/shared/versions/stack/8.1.asciidoc#L36)). Updating further back, like 8.0 and 7.17 may also be a good idea if 2.x is compatable with those stack versions.

bmorelli25 (approved):
LGTM

PR #2404, opened 2022-03-24
docs: add `apm-aws-lambda` as a dependency

bmorelli25:
### Summary

Adds the docs directory in [`elastic/apm-aws-lambda`](https://github.com/elastic/apm-aws-lambda) as a dependency to the following repos:

* `elastic/apm-agent-java`
* `elastic/apm-agent-nodejs`
* `elastic/apm-agent-python`

For https://github.com/elastic/observability-docs/issues/1707.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -1321,6 +1321,11 @@ contents:
                         repo:   apm-agent-java
                         path:   CHANGELOG.asciidoc
                         exclude_branches:   [ 0.7, 0.6]
+                      -
+                        repo:   apm-aws-lambda
+                        path:   docs
+                        exclude_branches:   [ 0.7, 0.6 ]
+                        map_branches: *mapMasterToMain
                   - title:      APM .NET Agent
                     prefix:     dotnet
                     current:    1.12
@@ -1354,6 +1359,11 @@ contents:
                         repo:   apm-agent-nodejs
                         path:   CHANGELOG.asciidoc
                         exclude_branches:   [ 1.x, 0.x ]
+                      -
+                        repo:   apm-aws-lambda
+                        path:   docs
+                        exclude_branches:   [ 2.x, 1.x, 0.x ]
+                        map_branches: *mapMasterToMain
                   - title:      APM PHP Agent
                     prefix:     php
                     current:    1.x
@@ -1387,6 +1397,11 @@ contents:
                         repo:   apm-agent-python
                         path:   CHANGELOG.asciidoc
                         exclude_branches:   [ 4.x, 3.x, 2.x, 1.x ]
+                      -
+                        repo:   apm-aws-lambda
+                        path:   docs
+                        exclude_branches:   [ 5.x, 4.x, 3.x, 2.x, 1.x ]
+                        map_branches: *mapMasterToMain
                   - title:      APM Ruby Agent
                     prefix:     ruby
                     current:    4.x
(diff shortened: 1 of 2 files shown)


bmorelli25:
@gtback I might need to pick your brain on this one. As the summary says, we're trying to add one repository as a dependency of the other three. The difficult part is that all four repositories are versioned differently.

conf.yaml
```diff
@@ -1321,6 +1321,11 @@ contents:
                         repo:   apm-agent-java
                         path:   CHANGELOG.asciidoc
                         exclude_branches:   [ 0.7, 0.6]
+                      -
+                        repo:   apm-aws-lambda
+                        path:   docs
+                        exclude_branches:   [ 0.7, 0.6 ]
+                        map_branches: *mapMasterToMain
```
gtback, line 1328:
```suggestion
                        map_branches: 
                          1.x: main
```

Or `8.1` instead of main, etc. The left side is the branch in the "branches" list of this book (and therefore corresponds to the branch apm-agent-java repo, and the right side is the branch in "this" repo (`apm-aws-lambda`)

conf.yaml
```diff
@@ -1354,6 +1359,11 @@ contents:
                         repo:   apm-agent-nodejs
                         path:   CHANGELOG.asciidoc
                         exclude_branches:   [ 1.x, 0.x ]
+                      -
+                        repo:   apm-aws-lambda
+                        path:   docs
+                        exclude_branches:   [ 2.x, 1.x, 0.x ]
+                        map_branches: *mapMasterToMain
```
gtback, line 1366:
```suggestion
                        map_branches: 
                          2.x: main
```

conf.yaml
```diff
@@ -1387,6 +1397,11 @@ contents:
                         repo:   apm-agent-python
                         path:   CHANGELOG.asciidoc
                         exclude_branches:   [ 4.x, 3.x, 2.x, 1.x ]
+                      -
+                        repo:   apm-aws-lambda
+                        path:   docs
+                        exclude_branches:   [ 5.x, 4.x, 3.x, 2.x, 1.x ]
+                        map_branches: *mapMasterToMain
```
gtback, line 1404:
```suggestion
                        map_branches: 
                          6.x: main
```

conf.yaml
gtback: When you add a 2.x branch of `apm-agent-java`, you'll just add a new key/value pair (a new line) to this mapping.

gtback (commented):
> The difficult part is that all four repositories are versioned differently.

This is exactly what `map_branches` is for. The `*mapMasterToMain` and `*mapMainToMaster` are just shortcuts for when that's the only mismatch in branch names (and since all four of the repos use `main`, you won't need either of those).

conf.yaml
bmorelli25: Neat. Thanks for the explanation. The `apm-agent-*` branches will need to be mapped to whatever `current` is. Is there a way to map to `current` so these lines don't need to be updated on a minor release?

bmorelli25: Ohhhh or maybe even a new `*mapToCurrent` attribute? So that only one location needs to be updated?

gtback: You can probably use `stackcurrent` for this. It's defined near the top of `conf.yaml`. This would look like 

```
1.x: *stackcurrent
```

bmorelli25: Nice. Thanks, Greg!

gtback (approved):
LGTM (assuming tests pass and the preview looks good)

gtback:
Build failed for the same reason the main build is failing. If we re-run this tomorrow, it will hopefully pass.

bmorelli25:

PR #2406, opened 2022-03-29
Removes branches to generate for PHP Client

szabosteve:
## Overview

This PR removes 5.x, 2.x, 1.x, 0.4 from the docs branches generated for the PHP Client book from the conf.yml file. These branches no longer needed to be in the docs.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -567,7 +567,7 @@ contents:
               - title:      PHP Client
                 prefix:     php-api
                 current:    *stackcurrent
-                branches:   [ master, 8.1, 8.0, 7.17, 7.16, 7.x, 6.x, 5.x, 2.x, 1.x, 0.4 ]
+                branches:   [ master, 8.1, 8.0, 7.17, 7.16, 7.x, 6.x ]
                 live:       *stacklivemain
                 index:      docs/index.asciidoc
                 chunk:      1


gtback (commented):
Is there a reason we're removing these docs? Do they no longer build cleanly? Require too much maintenance/backporting? 

I'm actually not sure that removing these branches from here will remove the files at the next full rebuild or not; we might have to clear them out manually from built-docs. 

Should we set up redirects (for example, to the latest version, or to 6.x)?

ezimuel:
@gtback there is not a real need here. I'm just cleaning up someold branches in the repository. If you think this will have an impace on all the docs maybe with some broken links we can leave as is, no problem.

gtback:
@ezimuel It doesn't seem like it's going to break any links within the docs, but in case there are other links to these docs from elsewhere on the Internet, we generally try to either leave the old versions up indefinitely or at least set up a redirect so that people don't get a 404 if they follow those links. In the latter case, we can set up a blanket redirect so that any page under the old versions gets redirected to (for example) https://www.elastic.co/guide/en/elasticsearch/client/php-api/current/overview.html . If you decide to go with the latter, I can help get the correct redirects in place.

szabosteve:
After checking it with @ezimuel, we leave the branches as they are now, and I close this PR. Thanks for the insights @gtback!

ezimuel:
@gtback thanks for the feedback. I think it's more safe to leave the docs as they are now.

PR #2432, opened 2022-05-06
add simple docker-compose.yml for simpler air-gapped docs starting

face0b1101:
Air-gapped docs are really useful, but it would be nicer to be able to run `docker-compose up -d` and `docker-compose down`.


face0b1101:
forgot to add README.md

PR #2433, opened 2022-05-06
Air gapped docker compose

face0b1101:
Air-gapped docs are really useful, but it would be nicer to be able to run `docker-compose up -d` and `docker-compose down`.
Also added a README with step-by-step instructions for both vanilla docker and docker-compose.

### DIFF
--- a/air_gapped/README.md
+++ b/air_gapped/README.md
@@ -1,0 +1,67 @@
+# Air-Gapped Elastic Documentation #
+
+## Build the container ##
+
+1. Go [here](https://github.com/elastic/docs). Read the docs if desired.
+2. Clone the repo:
+
+    ```bash
+    git clone https://github.com/elastic/docs.git
+    ```
+
+3. navigate to air_gapped:
+
+    ```bash
+    cd docs/air_gapped
+    ```
+
+4. Build the latest container:
+
+    ```bash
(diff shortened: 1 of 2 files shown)


jonny-mcc:
Gave the instructions in your README a try but found it didn't work for me. I just get the error
`~/.git-references/built-docs.git must exist and contain a reference clone of the built-docs repo`
when I try and run the build.sh script. This is caused by the if statement within the script. Is this really needed seeing as the next line clones the repository that it complains is missing? Unless I missing something obvious here

gtback:
@jonny-mcc the current `air_gapped/build.sh` script (which is being leveraged for local builds in this PR) is designed to run in CI, and given the size of the built-docs repo we want to cache as much as we can on the build image rather than cloning the entire repo on every job. If that fails for any reason we'd rather it fail (to let us know) rather than spending lots of time repeatedly cloning the repo. Not saying it's a _good_ reason, but it is _the_ reason. 😁 

You can probably set up a local reference repo of `built-docs` if you want to use this to build the air-gapped docs locally.

That said, you shouldn't need to build the container if you can instead just download and run the "air-gapped" container image from our (private, internal) Docker registry. 

Let me know if that helps, or if you have any other questions.

PR #2446, opened 2022-06-07
elastic/cloud-on-k8s move master -> main

naemono:
Needs to be merged after https://github.com/elastic/cloud-on-k8s/pull/5740.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -951,7 +951,7 @@ contents:
             tags:       Kubernetes/Reference
             subject:    ECK
             current:    2.2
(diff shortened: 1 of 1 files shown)


lcawl:
I think it might be too soon for this change. I think we were unable to change the URL to "main" in our docs until *all* books used main, otherwise there'd be linking issues. @gtback ?

naemono:
@lcawl Honestly, the order of these changes is a bit confusing to me as well.  Let me know if I need to do something different here.  Thanks!

gtback:
We're not ready to switch all the URLs from `master` to `main`. I'd like to do it all at once for simplicity's sake; it's probably possible to do them piecemeal, but I haven't fully thought through the implications. There are SEO considerations (putting this on your radar, @SEOEBQ !), some redirects we'll need to set up, and some improvements to the build system that will be needed.

I've also not been prioritizing this work, so thanks for the reminder that it still needs to get done!

naemono:
Understood.  Closing.

PR #2450, opened 2022-06-17
Add shared attribute for Elasticsearch Service Console

lcawl:
This attribute is defined in index.asciidoc files in the cloud repo, but those are not available to other content. Therefore, this PR adds it to the list of shared attributes.

It can then be used in PRs like https://github.com/elastic/kibana/pull/134552


lcawl:

PR #2452, opened 2022-06-20
docs: update versions/stack rum agent

devcorpio:
This makes stack links from master, and 8.x point to the 5.x RUM agent documentation. 

Thank you @yakhinvadim for bringing this to our notice!

Example of docs pointing to 4.x:  https://www.elastic.co/guide/en/apm/guide/current/apm-rum.html

**Question:** 

Does the current documentation will start pointing to the new link automatically once this is merged? Or do I need to do something else? Many thanks!!!

### DIFF
--- a/shared/versions/stack/8.0.asciidoc
+++ b/shared/versions/stack/8.0.asciidoc
@@ -36,7 +36,7 @@ APM Agent versions
 :apm-go-branch:         2.x
 :apm-ios-branch:        0.x
 :apm-java-branch:       1.x
-:apm-rum-branch:        4.x
+:apm-rum-branch:        5.x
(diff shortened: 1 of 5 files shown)


bmorelli25 (approved):
Thank you @devcorpio!

> Does the current documentation will start pointing to the new link automatically once this is merged? Or do I need to do something else? Many thanks!!!

I'm pretty sure a change like this is cached and does not update automatically with our regular doc builds. To be safe, I'll kick off a full rebuild of our documentation website. This will rebuild the docs from scratch and ensure this update occurs. It may not be necessary, but it only takes 30 mins to run, so we might as well do it to be safe 🚢

PR #2463, opened 2022-06-23
Remove excess margin and padding on guide index

bmorelli25:
### Summary

I feel like there's a lot of empty space in the custom header on [elastic.co/guide](https://www.elastic.co/guide/index.html). This PR removes excess margin and padding to tighten things up a bit. I really wanted to figure out to get this to work for the rest of the elements on the page, but unfortunately, it appears that the doc build treats the TOC page the same as every other documentation page, so the only way I could specifically target the TOC H3s was by using jQuery.

Here's a screenshot of this change (or see the [preview](https://docs_2463.docs-preview.app.elstc.co/guide/index.html))

Screen Shot 2022-06-23 at 2 54 41 PM

### DIFF
--- a/extra/docs_landing.html
+++ b/extra/docs_landing.html
@@ -90,7 +90,7 @@
   </table>
 
   <h3 style="margin-top: 0px !important">Need help?</h3>
-  <table width="100%">
+  <table width="100%;margin-bottom: 5px !important;">
     <tr>
       <td style="width: 50%; padding-top: 0px !important;">
         <div class="itemizedlist">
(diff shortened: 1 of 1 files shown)


ollyhowell (approved):
Great work! Looks so much cleaner. Thanks for volunteering for this!

gtback (approved):
I agree. The new version (on the right) looks a lot cleaner than the current version (on the left). I also tried at a few different window sizes. 

image


I'd normally shy away from using inline CSS, but since we *don't* want to impact the rest of the site, and the changes are pretty well contained to this page, I'm OK with this approach!

gtback:
I'm hoping this just needed the change from #2464 to get the build to pass. I merged it in.

PR #2467, opened 2022-06-28
Add shared attribute to flag features that aren't available yet

debadair:


kilfoyle (approved):
Looks good! 🚢 

I was just about to type that we should use this attribute inside of an admonition, and then I saw your very same comment in Slack. 

So just for the record, we'll plan to use this as `IMPORTANT: {es-feature-flag}`

Screen Shot 2022-06-28 at 12 03 15 PM

PR #2478, opened 2022-07-18
`elastic.co/guide/*` redesign

bmorelli25:
[**Click here to view the interactive preview**](https://docs_2478.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/current/elasticsearch-intro.html)

---

### Summary

This PR updates the style and layout of our `elastic.co/guide/*` documentation.

**Page**
* Removes margin and widens the container using a new container class to accommodate sticky content on both sides.

**Table of contents (left column)**
* Moves the table of contents to the left side of the page and makes it sticky.
* Scrolls the TOC to the selected element.
* Styles the table of contents to be more accessible with higher contrast between the text and the background (the blue on gray was difficult to read).
* Adds SVG chevron icons to the table of contents—great for screens of any resolution.

**On this page & marketing (right column)**
* Makes the "On this page" content sticky.
* Adds an `IntersectionObserver()` to highlight one item at a time in the "On this page" list.
* Simplifies the styling of the "Most popular" and "We're hiring" boxes and makes them sticky.

### Why?

Our documentation is difficult to navigate, and readers constantly have to scroll up and down to maintain context or navigate to new pages. There are a number of reasons why this happens:

* The table of contents is hidden below the fold.
* The "On this page" widget disappears when you scroll down the page.
* Many elements, including the page container have margins that waste too much screen real estate.

### What does it look like?

[**Click here to view the interactive preview**](https://docs_2478.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/current/elasticsearch-intro.html)

Screen Shot 2022-07-31 at 7 04 10 PM

### Known limitations

* If no section heading is visible on a page, highlighting is removed from the OTP. [Example](https://docs_2478.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/current/how-watcher-works.html). 
* The OTP does not scroll with the content on longer pages. [Example](https://docs_2478.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/current/eql-syntax.html).

### Stakeholders

- [x] The documentation team (see Aug. 3rd cross-doc recording)
- [x] Docs infra (@gtback)
- [x] Docs PM (@KOTungseth)
- [x] Front-end (@chandlerprall or @glitteringkatie)
- [x] Marketing / web (see linked internal issue)
- [x] QA (@rashmivkulkarni) 
Others?

### QA

Sorry, this is an internal link. [Bug history and QA checklist](https://docs.google.com/document/d/1craSp_6CA2rk-ZJDTLEONfUTE6MQ-9IBJ5_1Kf5iKh4/edit).

### DIFF
--- a/resources/web/docs_js/index.js
+++ b/resources/web/docs_js/index.js
@@ -342,10 +342,15 @@ $(function() {
   if($(window).width() >= 769){
     var scrollToSelectedTOC = setInterval(() => {
       if ($('.current_page').length) {
-        $('.current_page')[0].scrollIntoView({
-          behavior: "auto",
-          block: "center"
-        });
+          // Get scrollable element
+          var container = document.querySelector("#left_col");
+          // Get active table of contents element
+          var activeItem = document.querySelector(".current_page")
+          // If the top of the active item is out of view
+          if (container.offsetHeight - 100 <= activeItem.offsetTop) {
+            // Scroll to active item
+            container.scrollTop = activeItem.offsetTop
+          }
         clearInterval(scrollToSelectedTOC);
       }
     }, 150);
--- a/resources/web/docs_js/index.js
+++ b/resources/web/docs_js/index.js
@@ -194,7 +194,6 @@ function highlight_otp() {
   })
 
   document.querySelectorAll('#guide a[id]').forEach((heading) => {
-    console.log("heading", heading);
     observer.observe(heading);
   })
 }
--- a/resources/web/docs_js/index.js
+++ b/resources/web/docs_js/index.js
@@ -337,6 +337,18 @@ $(function() {
     $('a.edit_me_private').show();
   }
 
+  // scroll to selected TOC element
+  // if it doesn't exist yet, wait and try again
+  var scrollToSelectedTOC = setInterval(() => {
+    if ($('.current_page').length) {
+      $('.current_page')[0].scrollIntoView({
+        behavior: "auto",
+        block: "center"
+      });
+      clearInterval(scrollToSelectedTOC);
+    }
+ }, 150);
+
   // Test comment used to detect unminifed JS in tests
 });
 
--- a/resources/web/docs_js/index.js
+++ b/resources/web/docs_js/index.js
@@ -250,7 +250,6 @@ $(function() {
   AlternativeSwitcher(store());
 
   var sticky_content = $('#sticky_content'); // Move rtp container to top right and make visible
-  var elastic_footer = $('#elastic-footer');
   var left_col = $('#left_col'); // Move OTP to top left
 
   $('.page_header > a[href="../current/index.html"]').click(function() {
--- a/resources/web/docs_js/index.js
+++ b/resources/web/docs_js/index.js
@@ -19,6 +19,9 @@ import "../../../../../node_modules/details-polyfill";
 // Add support for URLSearchParams Web API in IE
 import "../../../../../node_modules/url-search-params-polyfill";
 
+// Vocab:
+// TOC = table of contents
+// OTP = on this page
 export function init_headers(sticky_content, lang_strings) {
   // Add on-this-page block
   var this_page = $('<div id="this_page"></div>').prependTo(sticky_content);
@@ -170,6 +173,10 @@ function init_toc(lang_strings) {
     });
 }
 
+// In the OTP, highlight the heading of the section that is
+// currently visible on the page.
+// If more than one is visible, highlight the heading for the
+// section that is higher on the page.
 function highlight_otp() {
   let visibileHeadings = []
   const observer = new IntersectionObserver(entries => {
@@ -257,11 +264,16 @@ $(function() {
 
   AlternativeSwitcher(store());
 
-  var sticky_content = $('#sticky_content'); // Move rtp container to top right and make visible
-  var left_col = $('#left_col'); // Move OTP to top left
+  // Move rtp container to top right and make visible
+  var sticky_content = $('#sticky_content');
+  // Left column that contains the TOC
+  var left_col = $('#left_col'); 
+  // Middle column that contains the main content
   var middle_col = $('#middle_col');
-  var right_col = $('#right_col');
-  var bottom_left_col = $('#bottom_left_col')
+  // Right column that contains the OTP and demand gen content
+  var right_col = $('#right_col'); 
+  // Empty column below TOC on small screens so the demand gen content can be positioned under the main content
+  var bottom_left_col = $('#bottom_left_col');
 
   $('.page_header > a[href="../current/index.html"]').click(function() {
     utils.get_current_page_in_version('current');
@@ -313,10 +325,15 @@ $(function() {
     });
   } else {
     init_toc(LangStrings);
-    left_col.removeClass().addClass('col-0')
-    bottom_left_col.removeClass().addClass('col-0')
-    middle_col.removeClass().addClass('col-12 col-lg-9 guide-section')
-    right_col.removeClass().addClass('col-12 col-lg-3 sticky-top-md h-almost-full-lg')
+    // Style book landing page (no main content, just a TOC and demand gen content)
+
+    // Set the width of the left column to zero
+    left_col.removeClass().addClass('col-0');
+    bottom_left_col.removeClass().addClass('col-0');
+    // Set the width of the middle column (containing the TOC) to 9
+    middle_col.removeClass().addClass('col-12 col-lg-9 guide-section');
+    // Set the width of the demand gen content to 3
+    right_col.removeClass().addClass('col-12 col-lg-3 sticky-top-md h-almost-full-lg');
   }
 
   PR.prettyPrint();
@@ -346,7 +363,8 @@ $(function() {
           var container = document.querySelector("#left_col");
           // Get active table of contents element
           var activeItem = document.querySelector(".current_page")
-          // If the top of the active item is out of view
+          // If the top of the active item is out of view (or in the bottom 100px of the visible portion of the TOC)
+          // scroll so the top of the active item is at the top of the visible portion TOC
           if (container.offsetHeight - 100 <= activeItem.offsetTop) {
             // Scroll to active item
             container.scrollTop = activeItem.offsetTop
--- a/resources/web/docs_js/index.js
+++ b/resources/web/docs_js/index.js
@@ -259,6 +259,8 @@ $(function() {
 
   var sticky_content = $('#sticky_content'); // Move rtp container to top right and make visible
   var left_col = $('#left_col'); // Move OTP to top left
+  var middle_col = $('#middle_col');
+  var right_col = $('#right_col');
 
   $('.page_header > a[href="../current/index.html"]').click(function() {
     utils.get_current_page_in_version('current');
@@ -310,6 +312,9 @@ $(function() {
     });
   } else {
     init_toc(LangStrings);
+    left_col.removeClass().addClass('col-0')
+    middle_col.removeClass().addClass('col-12 col-md-9 guide-section')
+    right_col.removeClass().addClass('col-12 col-md-3 sticky-top h-almost-full')
   }
 
   PR.prettyPrint();
--- a/resources/web/docs_js/index.js
+++ b/resources/web/docs_js/index.js
@@ -22,7 +22,7 @@ import "../../../../../node_modules/url-search-params-polyfill";
 export function init_headers(sticky_content, lang_strings) {
   // Add on-this-page block
   var this_page = $('<div id="this_page"></div>').prependTo(sticky_content);
-  this_page.append('<p id="otp">' + lang_strings('On this page') + '</p>');
+  this_page.append('<p id="otp" class="aside-heading">' + lang_strings('On this page') + '</p>');
   var ul = $('<ul></ul>').appendTo(this_page);
   var items = 0;
   var baseHeadingLevel = 0;
@@ -57,7 +57,7 @@ export function init_headers(sticky_content, lang_strings) {
             .remove();
           var text = title_container.html();
           const adjustedLevel = hLevel - baseHeadingLevel;
-          const li = '<li id="otp-text" class="heading-level-' + adjustedLevel + '"><a href="#' + this.id + '">' + text + '</a></li>';
+          const li = '<li id="otp-text-' + i + '" class="otp-text heading-level-' + adjustedLevel + '"><a href="#' + this.id + '">' + text + '</a></li>';
           ul.append(li);
         }
       }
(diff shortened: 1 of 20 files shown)


dedemorton:
I love where this is heading. One thing that we should change when we have time: the font in right-side TOC is too small relative to other text. I wonder if we should consider increasing the width of the page so that we do not have so much whitespace, and then we could increase the font size.

bmorelli25:
> I wonder if we should consider increasing the width of the page so that we do not have so much whitespace, and then we could increase the font size.

Agreed. Our docs use Boostrap v4, which defines XL screens as anything above 1200 pixels. The screenshots above were taken on my laptop with a width of ~1400 pixels.

There's a ton of whitespace in my screenshot above because we use the `container` class which sets our max-width to 1140px:

https://github.com/elastic/docs/blob/0b1e761e2f9424f7acaad147887d888b0751ed80/resources/web/lib/bootstrap/bootstrap.css#L615-L619

I haven't done a lot of research, but it sounds like we can use the `container-fluid` class instead to remove that limitation.

> One thing that we should change when we have time: the font in right-side TOC is too small relative to other text.

Yeah, good point. That was me. We can definitely make it larger.

resources/web/style/base_styles.pcss
```diff
@@ -30,4 +30,67 @@
...
+  .content-container {
+    width: 100%;
+    padding-right: 20px;
+    padding-left: 20px;
+    margin-right: auto;
+    margin-left: auto;
+    @media screen and (min-width: 1560px) {
+      max-width: 1500px;
+    }
+  }
```
bmorelli25, line 43: Is there an advantage to using this over Bootstrap's built-in `container-fluid` class? The main difference looks to be the media query—does it make more sense to use the Bootstrap class and just add the query to this style sheet?

https://github.com/elastic/docs/blob/4110e2b593906d7ed7c1e1c8c93c9f2d8317d8a1/resources/web/lib/bootstrap/bootstrap.css#L621-L627

resources/web/style/link.pcss
```diff
@@ -1,6 +1,6 @@
 #guide {
   a {
-    color: #0077CC;
+    color: #343741;
```
colleenmcginnis, line 3: I think we should use a different color for links outside of the table of contents. It's really difficult to determine if text is a link in the main content especially without `text-decoration`. I think I changed it from `#00a9e5` to `#0077CC` because it was bothering me that it was a slightly different shade of blue than than the color that's used in the buttons in the top nav. 😛 

For example there's no way to know all the values in the left column are links until you hover over them (I'm hovering over `cloudfront` in the screenshot below):

image

resources/web/style/base_styles.pcss
colleenmcginnis: I didn't know about `container-fluid`! Feel free to make the switch. 🙂

resources/web/style/link.pcss
bmorelli25: Oops! I didn't realize I changed these links too.

resources/web/style/on_this_page.pcss
```diff
@@ -5,16 +5,19 @@
 
   .heading-level-1 {
     display: block;
-    padding-left: 1.5em !important;
+    padding-left: 1em !important;
+    a {
+      font-size: 0.9em;
+    }
```
bmorelli25, line 11: Should we remove this? Only `.heading-level-1` has a unique font-size—`*0`, `*2`, and `*3` do not. The OTP looks a little goofy when there are multiple heading levels. I'm not sure why I added this in the first place 🙃.

bmorelli25:
```suggestion
```

leemthompo:
This is a great initiative! What do we thing of making subheadings more visually distinct, i.e. render `h2`, `h3`, etc. more easily distinguishable at a glance? This, plus the work you've already started here would solve 99% of my issues with the current docs reading UX. 🙂

bmorelli25:
> What do we think of making subheadings more visually distinct, i.e. render h2, h3, etc. more easily distinguishable at a glance?

@colleenmcginnis recently brought up the topic of headings. We discussed it, but it might be beyond the scope of this PR. Changing the headings is a bit complex—mostly because asciidoc converts headings based on the heading level in the file. That means some pages start with an H2, and others start with an H4. We could update the styles of heading level, but pages would still differ due to how they are built.

resources/web/template.html
```diff
@@ -82,22 +82,48 @@
...
-                        <li class="icon-logstash-white"><a href="https://www.elastic.co/webinars/introduction-elk-stack?baymax=default&elektra=docs&storm=top-video">ELK for Logs & Metrics: Video</a></li>
-                      </ul>
+
+                <div class="col-12 order-3 col-lg-2 order-lg-3 h-almost-full-lg sticky-top-lg" id="right_col">
+                  <div id="sticky_content">
+                    <!-- The OTP is appended here -->
+                    <div class="row">
+                      <div class="col-0 col-md-4 col-lg-0" id="bottom_left_col"></div>
+                      <div class="col-12 col-md-8 col-lg-12">
+                        <div id="rtpcontainer">
```
gtback, line 103: These changes will only affect the style for people who don't load content from Optimizely (I think?) I'd be curious to see how actual changes from Optimizely or the recommender look.

gtback (approved):
I don't feel particularly qualified to review the CSS or JavaScript changes, but I'm 👍🏻 for these changes as long as others approve. The pages look a lot better to me!

resources/web/template.html
bmorelli25: That's correct. We still need to work with marketing to figure out how to test the Optimizely changes. Staging sounds like it could be our best bet.

colleenmcginnis: A very detailed issue coming soon in /website-requests! 💁‍♀️

colleenmcginnis: ✨  https://github.com/elastic/website-requests/issues/678

bmorelli25: > I don't feel particularly qualified to review the CSS or JavaScript changes

I'd love to find a front-end engineer to review this. At a bare minimum, I'm confident there's a better way to scroll the TOC than what I implemented in 936464bcffb42bb4f17809dc919f95665ba4a0b2.

gtback:
> I'd love to find a front-end engineer to review this.

@glitteringkatie @chandlerprall would you have some time to review this, or nominate someone else who might be able to help?

alaudazzi:
Thank you for leading this initiative! It looks much better than what we have today :-)
QQ: If I search for a term, the results are shown in the old system. Is it because the new system is still on staging?

colleenmcginnis:
>If I search for a term, the results are shown in the old system. Is it because the new system is still on staging?

@alaudazzi By "search for a term" do you mean use the site-wide search and click on a result (see screenshot below)? 

It looks like those links go to `https://elastic.co/*` pages from the staging site. (My guess is that we don't have a staging search set up.) All production pages at `https://elastic.co/*` URLs use the old layout. Only pages on the staging site show the new layout. 

Screen Shot 2022-08-04 at 8 53 24 AM

resources/web/template.html
chandlerprall: That scrolling logic looks good to me! jQuery used to have a way to wait for an element selector to resolve, but doesn't look like that's still the case.

resources/web/docs_js/index.js
```diff
@@ -170,6 +170,34 @@ function init_toc(lang_strings) {
...
+      let element = document.querySelector(`#sticky_content #this_page a[href="#${id}"]`);
+      let itemId = $(element).parent().attr('id')
+      if (entry.intersectionRatio > 0){
+        visibileHeadings.push(itemId);
+      } else {
+        const position = visibileHeadings.indexOf(itemId);
+        visibileHeadings.splice(position, position + 1)
+      }
+      if (visibileHeadings.length > 0) {
+        visibileHeadings.sort()
```
chandlerprall, line 187: [I'm assuming ids are numerical]

At this point the `itemId`s will be strings and the sort will start failing when the id hits double digits,

```
const arr = [];
for (let i = 0; i < 11; i++) arr.push(`${i}`)
arr.sort();
// ['0', '1', '10', '2', '3', '4', '5', '6', '7', '8', '9']
```

need to parse them to numbers first,

```
let itemId = parseInt($(element).parent().attr('id'), 10)
```

resources/web/template.html
bmorelli25: Ah, great! Thank you so much for taking a look.

resources/web/docs_js/index.js
```diff
@@ -170,6 +170,34 @@ function init_toc(lang_strings) {
...
+  const observer = new IntersectionObserver(entries => {
+    entries.forEach(entry => {
+      const id = entry.target.getAttribute('id');
+      let element = document.querySelector(`#sticky_content #this_page a[href="#${id}"]`);
+      let itemId = $(element).parent().attr('id')
+      if (entry.intersectionRatio > 0){
+        visibileHeadings.push(itemId);
+      } else {
+        const position = visibileHeadings.indexOf(itemId);
+        visibileHeadings.splice(position, position + 1)
```
chandlerprall, line 184:
```suggestion
        visibileHeadings.splice(position, 1)
```

The second argument to `splice` is how many elements to delete. This code is working because, as the heading scrolls off the page any following elements in the array will also be off the page, so removing `position + 1` elements overlaps conceptually. Still, better to remove just the intended element.

chandlerprall (changes requested):

resources/web/docs_js/index.js
colleenmcginnis: Ahhh thanks @chandlerprall! 

The `itemId`s are strings (for example `otp-text-1`, `otp-text-2`).  

https://github.com/elastic/docs/blob/99eab4ec17a3f9248b2a30b31d242a7575e94c6d/resources/web/docs_js/index.js#L60

I don't _think_ we want to use just a numbered `id` for list items without the context of `otp-text-` (but maybe it would be ok?).

colleenmcginnis: This seems to work, but there is probably a more elegant solution. 😝 

```js
    entries.forEach(entry => {
      const id = entry.target.getAttribute('id');
      let element = document.querySelector(`#sticky_content #this_page a[href="#${id}"]`);
      let itemId = $(element).parent().attr('id')
      let itemNumber = parseInt(itemId.match(/otp-text-(\d*)/)[1], 10)
      if (!itemNumber) return
      if (entry.intersectionRatio > 0){
        visibileHeadings.push(itemNumber);
      } else {
        const position = visibileHeadings.indexOf(itemNumber);
        visibileHeadings.splice(position, 1)
      }
      if (visibileHeadings.length > 0) {
        visibileHeadings.sort((a, b) => a - b)
        // Remove existing active classes
        $('a.active').removeClass("active");
        // Add active class to the first visible heading
        $('#otp-text-' + visibileHeadings[0] + ' > a').addClass('active')
      }
    })
```

chandlerprall: Ah, that's still workable; it can parse out the numeric id with a regex instead,

```
const itemId = parseInt(
  $(element).parent().attr('id').match(/\d+/)[0],
  10
);
```

If that creates too much work in other places that wants to use the full string id, the call to `.sort()` can do the parsing work. A bit more CPU overhead, but as long as we don't have pages with 100s of headings it shouldn't be noticable,

```
visibleHeadings.sort((a, b) => {
  const aId = parseInt(
    a.match(/\d+/)[0],
    10
  );
  const bId = parseInt(
    a.match(/\d+/)[0],
    10
  );
  if (aId < bId) return -1;
  if (aId > bId) return 1;
  return 0;
});
```

chandlerprall (approved):
LGTM 👍

KOTungseth (approved):
LGTM👍

dedemorton:
I found some places where the page is scrolled instead of starting at the top like it should.

I'm on Chrome Version 103.0.5060.134 (Official Build) (x86_64)

https://user-images.githubusercontent.com/14206422/183477608-3ff823aa-5f94-446d-923e-9d24b51293ab.mp4

rashmivkulkarni (approved):
Checked for spacing, paragraph breaking, proper rendering, spot checks on various links etc.. LGTM.  ( some of my nits are addressed in the PR.)

colleenmcginnis:
☝️ @bmorelli25 that commit switches from `scrollIntoView` to `scrollTop`, which adds a little more complexity, but some [light internet research](https://stackoverflow.com/questions/11039885/scrollintoview-causing-the-whole-page-to-move) suggests it might be necessary to prevent the full page from scrolling.

lcawl:
IMO the initial list of pages and sections looks plain and it's not obvious you can click them until you get a mouse over them. For example:

image

However, the benefits once you start clicking outweigh this issue, so I think it's worth merging as-is.

resources/web/docs_js/index.js
```diff
@@ -299,6 +337,25 @@ $(function() {
...
+  // window.width must match the breakpoint of `.sticky-top-md`
+  if($(window).width() >= 769){
+    var scrollToSelectedTOC = setInterval(() => {
+      if ($('.current_page').length) {
+          // Get scrollable element
+          var container = document.querySelector("#left_col");
+          // Get active table of contents element
+          var activeItem = document.querySelector(".current_page")
+          // If the top of the active item is out of view
+          if (container.offsetHeight - 100 <= activeItem.offsetTop) {
```
bmorelli25, line 350: Where does the `100` come from?

colleenmcginnis: The value `100` is somewhat arbitrary. It specifies that if an active item is in the bottom 100px of the visible portion of the table of contents, we should scroll to avoid the active item being _barely_ visible like this (where the active item, "Migration guide" is technically within view but barely):

image

dedemorton:
>IMO the initial list of pages and sections looks plain and it's not obvious you can click them until you get a mouse over them.

@lcawl TBH for most "books" I wish we could just skip the initial list and go directly to content. In most cases, the list page does not add value. I think there are _some_ books that have introductory text on this page in addition to the list.

Maybe as a next step we could try to beef up the introductory text for each book? Just a thought, not a hard requirement.

richkuz:
I love the new redesign, really great grassroots effort to improve our docs 💯 😍 

Now that the site is live, a lot of the feedback I've heard is very positive. I heard some constructive feedback, too. Listing it here (if there's another place to put this, please let me know):
1. It's new/different, which requires adjustment.
2. It's not obvious that the left-hand column can be scrolled up and down.
3. _Maybe_ consider making the version dropdown list sticky, so that when you scroll down it stays at the top.
4. If the announcement in the right-hand nav gets too large, it could make it hard to see the right-hand navigation content. The announcement today ("We're hiring!") is small, with no impact on the right-hand navigation, so the problem isn't apparent today.

Nothing super actionable, just sharing.

PR #2479, opened 2022-07-19
Update og:image meta tag

### LINKED ISSUE #2428: Fix `og:image` in template
This URL doesn't exist anymore:

https://github.com/elastic/docs/blob/e9568b3778e5f095618592f207edd9a9d550282a/resources/web/template.html#L32

gtback:
The previous image is a 404, so this puts an image on Twitter, Facebook,
etc.

Fixes #2428

### DIFF
--- a/resources/web/template.html
+++ b/resources/web/template.html
@@ -29,7 +29,7 @@
     <meta name="yandex-verification" content="d8a47e95d0972434" />
     <meta name="localized" content="true" />
     <meta name="st:robots" content="follow,index" />
-    <meta property="og:image" content="https://www.elastic.co/static/images/elastic-logo-200.png" />
+    <meta property="og:image" content="https://static-www.elastic.co/v3/assets/bltefdd0b53724fa2ce/blt280217a63b82a734/6202d3378b1f312528798412/elastic-logo.svg" />
     <link rel="shortcut icon" href="/favicon.ico" type="image/x-icon">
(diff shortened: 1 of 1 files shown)


bmorelli25 (commented):
I attempted to test this PR with the [Meta sharing debugger](https://developers.facebook.com/tools/debug/). I see this warning:

Screen Shot 2022-07-18 at 6 53 24 PM

Do we want to specify a default height and width? [Meta for developers](https://developers.facebook.com/docs/sharing/best-practices/#precaching) says this:

> Use `og:image:width` and `og:image:height` Open Graph tags to specify the image dimensions to the crawler so that it can render the image immediately without having to asynchronously download and process it.

Testing with Twitter's dev tools produces a similar warning.

gtback:
Thanks, @bmorelli25 . Yep, I probably should. I wonder if the fact that it's an SVG image has any impact.

gtback:
@bmorelli25 once the build is finished, would you be able to test this again? I don't have a Facebook account to test with.

bmorelli25 (approved):
Approving as it looks like this problem has been fixed. I now get other warnings when testing though:

gtback:
Thanks, @bmorelli25. I'll merge this to fix the immediate issue (the image not existing) and create a separate issue (TBD when anyone will get to it) for the other fields; a lot of those will need to be templated per-page rather than hard-coding in the template.

PR #2486, opened 2022-08-04
ESS - Change current to ms-78

kilfoyle:
This changes "current" for the Cloud ESS docs to ms-78 and should be merged on ms-78 release day.


abdonpijpelink (approved):
LGTM

PR #2488, opened 2022-08-09
[DOCS] Add shared attributes for Webhook connectors

lcawl:
Relates to https://github.com/elastic/kibana/pull/137726

This PR adds shared attributes for the "Webhook" and "Webhook - Case Management" connector names.


szabosteve (approved):
LGTM!

PR #2491, opened 2022-08-11
Add attributes for various client library URLs

gtback:
This will help referring to these books from other documentation


KOTungseth (approved):

karenzone (approved):
We <3 attributes. Thank you for adding these.

PR #2501, opened 2022-08-22
Add `infrastructure-app` shared attribute

colleenmcginnis:
Ref https://github.com/elastic/observability-docs/issues/2097, https://github.com/elastic/observability-docs/pull/2102

In 8.4 the high-level "Metrics" section of the Observability app will be changed to "Infrastructure". This PR adds an `infrastructure-app` shared attribute. I don't _think_ we want to replace the value of the `metrics-app` attribute since it will continue to be needed in pre-8.4 content, but going forward we should use `infrastructure-app` when referring to this section.


dedemorton (approved):
LGTM

PR #2522, opened 2022-09-22
Create a home for our graphics source files

kilfoyle:
When diagrams in our docs need to be updated we often need to hunt around to find the associated source file from which a graphic is generated. Having a common home for all of these source files could make that process a lot easier, and it means we don't have a dependency on anyone having access to specific team spaces in Figma.

@lcawl @gtback Do you think this would be the right place to store these source files? @karenzone I think you had asked for this wrt the ingest diagram so any suggestions are most welcome.


gtback:
@kilfoyle I would lean towards keeping the source files in the same directory as the images they're being used to generate. I don't think they'll get copied into `built-docs` unless the .fig files are directly referenced, but even if they are, that's not the worst thing. For images that are only in one book, they'd just be in the content source repo, but for images that are shared between lots of books, somewhere like https://github.com/elastic/docs/tree/master/shared/images might be an OK place.

Is there a compelling reason I'm not thinking of to keep the source files in this repo even if the images themselves are in a different repo?

kilfoyle:
Thanks @gtback! I was thinking of using the docs repo because I didn't want to expand the `built-docs` download, but I gather that's not an issue and also I notice that the source files are smaller than I expected.

I'll close this PR and will store the source files in the repos where the output .svg files are used.

gtback:
> didn't want to expand the built-docs download, but I gather that's not an issue

Yeah, for now, it would just be a drop in an enormous bucket, even if they did get copied (which I don't think they will).

PR #2523, opened 2022-09-23
Add stylesheet for generated APIs

lcawl:
This PR is a first draft of a new stylesheet related to auto-generated API output

Relates to https://github.com/elastic/kibana/issues/137240

### DIFF
--- a/resources/web/style/openapi.pcss
+++ b/resources/web/style/openapi.pcss
@@ -1,4 +1,60 @@
 #guide {
+  /*
+ * The following styles were defined in the generator but are not used here
+ *
+ *  body {
+ *	font-family: Trebuchet MS, sans-serif;
+ *	font-size: 15px;
+ *	color: #444;
+ *	margin-right: 24px;
+ * }
+ * 
+ * h1	{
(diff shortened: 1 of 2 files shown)


resources/web/style/openapi.pcss
```diff
@@ -0,0 +1,182 @@
+#guide {
```
gtback, line 1: I think we should wrap all of this in a `<div class="openapi">` (which would require modifying the HTML template a bit), and nest everything in a `.openapi` block in the PCSS, to reduce the chances that these styles will conflict or override styles defined for the rest of the page.

gtback: Nevermind, I see you literally just added this 10 seconds before I made my comment 🤣

gtback (approved):
I just pushed a commit that formatted the PCSS a bit more consistently. 

LGTM!

resources/web/style/openapi.pcss
lcawl: I woke up with that realization this morning, glad we're on the same wavelength!

PR #2527, opened 2022-09-26
Add attributes for versioned welcome book

lcawl:
This PR adds versioning to the shared attributes for the Welcome guide, since it will soon have multiple versions.


kilfoyle (approved):
LGTM! 🌵

PR #2534, opened 2022-09-28
Adds machine learning shortcut to Kibana link checker

szabosteve:
## Overview

This PR adds the ML URL shortcut to the KIbana link check in `build_docs.pl`.


lcawl (approved):
LGTM and I think it's okay to merge before https://github.com/elastic/kibana/pull/142091

gtback:
In the past we've added tests for each new shortcut, but there are so many of these now, and they follow the same pattern, that I'm not worried about it.

PR #2535, opened 2022-09-30
Add APM extension books

bmorelli25:
Adds new APM extension books. Why? See https://github.com/elastic/observability-docs/issues/2207.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -1440,6 +1440,32 @@ contents:
                         repo:   apm-agent-rum-js
                         path:   CHANGELOG.asciidoc
                         exclude_branches:   [ 3.x, 2.x, 1.x, 0.x ]
+              - title:      APM AWS Lambda extension
+                prefix:     lambda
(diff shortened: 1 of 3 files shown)


conf.yaml
```diff
@@ -1440,6 +1440,32 @@ contents:
                         repo:   apm-agent-rum-js
                         path:   CHANGELOG.asciidoc
                         exclude_branches:   [ 3.x, 2.x, 1.x, 0.x ]
+              - title:      APM AWS Lambda extension
```
dedemorton, line 1443: I don't have strong feelings about this, but I think we use title caps for all the other book names.

dedemorton (approved):
LGTM by eyeballing it (assuming the build passes).

conf.yaml
bmorelli25: Ah. We do. Thanks.

bmorelli25:

PR #2538, opened 2022-10-06
Update .NET docs to version 1.17

bmorelli25:


gregkalapos:

_(I just pushed the missing `1.17` branch - let's see if it works now)_

bmorelli25:
Thanks. It looks like 1.13 to 1.16 are still missing. Should I remove those and just publish 1.17?

gregkalapos:
> It looks like 1.13 to 1.16 are still missing. Should I remove those and just publish 1.17?

I'd suggest let's just publish the latest for now - so yes, I'd just remove those and go with only `1.17` now.

gregkalapos (approved):

PR #2547, opened 2022-10-18
Add link in docs landing page to Elastic Stack components overview

kilfoyle:
This adds a link on the [docs landing page](https://www.elastic.co/guide/index.html), pointing to the new `An overview of the Elastic Stack` page, soon to come via https://github.com/elastic/tech-content/pull/240

@abdon and I aren't sure exactly where this link should go. It may not really belong in the "Featured topics" section of the landing page where I've put it, but the "Getting started" section is really for all the hands-on guides, so neither place seems ideal. Any suggestions are most welcome!

Rel: https://github.com/elastic/tech-content/issues/229

### DIFF
--- a/extra/docs_landing.html
+++ b/extra/docs_landing.html
@@ -68,6 +68,7 @@
       <td style="width: 50%">
         <div class="itemizedlist">
           <ul class="itemizedlist" type="disc">
+            <li class="listitem"><a href="en/welcome-to-elastic/current/stack-components.html">An overview of the Elastic Stack</a></li>
             <li class="listitem"><a href="en/cloud/current/ec-cloud-ingest-data.html">Add your data</a></li>
             <li class="listitem"><a href="en/cloud/current/ec-migrating-data.html">Migrate to Elastic Cloud</a></li>
             <li class="listitem"><a href="en/cloud/current/ec-planning.html">Plan for production</a></li>


nrichers (changes requested):
Thank you for this PR to add a link to the component overview from the front page of our docs! I don't know what the right answer is, but I think we want to tweak the location of this link a bit. The "Add your data" topic is white hot in the CrazyEgg heatmap and we likely do not want to put a conceptual overview above the task of adding data. 

image

One option might be to add it to the bottom of the content inside of the twistie, e.g. somewhere like this, maybe? Or somewhere within this section of text, so that the link is available to folks who have expanded the twistie to indicate they want to learn more about the bits that make up the stack. 

image

abdonpijpelink:
Considering the information on the new page is largely the same as the "About the Elastic Stack" section, maybe we can simply replace that section with a link to the new page? For example a link "Learn more about the Elastic Stack" as the end of the first sentence "Elastic provides..."?

KOTungseth:
> Considering the information on the new page is largely the same as the "About the Elastic Stack" section, maybe we can simply replace that section with a link to the new page? For example a link "Learn more about the Elastic Stack" at the end of the first sentence "Elastic provides..."?

I agree that the information on the new page is largely the same as the "About the Elastic Stack" section. Why not just replace the "About the Elastic Stack" text with the information on the new page? 

Or are we unable to do this because the new page includes a graphic? If so, then I agree with Nik's proposal.

abdonpijpelink:
@KOTungseth That's a good idea! This is what that would look like:

stack in landing page

If we want to go this route, I can submit a new PR for that. Note that because the landing page is not generated out of asciidoc, we'd have to author this section in HTML.

KOTungseth:
@abdonpijpelink that looks so good! If we can author this content in HTML, I say let's go for it.

abdonpijpelink:
Closing in favor of https://github.com/elastic/docs/pull/2557

PR #2554, opened 2022-10-31
Include User ID and hashed IP in Google Analytics

gtback:
Read these values from cookies; they will have been set elsewhere on elastic.co.


brianjolly (approved):
Looks good to me!

PR #2525, opened 2022-09-26
Add GA4 tracking information

gtback:
Once the preview build finishes, this can be tested at https://docs_2525.docs-preview.app.elstc.co/guide/index.html

PR #2561, opened 2022-11-10
Remove Byte order mark

gtback:
This snuck in as part of #2525


colleenmcginnis (approved):
✨  I can see that it's gone locally. Thanks @gtback!

PR #2571, opened 2022-11-21
Add glob include processor extension

lcawl:
This PR adds an extension that enables the use of wildcards in include statements.


kilfoyle (approved):
LGTM! 🐧

PR #2583, opened 2022-12-16
Fix race condition in link to current docs

abdonpijpelink:
Fixes a race condition that would break the link to the current release documentation in the banner at the top of older docs. Thank you @jloleysens  !

I've tested this fix on a local build of the docs and haven't seen the race condition since.

### DIFF
--- a/resources/web/docs_js/index.js
+++ b/resources/web/docs_js/index.js
@@ -309,7 +309,8 @@ $(function() {
   // Empty column below TOC on small screens so the demand gen content can be positioned under the main content
   var bottom_left_col = $('#bottom_left_col');
 
-  $('.page_header > a[href="../current/index.html"]').click(function() {
+  $('.page_header > a[href="../current/index.html"]').click(function(e) {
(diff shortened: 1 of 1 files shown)


gtback (changes requested):
Thanks, @abdonpijpelink 

This seems to work fine on pages that _do_ exist in the current version (I tried a bunch of times and never hit the race condition), but we probably do still need a fallback for older pages that don't have a corresponding page in the new version. Here's one example I found: https://docs_2583.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/6.3/installing-xpack-es.html

abdonpijpelink:
I've pushed a change with a fallback for pages that do not exist in current, [similar to how it's handled for the version picker](https://github.com/elastic/docs/blob/master/resources/web/docs_js/index.js#L168).

gtback (approved):
Looks good to me, and the page I was having problems with works now: it redirects to the index page of the book when the specific page doesn't exist.

PR #2584, opened 2022-12-16
fix: '_find_toplevel' fails due to "fatal: detected dubious ownership in repository" error from latest git

trentm:
A recent git update: add an owner check for the top-level directoy
  https://github.com/git/git/commit/8959555cee7ec045958f9b6dd62e541affb7e7d9
This was breaking `git rev-parse --show-toplevel` call made in
build_docs.pl's `sub _find_toplevel` with the message:
    INFO:build_docs:fatal: detected dubious ownership in repository at 'some dir'
for some (macos?) users.

Removing the `--user $localUid:$localGid` for the `docker run ...` that
runs the doc build avoids this error by running the build as root
instead of as the local user (which is some UID that doesn't necessarily
map to a user in the docker container).

I am not sure there isn't some fallout from this change.

### DIFF
--- a/build_docs
+++ b/build_docs
@@ -17,7 +17,7 @@
 
 import errno
 import logging
-from os import environ, getgid, getuid
+from os import environ
 from os.path import basename, dirname, exists, expanduser, isdir
 from os.path import join, normpath, realpath
 import re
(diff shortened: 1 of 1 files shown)


bmorelli25:
@gtback I know you're no longer maintaining elastic/docs, but would you be able to look at this and the attached issue?

trentm:
See @gtback's better alternative: https://github.com/elastic/docs/pull/2584

gtback:
Closing in favor of #2586

PR #2586, opened 2022-12-20
Tell git to treat all directories as safe

gtback:
With a recent change to git [1], running commands like `git rev-parse --show-toplevel` will fail when different directories are owned by different users. Because of the complexity of how we use Docker and expect it to work across different operating systems, the easiest approach is to restore the prior behavior. [2]

Rather than trying to run `git config` from within the build process (which runs into its own file permission errors depending on who is running the command), instead we just hardcode the "global" gitconfig file (which is located at the root of the Docker container).

[1]: https://github.com/git/git/commit/8959555cee7ec045958f9b6dd62e541affb7e7d9
[2]: https://git-scm.com/docs/git-config#Documentation/git-config.txt-safedirectory

### DIFF
--- a/build_docs
+++ b/build_docs
@@ -576,6 +576,8 @@ def standard_docker_args():
                         '--tmpfs', '/var/lib/nginx/proxy',
                         '--tmpfs', '/var/lib/nginx/uwsgi',
                         '--tmpfs', '/var/lib/nginx/scgi'])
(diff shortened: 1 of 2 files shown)


bmorelli25 (approved):
I'm not qualified to leave a review, but this PR works great locally for me 🚀

trentm (approved):
Looks reasonable to me.

PR #2592, opened 2023-01-05
Add kibana dependency to Observability Guide

lcawl:
This PR updates the Observability Guide to pull content from the kibana repo so that we can re-use some tagged regions.


ollyhowell (approved):
LGTM!

PR #2596, opened 2023-01-17
Update build config for Enterprise Search Node.js client

leemthompo:
Updates build to generate Ent Search node client docs, added in this [PR](https://github.com/elastic/enterprise-search-js/pull/20) and to 8.6 branch in [this PR](https://github.com/elastic/enterprise-search-js/pull/21).

- Updates `conf.yaml` to build Enterprise Search Node.js client docs from the [GH repo](https://github.com/elastic/enterprise-search-js/tree/main/packages/enterprise-search)
   - Remove deprecated clients from build
- Adds `enterprise-search-node-ref` to shared attributes

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -1060,7 +1060,7 @@ contents:
                 private:    1
                 single:     1
                 current:    *stackcurrent
-                branches:   [ {main: master}, 8.6, 8.5, 8.4, 8.3, 8.2, 8.1 ]
+                branches:   [ {main: master}, 8.6]
                 live:       [ main, 8.6 ]
                 index:      docs/guide/index.asciidoc
                 tags:       Enterprise Search Clients/Node.js
--- a/conf.yaml
+++ b/conf.yaml
@@ -1038,6 +1038,23 @@ contents:
                   -
                     repo:   docs
                     path:   shared/versions/stack/{version}.asciidoc
+              - title:      App Search Node.js client
+                prefix:     app-search-node
+                private:    1
+                single:     1
+                current:    *stackcurrent
+                branches:   [ {main: master}, 8.6, 8.5, 8.4, 8.3, 8.2, 8.1, 8.0, 7.17, 7.16 ]
+                live:       *stacklivemain
+                index:      client-docs/app-search-node/index.asciidoc
+                tags:       App Search Clients/Node.js
+                subject:    App Search Clients
+                sources:
+                  -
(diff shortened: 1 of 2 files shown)


conf.yaml
```diff
@@ -1060,15 +1060,15 @@ contents:
                 private:    1
                 single:     1
                 current:    *stackcurrent
-                branches:   [ {main: master}, 8.6, 8.5, 8.4, 8.3, 8.2, 8.1 ]
-                live:       [ main, 8.5 ]
-                index:      client-docs/enterprise-search-node/index.asciidoc
+                branches:   [ {main: master}, 8.6]
+                live:       [ main, 8.6 ]
```
chriscressman, line 1064: I think we can use this, and then this value won't have to updated for each release:

```suggestion
                live:       *stacklivemain
```

chriscressman (approved):

conf.yaml
leemthompo: ```18:44:50 INFO:build_docs:Live branch(es) <7.17> not in <branches> in book <Enterprise Search Node.js client> at lib/ES/Book.pm line 133.```


@chriscressman Am I right that this change is root of that failure? 😅

leemthompo: I'll revert to get CI running again and check.

chriscressman: Yeah, seems right. I thought `live` could contain "extra" branches, but apparently not.

leemthompo: Getting there one build error at a time 👿

leemthompo: Need to change filenames to `index.asciidoc` in:

- https://github.com/elastic/enterprise-search-js/pull/22
- https://github.com/elastic/enterprise-search-js/pull/23

leemthompo:

leemthompo:

leemthompo:
@chriscressman let me know if you agree with deleting those deprecated clients in https://github.com/elastic/docs/pull/2596/commits/58c4727d26c4f336757a0e882dcca3137b1f1227.

chriscressman:
> @chriscressman let me know if you agree with removing those deprecated clients in [58c4727](https://github.com/elastic/docs/commit/58c4727d26c4f336757a0e882dcca3137b1f1227).

No, we should continue to publish those docs. "Deprecated" essentially means "don't use on new projects". However, there are potentially many existing users of these clients who still need the docs.

PR #2599, opened 2023-01-18
Add attribute for stack rules feature

lcawl:
This PR adds a shared attribute for the "Stack Rules" feature, which is one of the groupings in Kibana feature privileges and referenced in the prerequisites of many rule APIs.


kilfoyle (approved):
LGTM! 🐇

PR #2598, opened 2023-01-18
Bump docs version to 8.6.1

kilfoyle:
🛑 Do not merge until release day.

This updates the stack docs shared version attributes to `8.6.1`.

[docs release issue](https://github.com/elastic/dev/issues/2195)

PR #2604, opened 2023-01-25
Bump docs version to 8.6.1

kilfoyle:
This updates the stack docs shared version attributes to `8.6.1`.

[docs release issue](https://github.com/elastic/dev/issues/2195)

Note that this is a replacement for #2598 that strangely started to repeatedly fail the `self-check` CI check.


abdonpijpelink (approved):
LGTM

PR #2610, opened 2023-02-01
Adding Curator 8.0 as current

untergeek:
Will add 7 and 6 afterwards as single steps to avoid losing all at once.


untergeek:
🤦 I failed to merge my changes from `master` into the `8.0` branch, so the failure is the same.

PR #2612, opened 2023-02-01
Adding Curator 6.0 and 7.0 doc branches

untergeek:
I previously fixed these as well, but wanted 8.0 in for certain, rather than have it held up.


untergeek:
Anchor links with underscores are not smart in Asciidoc. Pruned them. Will re-submit.

PR #2613, opened 2023-02-01
Adding Curator 6.0 and 7.0 doc branches (take 2)

untergeek:
I previously fixed these as well, but wanted 8.0 in for certain, rather than have it held up.

Apparently, anchor links with underscores are problematic in Asciidoc. I pruned the offenders.


lcawl (approved):
LGTM

PR #2617, opened 2023-02-12
Add dropdown for APM and ECS logging books

bmorelli25:
### Summary

This PR adds a simple breadcrumb dropdown to the APM and ECS logging books. This dropdown makes it easier for users to navigate between books with related content.

Screenshot 2023-02-11 at 6 40 51 PM

### Preview

APM: https://docs_2617.docs-preview.app.elstc.co/guide/en/apm/guide/current/apm-components.html
ECS Logging: https://docs_2617.docs-preview.app.elstc.co/guide/en/ecs-logging/overview/current/intro.html

### Limitations

All navigation items are hardcoded, not generated. This means if a new APM or ECS logging book is added to the doc build, it will also need to be manually added to the dropdown array.

### Related

* Closes https://github.com/elastic/observability-docs/issues/2058.

PR #2619, opened 2023-02-15
Add dropdown for APM and ECS logging books

colleenmcginnis:
Related to #2617

### DIFF
--- a/resources/asciidoctor/lib/chunker/breadcrumbs.rb
+++ b/resources/asciidoctor/lib/chunker/breadcrumbs.rb
@@ -12,7 +12,63 @@ module Chunker
       result = ['<div class="breadcrumbs">']
       result += generate_breadcrumb_links(section).reverse
       result << '</div>'
+      if result[2].to_s.include? 'APM'
+        result.insert(2, generate_apm_breadcrumbs)
+      end
+      if result[2].to_s.include? 'ECS Logging'
+        result.insert(2, generate_ecslogging_breadcrumbs)
+      end
       Asciidoctor::Block.new doc, :pass, source: result.join("\n")
+    end
+
+    def generate_apm_breadcrumbs
+      <<~HTML.strip
+        <span class="breadcrumb-link">
+          <div class="dropdown">
+            <span>APM ▾</span>
+              <div class="dropdown-content">
+                <strong>APM</strong></br>
+                <a href="https://www.elastic.co/guide/en/apm/guide/current/index.html">APM Guide</a></br>
+                <strong>APM agents</strong></br>
+                <a href="https://www.elastic.co/guide/en/apm/agent/android/current/index.html">Android agent</a></br>
+                <a href="https://www.elastic.co/guide/en/apm/agent/go/current/index.html">Go agent</a></br>
+                <a href="https://www.elastic.co/guide/en/apm/agent/swift/current/index.html">iOS agent</a></br>
+                <a href="https://www.elastic.co/guide/en/apm/agent/java/current/index.html">Java agent</a></br>
+                <a href="https://www.elastic.co/guide/en/apm/agent/dotnet/current/index.html">.NET agent</a></br>
+                <a href="https://www.elastic.co/guide/en/apm/agent/nodejs/current/index.html">Node.js agent</a></br>
+                <a href="https://www.elastic.co/guide/en/apm/agent/php/current/index.html">PHP agent</a></br>
+                <a href="https://www.elastic.co/guide/en/apm/agent/python/current/index.html">Python agent</a></br>
+                <a href="https://www.elastic.co/guide/en/apm/agent/ruby/current/index.html">Ruby agent</a></br>
+                <a href="https://www.elastic.co/guide/en/apm/agent/rum-js/current/index.html">RUM Agent</a></br>
+                <strong>APM extensions</strong></br>
+                <a href="https://www.elastic.co/guide/en/apm/lambda/current/index.html">AWS Lambda extension</a></br>
+                <a href="https://www.elastic.co/guide/en/apm/attacher/current/index.html">APM attacher</a>
+            </div>
+          </div>
+        <span class="chevron-right">›</span>
+      HTML
+    end
+
+    def generate_ecslogging_breadcrumbs
+      <<~HTML.strip
+        <span class="breadcrumb-link">
+          <div class="dropdown">
+            <span>ECS Logging ▾</span>
+              <div class="dropdown-content">
+                <strong>ECS Logging</strong></br>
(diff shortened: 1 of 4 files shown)


resources/asciidoctor/lib/chunker/breadcrumbs.rb
```diff
@@ -12,9 +12,76 @@ def generate_breadcrumbs(doc, section)
...
+                </ul>
+                <li class="dropdown-category">APM extensions</li>
+                <ul>
+                  <li><a href="https://www.elastic.co/guide/en/apm/lambda/current/index.html">Monitoring AWS Lambda Functions</a></li>
+                  <li><a href="https://www.elastic.co/guide/en/apm/attacher/current/index.html">Attacher</a></li>
+                </ul>
+              </ul>
+          </div>
+        </div>
+        <span class="chevron-right">›</span>
```
bmorelli25, line 59: This is an interesting bug that I can't quite figure out. Selecting top-level heading doesn't show in the breadcrumbs. Have to select second level to get it to show

Screenshot 2023-02-15 at 9 55 59 AM


Screenshot 2023-02-15 at 9 56 04 AM

bmorelli25: oops didn't mean to comment in the file, sorry

bmorelli25: Ah! That's not the bug. The bug is that the chevron appears when it shouldn't.

bmorelli25: This is a complication of replacing instead of inserting

bmorelli25: Might have to test the length of the array. If length of the array is greater than 3, include chevron

resources/asciidoctor/lib/chunker/breadcrumbs.rb
```diff
@@ -12,9 +12,76 @@ def generate_breadcrumbs(doc, section)
...
+                  <li><a href="https://www.elastic.co/guide/en/apm/agent/go/current/index.html">Go Agent Reference</a></li>
+                  <li><a href="https://www.elastic.co/guide/en/apm/agent/swift/current/index.html">iOS Agent Reference</a></li>
+                  <li><a href="https://www.elastic.co/guide/en/apm/agent/java/current/index.html">Java Agent Reference</a></li>
+                  <li><a href="https://www.elastic.co/guide/en/apm/agent/dotnet/current/index.html">.NET Agent Reference</a></li>
+                  <li><a href="https://www.elastic.co/guide/en/apm/agent/nodejs/current/index.html">Node.js Agent Reference</a></li>
+                  <li><a href="https://www.elastic.co/guide/en/apm/agent/php/current/index.html">PHP Agent Reference</a></li>
+                  <li><a href="https://www.elastic.co/guide/en/apm/agent/python/current/index.html">Python Agent Reference</a></li>
+                  <li><a href="https://www.elastic.co/guide/en/apm/agent/ruby/current/index.html">Ruby Agent Reference</a></li>
+                  <li><a href="https://www.elastic.co/guide/en/apm/agent/rum-js/current/index.html">Real User Monitoring JavaScript Agent Reference</a></li>
```
bmorelli25, line 49: Do you think relative links would work? The preview sending you to the public docs is confusing.

resources/asciidoctor/lib/chunker/breadcrumbs.rb
colleenmcginnis: Yea I thought that was strange at first too, but that's actually how it works in production now (we just have to hide that last >).

**Production:**

Screenshot 2023-02-15 at 12 13 43 PM


**Preview:**

Screenshot 2023-02-15 at 12 14 17 PM

colleenmcginnis: I guess the question is: do breadcrumbs usually include the title of the current page or is that assumed because it's directly under the breadcrumbs? I guess it would be silly to add a link to the current page in the breadcrumbs...

resources/asciidoctor/lib/chunker/breadcrumbs.rb
colleenmcginnis: I don't see why not.

colleenmcginnis:
@bmorelli25 Do you think it's important for us to add the version number to the title in the breadcrumb or do you think having the version dropdown near it is sufficient? 

| Without version | With version |
|---|---|
| Screenshot 2023-02-15 at 12 23 46 PM | Screenshot 2023-02-15 at 12 23 26 PM |

You can't choose other versions in the related product drop down so it might be a bit misleading to add the version number there, but I do think it's possible. For the screenshot I faked it (because locally you don't see version numbers), but we should be able to get it via [`title-extra`](https://github.com/elastic/docs/blob/master/resources/asciidoctor/lib/chunker/breadcrumbs.rb#L23).

resources/asciidoctor/lib/chunker/breadcrumbs.rb
colleenmcginnis: >Might have to test the length of the array. If length of the array is greater than 3, include chevron

OR add a chevron _before_ each item unless it's the manually-built, first item "Elastic Docs". That seems less confusing to me at least.

bmorelli25:
> Do you think it's important for us to add the version number to the title in the breadcrumb or do you think having the version dropdown near it is sufficient?

I think dropdown is sufficient

resources/asciidoctor/lib/chunker/breadcrumbs.rb
bmorelli25: > OR add a chevron before each item unless it's the manually-built, first item "Elastic Docs". That seems less confusing to me at least.

I think this will be very complicated. The array is built and then reversed

resources/asciidoctor/lib/chunker/breadcrumbs.rb
bmorelli25: Testing in https://github.com/elastic/docs/pull/2617

bmorelli25:
@colleenmcginnis The tests are now fixed. I tried to add a new test for the APM books, but  https://github.com/elastic/docs/blob/master/resources/asciidoctor/spec/chunker_spec.rb is a bit too complex for me and I can't get a new APM test to work. I think it's okay to merge this as-is and open an issue to add APM and ECS tests later. Sound good?

bmorelli25:
Tested on mobile. GitHub won't let me upload an image for some reason but it looks fine.

bmorelli25 (approved):
Looks awesome! Let's 🚢 it!

PR #2625, opened 2023-03-06
Remove .NET example reference

stevejgordon:
Removing the reference to the outdated .NET examples folder which doesn't seem to be used.


szabosteve (approved):
LGTM, thanks!

PR #2626, opened 2023-03-06
Add noindex tag to Azure Marketplace docs

kilfoyle:
We've been requested by the marketing team working with MS Azure to help reduce the prominence of the [Azure Marketplace Solution](https://www.elastic.co/guide/en/elastic-stack-deploy/master/azure-marketplace.html) docs in the Google search results. The change here will help with that.

Separately, I've added a recommendation notice at the top of those docs guiding people toward the newer Elastic Cloud offering in Azure Marketplace, as well as the associated docs.


abdonpijpelink (approved):
LGTM

PR #2630, opened 2023-03-20
Update Rust Client to build from main

kilfoyle:
The elasticsearch-rs repo now builds from main rather than master as a result of [this commit](https://github.com/elastic/elasticsearch-rs/commit/790d523e0b436da3765e80c4da6927814a58c2a1). This updates the doc builds accordingly.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -595,8 +595,8 @@ contents:
               - title:      Rust Client
                 prefix:     rust-api
                 current:    master
-                branches:   [ master, 8.7, 8.6, 8.5, 8.4, 8.3, 8.2, 8.1, 8.0 ]
-                live:       [ master, 8.7, 8.6, 8.5, 8.4, 8.3, 8.2, 8.1, 8.0 ]
+                branches:   [ main, 8.7, 8.6, 8.5, 8.4, 8.3, 8.2, 8.1, 8.0 ]
+                live:       [ main, 8.7, 8.6, 8.5, 8.4, 8.3, 8.2, 8.1, 8.0 ]
                 index:      docs/index.asciidoc
                 chunk:     1
                 tags:       Clients/Rust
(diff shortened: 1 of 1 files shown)


conf.yaml
```diff
@@ -595,8 +595,8 @@ contents:
               - title:      Rust Client
                 prefix:     rust-api
                 current:    master
```
bmorelli25, line 597: I think this also needs to be `main`
```suggestion
                current:    main
```

bmorelli25 (approved):
🍁

conf.yaml
```diff
@@ -594,9 +594,9 @@ contents:
                     path:   docs/guide
               - title:      Rust Client
                 prefix:     rust-api
-                current:    master
-                branches:   [ master, 8.7, 8.6, 8.5, 8.4, 8.3, 8.2, 8.1, 8.0 ]
-                live:       [ master, 8.7, 8.6, 8.5, 8.4, 8.3, 8.2, 8.1, 8.0 ]
+                current:    main
+                branches:   [ main, 8.7, 8.6, 8.5, 8.4, 8.3, 8.2, 8.1, 8.0 ]
```
lcawl, line 598: Since we're still publishing as "master" even when the repos use "main", I think you need to align with what we've done for the other books:
```suggestion
                branches:   [ {main: master}, 8.7, 8.6, 8.5, 8.4, 8.3, 8.2, 8.1, 8.0 ]
```

kilfoyle: Roger that. Thanks for spotting that, and for explaining!

lcawl (approved):
LGTM

lcawl:
This PR actually seems like a duplicate of https://github.com/elastic/docs/pull/2199

kilfoyle:
Very nice catch @lcawl!
So that explains the mystery of how the docs build just passed prior to this PR having merged.

Closing...

PR #270, opened 2017-12-07
Handle kibana-extra folder

kjbekkelund:
Preparing for x-pack-kibana moving into a `kibana-extra` folder, similarly to `x-pack-elasticsearch` and `x-pack-logstash`, see https://github.com/elastic/kibana/issues/15452.

This is not intended for merge just yet. I'm just going through and preparing all the necessary PRs for this move.

PR #2634, opened 2023-03-25
Start building ESF documentation

dedemorton:
### Related issues and PRs:

* https://github.com/elastic/observability-docs/issues/2716
* https://github.com/elastic/elastic-serverless-forwarder/pull/270
* https://github.com/elastic/elastic-serverless-forwarder/pull/311

Note that https://github.com/elastic/elastic-serverless-forwarder/pull/311 contains https://github.com/elastic/elastic-serverless-forwarder/pull/270. Just wanted to work in an isolated branch until [#270](https://github.com/elastic/elastic-serverless-forwarder/pull/270) is merged.

### Updated docs page (proposed--please opine)

image

### TODOs

- [x] Update conf.yaml
- [x] Add shared attribute for esf-ref (path for new ESF docs).
- [x] Add shared attribute for esf (Elastic Serverless Forwarder).
- [ ] After merging this PR, stop building ESF docs from main in obs-docs repo.
- [ ] Set up global redirects for ESF links in master.
- [ ] Make sure any in-progress PRs in the obs-docs repo get handled correctly (for example, https://github.com/elastic/observability-docs/pull/2661)

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -49,6 +49,7 @@ repos:
     enterprise-search-php: https://github.com/elastic/enterprise-search-php.git
     enterprise-search-python: https://github.com/elastic/enterprise-search-python.git
     enterprise-search-ruby: https://github.com/elastic/enterprise-search-ruby.git
+    esf:                  https://github.com/elastic/elastic-serverless-forwarder.git
     guide:                https://github.com/elastic/elasticsearch-definitive-guide.git
     guide-cn:             https://github.com/elasticsearch-cn/elasticsearch-definitive-guide.git
     ingest-docs:          https://github.com/elastic/ingest-docs.git
@@ -1808,6 +1809,22 @@ contents:
               -
                 repo:   package-spec
                 path:   spec
+
+    -   title:      "Cloud Native: Ship Data from your Cloud Provider"
+        sections:
+          - title:      Elastic Serverless Forwarder
+            prefix:     en/esf
+            index:      docs/index.asciidoc
+            current:    master
+            branches:   [ {main: master} ]
+            live:       master
+            chunk:      2
+            tags:       Cloud Native Ingest/Reference
+            subject:    cloud native ingest
+            sources:
+              -
+                repo:   esf
+                path:   docs
 
     -   title:      "Beats: Collect, Parse, and Ship"
         sections:
--- a/conf.yaml
+++ b/conf.yaml
@@ -1810,14 +1810,14 @@ contents:
                 repo:   package-spec
                 path:   spec
 
-    -   title:      "Cloud Native: Ship Data from your Cloud Provider"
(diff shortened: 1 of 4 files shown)


dedemorton:
We need to decide where to put the new ESF "book" in the current docs landing page. Probably should go somewhere near the Elastic Agent and Beats docs. (I'm not sure why Logstash comes before Elastic Agent on this page :shrug:.)

Also we need to decide what category to put this "book" under. Should we group all the docs for cloud native data ingestion under a single category? We could add the [new Kinesis firehose docs](https://observability-docs_2802.docs-preview.app.elstc.co/guide/en/observability/master/aws-firehose.html) under the new category instead of including it in the observability guide. I also wonder about the docs for the Native Azure integration (currently covered in the cloud docs, which I think makes sense because of how that works, but could be confusing to users).

cc @bmorelli25 WDYT?

image

bmorelli25:
> Should we group all the docs for cloud native data ingestion under a single category?

I think this makes sense.

> We could add the [new Kinesis firehose docs](https://observability-docs_2802.docs-preview.app.elstc.co/guide/en/observability/master/aws-firehose.html) under the new category instead of including it in the observability guide

Did we ever find out if this is going to be versioned separately from the stack? Maybe it doesn't matter anymore? If the ESF docs have their own book, it probably makes sense for firehose to have their own too.

dedemorton:
< it probably makes sense for firehose to have their own too.

@bmorelli25 Looking more closely at those docs, I'm not sure we want to do put the Kinesis firehose docs in the same section as ESF docs. We are putting the ESF docs into a separate book because ESF has its own release cycle (and versioning). It is cloud native and not part of the stack. It looks like the docs written about the Kinesis firehose describe how to use the integration. Remember I also have that big PR open that adds cloud monitoring docs (specifically AWS)--that work was driven by the dev team. And we also have those tutorials that are similar to the firehose docs. Plus of course the integrations documentation.

I feel like we need to step back and have a plan rather than just shoe horning content into various places. From the user's perspective, this is not a great user experience.

dedemorton:
Marking this as "ready for review" because the docs are building.

@aspacca @bmorelli25 Would appreciate your input on the contents of the docs landing page. (It's a bit difficult to find a good location on a page that does not follow any obvious organizing principle. 😄  )

If we want to have a high-level container for grouping docs about cloud native ingest approaches, we can do something like this. (Not sure if it's OK to use "cloud native" here...just putting something out there to start the conversation going.)

image

If we want to keep things under observability, we can do something this:

image

dedemorton:
@karenzone FYI...these docs are about cloud native monitoring but also related to ingest. :-)

dedemorton:
@tommyers-elastic Wondering if you have an opinion about the location and title for the new section on the docs landing page (see https://github.com/elastic/docs/pull/2634#issuecomment-1504198413).

bmorelli25 (approved):

PR #2648, opened 2023-03-31
Update air_gapped_template.html

bmorelli25:
Attempt to fix air-gapped docs number two


nik9000 (approved):
I don't know how to test this locally, but if it works for you it works for me.

PR #2651, opened 2023-04-04
Add padding for feedback button

abdonpijpelink:
Ironically, the most common [feedback](https://ela.st/docs-feedback) we get via the feedback button in the docs is about the feedback button itself. On small displays, the feedback button overlaps the text. This makes the documentation hard to read on a phone. 

This PR adds 25px of extra padding to the right, creating enough space for the feedback button.

Before:
image

After:
image


chriscressman (approved):
LGTM

colleenmcginnis (changes requested):
You shouldn't update the `bootstrap.css` file directly. Looking for the right place to put this and will comment here!

bmorelli25:
Is this a problem on larger screens? We should probably only apply this padding to smaller screens if not.

colleenmcginnis (commented):
Ok! More details. 🙃 

1. We shouldn't add to the `resources/web/lib/bootstrap/bootstrap.css` because this is just a copy of the CSS provided by [Bootstrap](https://bootstrapdocs.com/v3.0.3/docs/) saved locally in this repo. Instead, custom styles should be defined in the various `*.pcss` files in `resources/web/style/`. In this case, I think we can use the existing [`.guide-section` selector in the `resources/web/style/docbook.pcss`](https://github.com/abdonpijpelink/docs/blob/padding-for-feedback-button/resources/web/style/docbook.pcss#L18-L20) file.
2. We should use a media query so we're only adding the extra padding on small and medium screens. In the docs we're using [Bootstrap's responsive classes](https://bootstrapdocs.com/v3.0.3/docs/css/#responsive-utilities) to determine when to hide the right side bar (containing "On this page"). That happens when the screen gets smaller than 993px. We only want to add right side padding to the `guide-section` class when there's no right side bar. Something like this should do the trick:

```diff
 .guide-section {
   margin-bottom: 30px;
+  @media screen and (max-width: 992px) {
+    padding-right: 40px;
+  }
 }
```

Aside: If we also want to prevent the feedback button from overlapping with the contents of the right side bar on screens wider than 992px, we'd have to take a different approach. I'm not sure if that's also a concern.

Screenshot 2023-04-04 at 9 38 27 AM

abdonpijpelink:
Thanks @colleenmcginnis . I've tested your suggested change and it works. I've pushed the change to this PR.

resources/web/lib/bootstrap/bootstrap.css
```diff
@@ -9268,4 +9268,4 @@ a.text-dark:hover {
   .table-bordered th {
       border: 1px solid #ddd!important
   }
-}
+}
```
colleenmcginnis, line 9271: You could add a newline to clean up the diff, but it's up to you. 

```suggestion
}

```

colleenmcginnis (approved):
Thanks @abdonpijpelink! The preview looks good. ✨

PR #2658, opened 2023-04-11
Add DLM related attributes

gmarouli:
We are introducing new documentation for data lifecycle management (DLM) and we would like to have similar macros like we have for ILM.

Part of: https://github.com/elastic/elasticsearch/issues/93596


andreidan (approved):
Thanks Mary

PR #2660, opened 2023-04-12
Update current version for .NET client docs

stevejgordon:
We have now released 8.1.0 of the client and, therefore, should direct the current version in the docs to that branch.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -527,7 +527,7 @@ contents:
                     path:   .doc/
               - title:      .NET Clients
                 prefix:     net-api
-                current:    8.0
+                current:    8.1
                 branches:   [ {main: master}, 8.7, 8.6, 8.5, 8.4, 8.3, 8.2, 8.1, 8.0, 7.17, 7.16, 6.x, 5.x, 2.x, 1.x ]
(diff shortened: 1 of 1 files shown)


conf.yaml
```diff
@@ -527,7 +527,7 @@ contents:
                     path:   .doc/
               - title:      .NET Clients
                 prefix:     net-api
-                current:    8.0
+                current:    8.1
                 branches:   [ {main: master}, 8.7, 8.6, 8.5, 8.4, 8.3, 8.2, 8.1, 8.0, 7.17, 7.16, 6.x, 5.x, 2.x, 1.x ]
                 live:       [ main, 8.0, 7.17, 7.16, 6.x, 5.x, 2.x, 1.x ]
```
szabosteve, line 532: Please add 8.1 to the live branches, too.

```suggestion
                live:       [ main, 8.1, 8.0, 7.17, 7.16, 6.x, 5.x, 2.x, 1.x ]
```

szabosteve (changes requested):
The `live` property should also be updated with `8.1`.

stevejgordon:
> The `live` property should also be updated with `8.1`.

Yes it should, thanks!!

szabosteve (approved):
LGTM!

PR #2521, opened 2022-09-22
Output only main

lcawl:
This PR stops generating "master" documentation, so that we can find all the broken links.

PR #2663, opened 2023-04-12
Change die to return in link checks

lcawl:
This PR tests whether we can essentially make all link checking output a warning instead of an error by using `return` and `say` instead of `die`.

Relates to https://github.com/elastic/docs/pull/2521

### DIFF
--- a/lib/ES/LinkCheck.pm
+++ b/lib/ES/LinkCheck.pm
@@ -98,8 +98,7 @@ sub report {
         push @error, "  $file contains broken links to:";
         push @error, map {"   - $_"} sort keys %{ $bad->{$file} };
     }
-    die join "\n", @error, '';
-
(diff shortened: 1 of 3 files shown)


lib/ES/LinkCheck.pm
```diff
@@ -98,8 +98,7 @@ sub report {
         push @error, "  $file contains broken links to:";
         push @error, map {"   - $_"} sort keys %{ $bad->{$file} };
     }
-    die join "\n", @error, '';
-
+    warn join "\n", @error, '';
```
lcawl, line 101:
```suggestion
    return join "\n", @error, '';
```

lcawl:
This PR has been replaced by work in other PRs such as https://github.com/elastic/docs/pull/2670

PR #2665, opened 2023-04-18
Adds custom content to Clients head page

szabosteve:
## Overview

This PR customizes the Clients head page content.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -463,6 +463,7 @@ contents:
                 exclude_branches:   [ 6.2, 6.1, 6.0, 5.6, 5.5, 5.4, 5.3, 5.2, 5.1, 5.0, 2.4, 2.3, 2.2, 2.1, 2.0, 1.7, 1.6, 1.5, 1.4, 1.3, 0.90 ]
           - title:      Elasticsearch Clients
             base_dir:   en/elasticsearch/client
+            toc_extra: extra/client_docs_landing.html
             sections:
               - title:      Java Client
                 prefix:     java-api-client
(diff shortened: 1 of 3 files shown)


gtback (approved):
Looks like the build worked correctly and I see the preview of the new page [here](https://docs_2665.docs-preview.app.elstc.co/guide/en/elasticsearch/client/index.html). 

image

👍🏻 from me on the build, but I'll let @sethmlarson or someone else approve the content.

sethmlarson:
@szabosteve Is it expected to have the list of docs pages below everything? It looks strange to have that come after the marketing copy?

szabosteve:
@sethmlarson It is automatically appended to the end of the head page. 
Is it possible to remove the TOC below the custom content somehow, @gtback? Thank you for your help in advance! 🙇

gtback:
> Is it possible to remove the TOC below the custom content somehow, @gtback?

At the moment, I don't think so. I took a look briefly yesterday, and didn't see a way to do it. I know for the _book_ title/TOC pages, there's a way to use either a "custom index page" (which replaces the whole thing) or a "extra index page" (which just adds more content above the existing TOC). I don't believe this full-replacement "custom" behavior is possible for the _sub-section_ index pages, though. It could probably be added easily enough though by anyone who wants to dig into the guts of the Perl code 😁 .

szabosteve:
Thank you very much for your help and answer, @gtback!
@sethmlarson What do you think? Is it okay to merge as is or should we wait until someone can change the Perl code? I would volunteer but the only thing that I know about Perl is that its official logo is under a surprisingly strict license. 😄 I can still take a look though when I have a bit more time.

sethmlarson (approved):

PR #2666, opened 2023-04-18
Add shared attribute for kibana blob

lcawl:
This PR updates https://github.com/elastic/docs/blob/master/shared/attributes.asciidoc in order to remove the book-scoped `blob` attribute from https://github.com/elastic/kibana/blob/main/docs/index.asciidoc


lcawl:

PR #2667, opened 2023-04-24
Fix typo in esf version attribute

dedemorton:
Forgot that I used an underscore instead of dash to match the convention used for ECS. :-) 

This will fix the build error in https://github.com/elastic/observability-docs/pull/2853.


mdbirnstiehl (approved):
LGTM!

PR #2670, opened 2023-04-25
Add warnlinkcheck build option

lcawl:
This PR adds an optional --warnlinkcheck option, which enables you to run the build without failing if there's a broken link. Unlike the `--skiplinkcheck` option, it still performs the check and prints the details in the log.

For example, in a local test with broken links, the output still contains info like this:

```
INFO:build_docs:Bad cross-document links:
INFO:build_docs:  /tmp/docsbuild/target_repo/html/en/apm/agent/java/1.x/config-logging.html contains broken links to:
INFO:build_docs:   - en/ecs-logging/overview/master/intro.html
...
```

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -1129,4 +1129,7 @@ sub check_opts {
         die('--reference only compatible with --all or --preview') if $Opts->{reference};
         die('--target_repo only compatible with --all or --preview') if $Opts->{target_repo};
     }
+    if ($Opts->{skiplinkcheck} && $Opts->{warnlinkcheck} ) {
+        die('--warnlinkcheck is incompatible with --skiplinkcheck');
+    }
 }
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -337,9 +337,13 @@ sub check_links {
         say $link_checker->report;
     }
     else {
+      if ( $Opts->{warnlinkcheck} ){
+        say $link_checker->report;
+      }
+      else {
         die $link_checker->report;
+      }
     }
-
 }
 
 #===================================
@@ -1007,6 +1011,7 @@ sub command_line_opts {
         'reference=s',
         'reposcache=s',
         'skiplinkcheck',
+        'warnlinkcheck',
         'sub_dir=s@',
         'user=s',
         # Options only compatible with --preview
@@ -1069,6 +1074,7 @@ sub usage {
           --repos_cache     Directory to which working repositories are cloned.
                             Defaults to `<script_dir>/.repos`.
           --skiplinkcheck   Omit the step that checks for broken links
+          --warnlinkcheck   Checks for broken links but does not fail if they exist
           --sub_dir         Use a directory as a branch of some repo
                             (eg --sub_dir elasticsearch:master:~/Code/elasticsearch)
           --target_repo     Repository to which to commit docs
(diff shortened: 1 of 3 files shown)


build_docs.pl
```diff
@@ -337,9 +337,13 @@ sub check_links {
         say $link_checker->report;
```
gtback, line 337: I'm really confused how the logic in this method works. It seems like "has_bad" returns "true" if there _aren't_ any bad links? 😕 . In which case we could maybe combine the logic for `if ( $link_checker->has_bad || $Opts->{warnlinkcheck})`?

build_docs.pl
```diff
@@ -337,9 +337,13 @@ sub check_links {
...
     }
     else {
+      if ( $Opts->{warnlinkcheck} ){
```
gtback, line 340: Separately, I wonder if it would be good to pass this in as a boolean variable (to say whether broken links should be treated as "warn" or "die"), and then call the function as `check_links($build_dir, $Opts->{warnlinkcheck})` so we could potentially call this function in different ways in different scenarios without needing to worry about the command line args.

build_docs.pl
```diff
@@ -1069,6 +1074,7 @@ sub usage {
           --repos_cache     Directory to which working repositories are cloned.
                             Defaults to `<script_dir>/.repos`.
           --skiplinkcheck   Omit the step that checks for broken links
+          --warnlinkcheck   Checks for broken links but does not fail if they exist
```
gtback, line 1077: We might want to define how things work if both `--skiplinkcheck` and `--warnlinkcheck` are provided. I assume we'd ignore and not warn about anything if we skip link checking entirely. But do we want to raise an error saying "`--warnlinkcheck` cannot be used if `--skiplinkcheck` is provided"?

build_docs.pl
```diff
@@ -1113,6 +1119,7 @@ sub check_opts {
         die('--rebuild only compatible with --all') if $Opts->{rebuild};
         die('--reposcache only compatible with --all') if $Opts->{reposcache};
         die('--skiplinkcheck only compatible with --all') if $Opts->{skiplinkcheck};
+        die('--warnlinkcheck only compatible with --all') if $Opts->{warnlinkcheck};
```
gtback, line 1122: Down here is probably where you'd want to raise an error of both options are provided.

integtest/spec/helper/dest.rb
```diff
@@ -235,6 +235,11 @@ def skip_link_check
       self
     end
 
+    def warn_link_check
```
gtback, line 238: We can use this elsewhere in integration tests, but as of right now this isn't doing anything.

lcawl: Good point, I've removed it in 24d9d57f552c65e47f9493f7557962a7a6b4baa7

build_docs.pl
lcawl: Done!  I added a clause in fdc74aa2d651c62fb9eafa72b7c879489714a1ad

build_docs.pl
lcawl: I've add a check to fail if both are used: fdc74aa2d651c62fb9eafa72b7c879489714a1ad

build_docs.pl
lcawl: Yeah, that seems like a misnomer to me too. I considered renaming it as part of this PR, but was concerned what other things would crawl out from under the rock if I did so

lcawl: I've implemented your suggestion in ba5bd2165ac4badafd0797b8de9bb89ca0109849

build_docs.pl
lcawl: I'm not sure I understand how this helps us avoid command line args. Wouldn't you still need to specify "--warnlinkcheck=xyz"?

lcawl:

build_docs.pl
gtback: Sorry, it wouldn't affect the command line args at all. It just makes the `check_links` function easier to use in the future (if you want the warn-vs-die decision to be based on something other than the argument). It's probably pre-mature optimization at this point, though.

gtback (approved):

PR #2677, opened 2023-05-11
Fix tabbed widgets

colleenmcginnis:
There was a report today that the tabbed widgets were not working. Hovering over an inactive tab results in style changes, but the click event listener is not being applied, so the active tab doesn't change and the panel contents don't change.

After poking around, I found that the [`DOMContentLoaded` event listener](https://github.com/elastic/docs/blob/master/resources/web/docs_js/components/tabbed_widget.js#LL2C20-L2C20) in the tabbed widget component isn't working. I don't know why, but using the `readystatechange` event listener and waiting for the document's `readyState` to be `interactive` does work locally at least. 🤷‍♀️ 

## For reviewers

Find a couple pages in your team's docs that use the tabbed widget and make sure switching between tabs works as expected. Preview link: https://docs_2677.docs-preview.app.elstc.co/guide

### DIFF
--- a/resources/web/docs_js/components/tabbed_widget.js
+++ b/resources/web/docs_js/components/tabbed_widget.js
@@ -1,5 +1,7 @@
 export const switchTabs = () => {
-  window.addEventListener("DOMContentLoaded", () => {
+  
+  document.addEventListener("readystatechange", () => {
+   if (document.readyState === "interactive") {
     const tabs = document.querySelectorAll('[role="tab"]');
     const tabList = document.querySelector('[role="tablist"]');
     // Add a click event handler to each tab
@@ -30,6 +32,7 @@ export const switchTabs = () => {
         tabs[tabFocus].focus();
       }
     });
+   }
   });
 }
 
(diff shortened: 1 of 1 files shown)


colleenmcginnis:
Oh maybe the real trick is using `document` instead of `window` (not `readystatechange` vs. `DOMContentLoaded`). Still not sure why it would stop working suddenly though...

kilfoyle:
This works for me in a local build. I tested the ingest-docs and stack-docs repos on Chrome and Firefox.

chriscressman (approved):
:tada:

kilfoyle (approved):
LGTM! 👍 
I can't comment on the code, but the fix works swimmingly!

dedemorton (approved):
LGTM. Works as expected in the obs docs.

colleenmcginnis:
I _think_ this was a result of some changes to the marketing header that were fixed outside of this repo on Thursday. The tabs are working in production now, but I'm going to merge this to make this functionality less fragile in the event that another change to the header impacts when the `DOMContentLoaded` event occurs when listening on the `window` interface.

PR #2683, opened 2023-05-17
Add section about attribute scope to readme

dedemorton:
We're seeing some unexpected behavior when attributes are used to set metadata. Resetting the attribute when it's no longer needed should be a best practice (just as it is when variables are used in code).


kilfoyle (approved):
LGTM! 🦖

PR #2688, opened 2023-05-22
Add version attribute for Kinesis firehose docs

dedemorton:
Adds version attribute for Kinesis firehose docs because this feature "is not tied to stack release versions, but for under-the-hood reasons is tied to ESS release milestones. in the future though this we be silently decoupled from the ESS releases."


dedemorton:
@KOTungseth Would you prefer that the related links for the Kinesis firehose docs (and ESF) user current instead of master?

dedemorton:
i'm going to merge to unblock the WN PR, then I can follow up with a PR if we want to use current instead of master in the links.

PR #2692, opened 2023-06-05
Update doc_build_aliases.sh

bmorelli25:
### Summary

This PR updates the build alias for the Observability Guide.

For https://github.com/elastic/observability-docs/pull/2992.


bmorelli25:
Unfortunately, we don't have an easy way to remove the dependency from `conf.yaml`. `elastic/beats` is still a dependency for the Observability Guide in versions 6.9-8.7. Unless we backport https://github.com/elastic/observability-docs/pull/2992 all the way to 6.9, we can't remove the dependency.

We could update the `exclude_branches` key to include `8.8` and `main`, but it would need to be updated every time we release a new branch.

dedemorton:
Oh right. I'm being slowwww. 🐌 I think it's easier just to leave the conf.yaml.

dedemorton (approved):
LGTM

PR #1383, opened 2019-11-05
Integrate feedback mechanism

nik9000:
Integrate a fancy feedback mechanism.

PR #2693, opened 2023-06-06
Remove Qualtrics survey

jrodewig:
Removes the Qualtrics feedback survey from all elastic.co/guide pages.

feedback-survey

We're not currently acting on feedback from the survey, which can create a poor user experience. We may revisit the idea of another feedback mechanism in the future.


bmorelli25 (approved):
Ahh just revert #1383. Super easy. Thanks!

PR #2695, opened 2023-06-08
Make tables stylable

abdonpijpelink:
The current default styling of tables is not great. Even though Asciidoctor allows [assigning roles to a table](https://docs.asciidoctor.org/asciidoc/latest/tables/assign-a-role/), which are rendered as CSS `class` attributes in HTML, [our implementation of tables](https://github.com/elastic/docs/blob/master/resources/asciidoctor/lib/docbook_compat/convert_table.rb#L50) does not support that.

This PR enables that functionality, by passing any roles as `class` attributes of the rendered table. 

The PR also introduces two new CSS classes: `styled` and `fixed-width`, which can optionally be assigned as roles to a table, using for example `[.styled]`,  or even `[.styled.fixed-width]`, if you want to apply both.

It enables us to do something like this (note the `.styled`):

```
.Script {watcher-transform} settings
[.styled,options="header"]
|======
| Name      |Required | Default    | Description

| `inline`  | yes     | -          | When using an inline script, this field holds
                                     the script itself.

| `id`      | yes     | -          | When referring to a stored script, this
                                     field holds the id of the script.

| `lang`    | no      | `painless` | The script language

| `params`  | no      | -          | Additional parameters/variables that are
                                     accessible by the script
|======
```

This gets rendered as:

image

Which is an improvement over [the current styling](https://www.elastic.co/guide/en/elasticsearch/reference/master/transform-script.html#transform-script-settings). I'm sure we can make tables even more beautiful than this, but this is a step forward already ("Progress, perfection").

This change does not affect any existing tables. Those tables don't have any roles assigned to them, and they will be rendered as they have been before.

### DIFF
--- a/resources/asciidoctor/lib/docbook_compat/convert_table.rb
+++ b/resources/asciidoctor/lib/docbook_compat/convert_table.rb
@@ -53,7 +53,7 @@ module DocbookCompat
         %( border="#{border}" cellpadding="4px"),
         node.title ? %( summary="#{strip_tags node.title}") : nil,
         (width = node.attr 'width') ? %( width="#{width}") : nil,
-        node.role ? %( class=#{node.role}) : nil,
+        node.role ? %( class="#{node.role}") : nil,
         '>',
       ].compact.join
     end
--- a/resources/asciidoctor/lib/docbook_compat/convert_table.rb
+++ b/resources/asciidoctor/lib/docbook_compat/convert_table.rb
@@ -53,6 +53,7 @@ module DocbookCompat
         %( border="#{border}" cellpadding="4px"),
         node.title ? %( summary="#{strip_tags node.title}") : nil,
         (width = node.attr 'width') ? %( width="#{width}") : nil,
+        node.role ? %( class=#{node.role}) : nil,
         '>',
       ].compact.join
     end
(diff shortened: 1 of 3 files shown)


resources/asciidoctor/lib/docbook_compat/convert_table.rb
```diff
@@ -53,6 +53,7 @@ def convert_table_tag(node, border)
         %( border="#{border}" cellpadding="4px"),
         node.title ? %( summary="#{strip_tags node.title}") : nil,
         (width = node.attr 'width') ? %( width="#{width}") : nil,
+        node.role ? %( class="#{node.role}") : nil,
```
nik9000, line 56: Is there a chance you'd like to write a test for this one? At this point we're not great at maintaining all this, but a test would help to make sure that the next person who touches this doesn't break it. The tests are all hiding in `resources/asciidoctor/spec/docbook_compat_spec.rb` and `make check` should just run them all.

nik9000 (approved):
Ruby change looks great to me. It'd be lovely to have a test. I know that's more work than the change itself but it helps prevent someone in the future breaking it accidentally.

I wonder about the names of the styles - `fixed-width` sounds to me like it'd make the table's width fixed but you are really making the text inside fixed. And `styled` just sounds pretty generic. OTOH, the old tables really are "plain" and this is "pretty". So I dunno.

The names can be changed without too much trouble so I don't want to block this. Let it land, it's genuinely lovely.

nik9000:
All tests passed for me. Bring it home, jenkins!

bmorelli25:
AND all with signed commits. Great stuff! 🚢

PR #2697, opened 2023-06-09
Expand styles for ESQL

nik9000:
This implements the `esql` language in prettify so it'll look nice and it adds another couple of styles that let us merge the results like I'm trying to do here:
https://github.com/elastic/elasticsearch-internal/pull/1259


nik9000:
I can make tests for the style!

Issue #2380, opened 2022-02-17
Hide unsupported langauges from example dropdowns


gtback:
For example, [this page](https://www.elastic.co/guide/en/elasticsearch/reference/8.0/processors.html) shows a dropdown even though we only have the example as a "Console". We should remove languages that don't have the listed example, and ideally remove the dropdown entirely if there's only one choice.

PR #2705, opened 2023-06-20
Only enable console alternatives that have a snippet

abdonpijpelink:
In books that have configured alternative languages for `console` code examples, the alternative picker currently shows all the configured languages, regardless of whether a snippet actually exists for each of the languages:

image

This leads to a poor UX, as users can select a language, only to find out that a code snippet does not exist for that language.

This PR changes the alternative picker such that, per code snippet, only those languages for which snippets actually exist are enabled. Other languages are still shown greyed out, but cannot be selected:

image

If a user has previously selected a language on another code snippet, but that language does not exist on the current example, the option will be disabled, but still preselected, so their preference is not reset:

image

(If you want to give it a spin, take a look at the first two code snippets on [this page](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-terms-aggregation.html))

Closes https://github.com/elastic/docs/issues/2380
Related: https://github.com/elastic/tech-content/issues/314

### DIFF
--- a/resources/web/docs_js/__tests__/components/alternative_picker.test.jsx
+++ b/resources/web/docs_js/__tests__/components/alternative_picker.test.jsx
@@ -53,6 +53,37 @@ describe(_AlternativePicker, () => {
     });
   });
 
+  describe("when console alternative snippets are missing", () => {
+    const picker = render(<_AlternativePicker
+      consoleAlternative="console"
+      alternatives={{
+        console: {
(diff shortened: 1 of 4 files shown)


JoshMock (approved):
looks great from a JS/TS perspective :clap:

jrodewig (approved):

PR #2672, opened 2023-04-28
Adds 8.8 to conf.yaml and the 8.8.0 attributes

nastasha-solomon:
- Creates a new `/shared/versions/stack/8.8.asciidoc` file copied from `8.7.asciidoc`
- Updates the `master.asciidoc` to show the next version (8.9)
- Updates `conf.yaml` to include the 8.8 branch

PR #2708, opened 2023-06-21
Add ECS for 8.9 FF

bhapas:


shared/versions/stack/master.asciidoc
```diff
@@ -13,7 +13,7 @@ bare_version never includes -alpha or -beta
 :prev-major-version:     7.x
 :prev-major-last:        7.17
 :major-version-only:     8
-:ecs_version:            master
+:ecs_version:            8.9
```
kgeller, line 16: Can someone from docs verify this is ok? 

We noticed that as a part of the [8.8 FF PR](https://github.com/elastic/docs/pull/2672) this was changed to master from the specific version. Previously, our team (ECS) has always set this ([8.7 FF PR](https://github.com/elastic/docs/pull/2669/files))

kilfoyle: Yeah, that looks like it indeed should be `8.9`. Nice catch!

kilfoyle (approved):
LGTM! 🎸

shared/versions/stack/master.asciidoc
kgeller: Awesome, thanks @kilfoyle !

kilfoyle: Let me check on this...

kilfoyle: Checked again. Yes indeed, it should be `8.9`. :-)

kilfoyle (approved):
LGTM! ⛵

PR #2710, opened 2023-06-30
Temporarily pause Kibana link checks

jrodewig:
The checker isn't skipping GitHub links as intended. This is blocking https://github.com/elastic/kibana/pull/160535.

I think we can fix this by updating the [check_kibana_links](https://github.com/elastic/docs/blob/master/build_docs.pl#L381) function, but I don't want to halt the above PR while we research.

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -330,7 +330,8 @@ sub check_links {
 
     $link_checker->check;
 
-    check_kibana_links( $build_dir, $link_checker ) if exists $Conf->{repos}{kibana};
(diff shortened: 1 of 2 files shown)


nik9000 (approved):

bmorelli25 (changes requested):
Temporarily blocking this. I don't think the link checker is broken. See https://github.com/elastic/kibana/pull/160535#issuecomment-1614886819.

bmorelli25 (approved):
Ehh okay I don't know enough about the Kibana link checker to stop this PR :)

PR #2712, opened 2023-06-30
Add ELASTIC_GITHUB to link checker

lcawl:
This PR skips the "ELASTIC_GITHUB" URL added in https://github.com/elastic/kibana/pull/160535, which would otherwise cause the Kibana link checks to fail.

It also reverts the temporary changes from https://github.com/elastic/docs/pull/2710


jrodewig (approved):
LGTM. I added some commits to revert https://github.com/elastic/docs/pull/2710 and add a test. Feel free to get another pair of eyes if wanted.

PR #2713, opened 2023-07-06
Add an attribute for ES|QL

abdonpijpelink:
ESQL has been renamed ES|QL. This adds an attribute so we can use `{esql}` in the docs.


jrodewig (approved):
Looks|good|to|me 😄

nik9000 (commented):
L|G|T|M

PR #2714, opened 2023-07-07
Add attribute for "Elastic Artifact Registry"

kilfoyle:
We have an attribute for the Elastic Package Registry. For consistency, we should also have one for the Elastic Artifact Registry.


karenzone (approved):
Nice! LGTM

PR #2725, opened 2023-07-28
Update ECS docs to version 8.9

kilfoyle:
The ECS version in the 8.9 docs was accidentally left at `master` rather than `8.9`.


jrodewig (approved):
Thanks for catching this @kilfoyle!

PR #2732, opened 2023-08-10
Add esql qa resources to conf.yaml

ChrisHegarty:
This commit adds the x-pack/plugin/esql/qa/testFixtures/src/main/resources resource for building the Elasticsearch. The directory contains files that are used for code snippet testing.

Currently only required for the `feature/esql` branch, but will be required for `main` once merged, post 8.10.

PR #2734, opened 2023-08-14
Enable ESQL qa resources for the main branch

ChrisHegarty:
This commit enables the x-pack/plugin/esql/qa/testFixtures/src/main/resources resource for building the Elasticsearch main branch. The directory contains files that are used for code snippet testing.

This will be required for once ESQL merges into _main_, post 8.10, see https://github.com/elastic/elasticsearch/pull/98309.

relates #2732


ChrisHegarty:
The CI currently fails with:

```
INFO:build_docs:fatal: pathspec 'x-pack/plugin/esql/qa/testFixtures/src/main/resources' did not match any files
```

 , which is expected until after ESQL merges into _main_. After which this change will be merged.

PR #2735, opened 2023-08-14
Move some book-scoped attributes from /observability-docs

colleenmcginnis:
🧀  Pairs with https://github.com/elastic/observability-docs/pull/3138

Moves some book-scoped variables from /observability-docs here.


bmorelli25 (approved):
✨ Thanks!

PR #2738, opened 2023-08-16
Add 'Elastic Serverless' shared attribute

kilfoyle:
I expect we'll want this attribute for mentioning any restrictions or other references about Serverless from within the classic docs.

### DIFF
--- a/shared/attributes.asciidoc
+++ b/shared/attributes.asciidoc
@@ -175,6 +175,7 @@ Elastic Cloud
 :ess:         Elasticsearch Service
 :ece:         Elastic Cloud Enterprise
 :eck:         Elastic Cloud on Kubernetes
+:serverless:  Elastic Serverless
 :cloud:       https://www.elastic.co/guide/en/cloud/current
 :ess-trial:   https://www.elastic.co/cloud/elasticsearch-service/signup?baymax=docs-body&elektra=docs
 :ess-product: https://www.elastic.co/cloud/elasticsearch-service?baymax=docs-body&elektra=docs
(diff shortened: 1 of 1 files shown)


jrodewig (approved):
LGTM. Adding the Solution writers for awareness.

kilfoyle:
>Do we also want an attribute for just Serverless? For times when we don't want to say Elastic Serverless over and over, but want to use an attribute (in case they ever rename Serverless, etc.)?

Good idea @joepeeples! I've replaced the original with short and full versions.

>Also, how will this attribute get added to Docsmobile or Serverless staging as a site-wide variable?

I was going to offer to add it, but I see that on the staging site currently there are no [site-wide variables](https://github.com/elastic/docs-staging.elastic.dev/blob/main/config/variables.js) defined there yet, so maybe it'd be best for a mass import of the variables we'll need on staging, rather than moving variables over piecemeal. For now, I'm mainly interested to have the variable available in classic so that, for example, we can handle requests to say that "Feature X is not available in Serverless", etc.

joepeeples:
> on the staging site currently there are no site-wide variables defined there yet

The `docs-staging.elastic.dev` repo does have site-wide variable definitions, such as [terms.js](https://github.com/elastic/docs-staging.elastic.dev/blob/main/config/variables/terms.js), but I don't know if there's a process to keep those updated and in sync with the latest AsciiDoc attributes.

kilfoyle:
Thanks Joe! I mistakenly was looking at the empty `variables.js` file on staging. I'm adding the two Serverless product name variables to the staging site via https://github.com/elastic/docs-staging.elastic.dev/pull/45

PR #2765, opened 2023-09-18
Add APM PHP agent conditional

colleenmcginnis:
Related to https://github.com/elastic/docs/pull/2735

@bmorelli25 it looks like we missed the PHP agent in https://github.com/elastic/docs/pull/2735. I'm seeing unresolved attributes in the ["PHP" tab in the "Step 2: Install APM agents" section](https://www.elastic.co/guide/en/observability/current/ingest-traces.html#add-apm-integration-agents). I think might fix it? 

Screenshot 2023-09-18 at 3 07 40 PM


bmorelli25 (approved):
Looks good! Thanks for fixing.

PR #2771, opened 2023-09-26
Add `--build_live_only` flag to only build live branches

jrodewig:
**Problem:**
Stack releases often require a time-consuming full docs rebuild. Long build times can cause release delays, especially when multiple builds are needed due to various issues. Rebuilding unmaintained books significantly contributes to these long build times.

**Solution:**
Add an option to only rebuild actively maintained (`live`)  branches during a full docs rebuild.

**Build time benchmarks:**
Only rebuilding live branches: 23 min. [[link](https://elasticsearch-ci.elastic.co/job/elastic+docs+pull-request+build/1739/timings/)]
Full rebuild: 1 hr. [[link](https://elasticsearch-ci.elastic.co/job/elastic+docs+pull-request+build/1738/timings/)]

Log of a build using the `--build_live_only` flag [[link](https://elasticsearch-ci.elastic.co/job/elastic+docs+pull-request+build/1739/console)]

**Infra updates:**
To have this work in CI, we'll need to make updates to our Jenkins jobs: https://github.com/elastic/infra/pull/39906

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -1007,6 +1007,7 @@ sub command_line_opts {
         'reposcache=s',
         'skiplinkcheck',
         'warnlinkcheck',
+        'build_live_only',
         'sub_dir=s@',
         'user=s',
         # Options only compatible with --preview
@@ -1070,6 +1071,7 @@ sub usage {
                             Defaults to `<script_dir>/.repos`.
           --skiplinkcheck   Omit the step that checks for broken links
           --warnlinkcheck   Checks for broken links but does not fail if they exist
+          --build_live_only Only build `live` branches
           --sub_dir         Use a directory as a branch of some repo
                             (eg --sub_dir elasticsearch:master:~/Code/elasticsearch)
           --target_repo     Repository to which to commit docs
@@ -1115,6 +1117,7 @@ sub check_opts {
         die('--reposcache only compatible with --all') if $Opts->{reposcache};
         die('--skiplinkcheck only compatible with --all') if $Opts->{skiplinkcheck};
         die('--warnlinkcheck only compatible with --all') if $Opts->{warnlinkcheck};
+        die('--build_live_only only compatible with --all') if $Opts->{build_live_only};
         die('--sub_dir only compatible with --all') if $Opts->{sub_dir};
     }
     if ( !$Opts->{preview} ) {
--- a/integtest/spec/all_books_spec.rb
+++ b/integtest/spec/all_books_spec.rb
@@ -787,6 +787,29 @@ RSpec.describe 'building all books' do
     end
   end
 
+  context 'when run with --build_live_only' do
+    convert_before do |src, dest|
+      repo = src.repo_with_index 'repo', 'some text'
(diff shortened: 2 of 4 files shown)


jrodewig:
Log of a build using the flag: https://elasticsearch-ci.elastic.co/job/elastic+docs+pull-request+build/1737/console

kilfoyle (approved):
On a cursory look the code LGTM. If things don't work we'll know soon enough. :-)

jrodewig:

jrodewig:

jrodewig:
Thanks @kilfoyle.

I'll let this set for a couple days to give others a chance to comment. If I don't hear objections, I'll plan to merge it on Tue, Oct 3.

bmorelli25 (commented):
Great feature addition. I didn't test this, but the logic makes sense.

My only concern is on the Jenkins side. We're currently migrating to buildkite and I wonder if we should wait to make this change (https://github.com/elastic/infra/pull/39906) until after that migration is complete. Nassim is already in the process of migrating our scripts, and I'm concerned this would add unnecessary friction to that project. @gtback, any thoughts?

jrodewig:
> I wouldn't merge the Infra PR without coordinating with @nkammah and the team doing the Jenkins->Buildkite migration, so it's not lost.

Sounds reasonable. I'll probably just wait until the migration is wrapped up before moving that PR out of draft and making related Buildkite changes. Thanks all!

gtback:
Unrelated to the timing, I'm trying to think about how this interacts with the `--rebuild` option, especially on PRs. Does it ever make sense to include this flag if `--rebuild` isn't included? I imagine that could lead to unintuitive results (in a backport PR to a non-live branch, if you include this flag, it won't build anything, right?)

Would it be better to instead call this `--rebuild-live`?

jrodewig:
> Unrelated to the timing, I'm trying to think about how this interacts with the `--rebuild` option, especially on PRs in content repos. Does it ever make sense to include this flag if `--rebuild` isn't included? I imagine that could lead to unintuitive results (in a backport PR to a non-live branch, if you include this flag, it won't build anything, right?)

This is a good question. The intended use here is for a slimmed down use of the full `--rebuild`. I don't think that would be appropriate for more PRs, but we should account for it. I'll find some time to dig in.

> Would it be better to instead call this `--rebuild-live`?

Yes. Naming things is hard. Thanks for the suggestion!

jrodewig:
Moving this to draft while I investigate and work on feedback.

lcawl:
IIRC there were two reasons we needed to do a full rebuild: 1) the delta builds didn't always correctly determine the list of pages that needed to be refreshed, and (2) only a full rebuild updates the list of versions in the version selector. 

It seems like this PR helps with (1), which is great. For (2) this seems to be an ongoing issue until we figure out how to archive all the older versions of our docs.

jrodewig:
Thanks for the feedback everyone.

It seems like this may not be as useful as hoped, and I don't think I'll have time to tinker on it more any time soon.

If anyone is interested in reopening or reusing the code, feel free.

PR #2774, opened 2023-09-29
Build-docs jobs migrated to Buildkite

nkammah:
This run the docs jobs on merge to main (pushing to the `staging` branch of `elastic/built-docs`) and PR (pushing to docs_bk_${PR-NUM}). The build was tested on branches but the logic for master and PR only will be testable once this is merged. 

cc @bmorelli25

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -34,5 +34,7 @@ steps:
     agents:
       provider: "gcp"
       image: family/docs-ubuntu-2204
-
+notify:
+  - email: "olga.naidjonoka@elastic.co"
+  - slack: "@olga.naidjonoka"
 
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -26,7 +26,7 @@ steps:
     command: |
       export REBUILD="$(buildkite-agent meta-data get REBUILD --default false)"
       export ALLOW_BROKEN_LINKS="$(buildkite-agent meta-data get ALLOW_BROKEN_LINKS --default false)"
-      .buildkite/scripts/build-docs.sh
+      bash .buildkite/scripts/build-docs.sh
     #branches: "master"
     agents:
       provider: "gcp"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,5 +1,5 @@
 steps:
-  - label: "Build docs"
+  - label: ":white_check_mark: Build docs"
     command: .buildkite/build-docs.sh
     agents:
       provider: "gcp"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -22,6 +22,7 @@ steps:
           - label: "yes"
             value: true
         hint: "Should we ignore broken links? Ignoring broken links is dangerous not just because bad links will leak into the public site but because subsequent builds and pull requests that do not fix the links fail."
+  - wait
   - label: ":white_check_mark: Build docs"
     command: |
       export REBUILD="$(buildkite-agent meta-data get REBUILD --default false)"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -3,5 +3,4 @@ steps:
     command: .buildkite/build-docs.sh
     agents:
       provider: "gcp"
-        #image: family/docs-ubuntu-2204
-      image: "docs-ubuntu-2204-1695799161"
+      image: family/docs-ubuntu-2204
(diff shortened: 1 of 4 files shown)


nkammah:
fyi @bmorelli25

lib/ES/BaseRepo.pm
```diff
@@ -22,7 +22,8 @@ sub new {
         $reference_subdir =~ s|/$||;
         $reference_subdir =~ s|:*/*\.git$||;
         $reference_subdir =~ s/.*[\/:]//g;
-        $reference_dir = $args{reference}->subdir("$reference_subdir.git");
+        #$reference_dir = $args{reference}->subdir("$reference_subdir.git");
+        $reference_dir = $args{reference}->subdir("elastic-$reference_subdir");
```
gtback, line 26: If we merge this in as is, it will break existing builds (or at least really slow them down since it won't use any of the existing cached repos.

.buildkite/scripts/build-docs.sh
```diff
@@ -0,0 +1,56 @@
...
+fi
+
+# When running on a branch or on main
+if [[ "${BUILDKITE_PULL_REQUEST}" == "false" ]]; then
+  # temporary pushing to staging instead of master until the migration is over
+  build_args+=" --target_branch staging"
+else
+  build_args+=" --target_branch docs_bk_${BUILDKITE_PULL_REQUEST}"
+  build_args+=" --announce_preview https://docs_bk_${BUILDKITE_PULL_REQUEST}.docs-preview.app.elstc.co/diff"
+  rebuild_opt=" --rebuild"
```
gtback, line 44: I'm pretty sure don't want to do full rebuilds on PRs by default (though giving folks the option is probably a good thing). 

```suggestion
```

gtback (changes requested):

lib/ES/BaseRepo.pm
nkammah: oh yes - good catch! I forgot about that one!!!

oakrizan:
cla/check

jrodewig (approved):
LGTM. I didn't test, but the logic looks good.

jrodewig (approved):
LGTM. @nkammah set up a demo and I was able to rebuild docs using the Buildkite UI. Build parameters worked as intended. Thanks!

Screenshot 2023-10-05 at 12 21 31 PM

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,39 @@
...
+    command: |
+      export REBUILD="$(buildkite-agent meta-data get REBUILD --default false)"
+      export BROKEN_LINKS="$(buildkite-agent meta-data get BROKEN_LINKS --default false)"
+      bash .buildkite/scripts/build-docs.sh
+    branches: "master"
+    agents:
+      provider: "gcp"
+      image: family/docs-ubuntu-2204
+notify:
+  - email: "infra-root+build@elastic.co"
```
gtback, line 38: Should possibly use this instead.

```suggestion
  - email: "docs-status@elastic.co"
```

.buildkite/scripts/build-docs.sh
```diff
@@ -0,0 +1,55 @@
...
+build_args=""
+rebuild_opt=""
+broken_links_opt=""
+
+# From https://github.com/elastic/infra/blob/master/ci/jjb/elasticsearch-ci/defs/elastic-docs/pull-requests.yml#L110
+# Per https://github.com/elastic/docs/issues/1821, always rebuild all
+# books for PRs to the docs repo, for now.
+# When https://github.com/elastic/docs/issues/1823 is fixed, this
+# should be removed and the original behavior restored.
+#rebuild_opt="--rebuild"
```
gtback, line 22: Don't need this line 
```suggestion
```

.buildkite/scripts/build-docs.sh
```diff
@@ -0,0 +1,55 @@
...
+if [[ "${BROKEN_LINKS}" == 'skip' ]]; then
+  broken_links_opt="--skiplinkcheck"
+fi
+
+if [[ "${BROKEN_LINKS}" == 'warn' ]]; then
+  broken_links_opt="--warnlinkcheck"
+fi
+
+# When running on a branch or on main
+if [[ "${BUILDKITE_PULL_REQUEST}" == "false" ]]; then
```
gtback, line 37: Does this run when you push to a branch but don't create a PR? Or does the `branches: "master"` [line](https://github.com/elastic/docs/pull/2774/files#diff-a3991ebf1475eb82acab13946ffc1eca02b43917f6d5bec3898f45c8b0b9bd53R33)  in `pipeline.yml` prevent this?

We only want to use this for the `master` branch during the migration, and not have other builds fighting over this branch.

gtback (approved):
Overall, I'm OK with this, but we should avoid adding new message getting sent to the deprecated `infra-root` group.

.buildkite/scripts/build-docs.sh
nkammah: the branches: "master" line prevents it

PR #2779, opened 2023-10-04
Add Warn Broken Links build option

TrushynAliaksandr:

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -12,21 +12,23 @@ steps:
           - label: "yes"
             value: true
         hint: "Should all books be rebuilt, regardless of what has changed? Build once with this set to true after every release."
-      - select: "Allow Broken Links?"
-        key: "ALLOW_BROKEN_LINKS"
+      - select: "How should broken links be handled?"
+        key: "BROKEN_LINKS"
         default: false
         required: false
         options:
-          - label: "no"
+          - label: "Continue without warning"
+            value: "skip"
+          - label: "Continue, but log a warning"
+            value: "warn"
+          - label: "Fail the build"
             value: false
-          - label: "yes"
-            value: true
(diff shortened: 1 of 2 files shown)


.buildkite/pipeline.yml
```diff
@@ -12,21 +12,23 @@ steps:
...
         hint: "Should all books be rebuilt, regardless of what has changed? Build once with this set to true after every release."
-      - select: "Allow Broken Links?"
-        key: "ALLOW_BROKEN_LINKS"
+      - select: "How should broken links be handled?"
+        key: "BROKEN_LINKS"
         default: false
         required: false
         options:
-          - label: "no"
+          - label: "skip"
```
nkammah, line 20:
```suggestion
          - label: "Continue without warning"
```

.buildkite/pipeline.yml
```diff
@@ -12,21 +12,23 @@ steps:
...
+            value: "skip"
+          - label: "warn"
```
nkammah, line 22:
```suggestion
          - label: "Continue, but log a warning"
```

.buildkite/pipeline.yml
```diff
@@ -12,21 +12,23 @@ steps:
...
+            value: "warn"
+          - label: "fail the build"
```
nkammah, line 24:
```suggestion
          - label: "Fail the build"
```

.buildkite/scripts/build-docs.sh
```diff
@@ -13,6 +13,7 @@ export GIT_COMMITTER_EMAIL=$GIT_AUTHOR_EMAIL
 build_args=""
 rebuild_opt=""
 skiplinkcheck_opt=""
+warnlinkcheck_opt=""
```
nkammah, line 16: let's streamline a bit and use a single variable for the broken_links_opt that will either have the `--skiplinkcheck` or `--warnlinkcheck` value

TrushynAliaksandr: fixed

PR #2782, opened 2023-10-05
Remove x-pack directory dependency from docbldesx build alias

abdonpijpelink:
The Elasticsearch docs that used to live in the `x-pack` directory have moved (https://github.com/elastic/elasticsearch/pull/99209). We no longer need the dependency to that directory in the `docbldesx` build alias.

### DIFF
--- a/doc_build_aliases.sh
+++ b/doc_build_aliases.sh
@@ -10,7 +10,7 @@
 #
 
 # Elasticsearch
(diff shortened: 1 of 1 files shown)


leemthompo (approved):
Tested this and it works 👍

jrodewig (approved):
Good catch!

PR #2785, opened 2023-10-10
New pipeline for test execution

oakrizan:
To migrate Jenkins job with tests to Buildkite (issue [130](https://github.com/elastic/docs-projects/issues/130) new pipeline was created. It need to be pushed to master to continue with migrating scripts.

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -79,10 +79,9 @@ spec:
     metadata:
       name: pr-main-mr-test
     spec:
+      branch_configuration: "main"
       repository: elastic/docs
       pipeline_file: ".buildkite/pr_main_mr_test.yml"
-      provider_settings:
-        trigger_mode: none
       teams:
         ci-docs-migration-taskforce:
           access_level: MANAGE_BUILD_AND_READ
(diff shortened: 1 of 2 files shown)


catalog-info.yaml
```diff
@@ -57,3 +57,34 @@ spec:
...
+  implementation:
+    apiVersion: buildkite.elastic.dev/v1
+    kind: Pipeline
+    metadata:
+      name: pr-main-mr-test
+    spec:
+      repository: elastic/docs
+      pipeline_file: ".buildkite/pr_main_mr_test.yml"
+      provider_settings:
+        trigger_mode: none
```
nkammah, line 85: is this on purpose while developing the pipeline only, or a carry over ?

nkammah (approved):

catalog-info.yaml
nkammah: then maybe remove it so PRs and merges to main trigger builds.
You may want to set build_branches as false - see https://docs.elastic.dev/ci/getting-started-with-buildkite-at-elastic#2-adjust-your-pipeline-settings

oakrizan: I have added param "branch_configuration: "main"" to pipeline spec and removed the provider settings, since the default trigger is "code". As far as I understood "build_branches: false" will affect main branch as well, and it's not what we want in this case.

nkammah (approved):

PR #2789, opened 2023-10-11
Re organize docs pipelines

nkammah:
This PR renames so docs pipelines so that they're better organized in Buildkite (before they start getting used widely). Our original naive approach produced pipelines such as buildkite.com/elastic/preview-cleaner - which is not very descriptive or obvious.

This changeset also update some teams permissions to reflect the end-state.

cc @bmorelli25 for review on the teams permissions.

current looking state:

image

Note - this change will be fairly destructive from a terrazzo perspective and will require some manul tending from the CI team (I should also be able to help out with it).

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -22,8 +22,12 @@ spec:
       repository: elastic/docs
       pipeline_file: ".buildkite/pipeline.yml"
       teams:
+        docs-build-guild:
+          access_level: MANAGE_BUILD_AND_READ
         ci-docs-migration-taskforce:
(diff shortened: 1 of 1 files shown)


oakrizan (approved):

bmorelli25 (approved):
perms look good!

gtback (commented):
Changes look good to me, but I'd defer to the docs team's approvals.

nkammah:
Terrazzo applied in https://buildkite.com/elastic/terrazzo/builds/27070#018b201d-f956-4188-b526-540de835615d/6

PR #2791, opened 2023-10-12
Migrate build air-gapped job 132

TrushynAliaksandr:

### DIFF
--- a/.buildkite/air_gapped_pipeline.yml
+++ b/.buildkite/air_gapped_pipeline.yml
@@ -1,3 +1,8 @@
 steps:
-  - label: "Example Test"
-    command: echo "Hello World!"
+  - label: ":white_check_mark: Build air-gapped"
+    command: |
+      export AIR_GAPPED=docker.elastic.co/docs-private/bk_air_gapped:latest
+      bash .buildkite/scripts/publish_air_gapped_docs.sh
+    agents:
+      provider: "gcp"
+      image: family/docs-ubuntu-2204
--- a/.buildkite/preview_cleaner_pipeline.yml
+++ b/.buildkite/preview_cleaner_pipeline.yml
@@ -1,3 +1,6 @@
 steps:
-  - label: "Example Test"
-    command: echo "Hello World!"
+  - label: ":white_check_mark: Preview cleaning"
+    command: bash .buildkite/scripts//clean_preview_branches.sh
+    agents:
+      provider: "gcp"
+      image: family/docs-ubuntu-2204
--- a/.buildkite/scripts/clean_preview_branches.sh
+++ b/.buildkite/scripts/clean_preview_branches.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+set -eo pipefail
+
+set +x
+GITHUB_TOKEN=$(vault read -field=value secret/ci/elastic-docs/docs_preview_cleaner)
+export GITHUB_TOKEN
+set -x
+
+export REPO=git@github.com:elastic/built-docs.git
+export IMAGE=docker.elastic.co/docs/build:latest
+
+./build_docs --docker-build build
+ssh-agent bash -c '
+    ssh-add &&
+    docker run --rm \
+        -v $(pwd):/docs_build:cached,ro \
+        -v ~/.ssh/known_hosts:/root/.ssh/known_hosts:cached,ro \
+        -v $SSH_AUTH_SOCK:$SSH_AUTH_SOCK:cached,ro \
+        -e SSH_AUTH_SOCK=$SSH_AUTH_SOCK \
+        -e GITHUB_TOKEN=$GITHUB_TOKEN \
+        -v /opt/git-mirrors:/opt/git-mirrors:cached,ro \
+        -e CACHE_DIR=/opt/git-mirrors \
+        $IMAGE node /docs_build/preview/clean.js $REPO'
--- a/.buildkite/scripts/publish_air_gapped_docs.sh
+++ b/.buildkite/scripts/publish_air_gapped_docs.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+set -eo pipefail
+
+# Log in to docker.elastic.co.
+set +x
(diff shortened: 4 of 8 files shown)


.buildkite/scripts/publish_air_gapped_docs.sh
```diff
@@ -0,0 +1,16 @@
...
+set +x
+DOCKER_USERNAME=$(vault kv get -field=username secret/ci/elastic-docs/docker.elastic.co)
+DOCKER_PASSWORD=$(vault kv get -field=password secret/ci/elastic-docs/docker.elastic.co)
+source .buildkite/scripts/retry.sh
+retry 5 docker login -u $DOCKER_USERNAME -p $DOCKER_PASSWORD docker.elastic.co
+unset DOCKER_USERNAME
+unset DOCKER_PASSWORD
+set -x
+
+cd "$(git rev-parse --show-toplevel)"
```
nkammah, line 14: is this needed?

air_gapped/build_bk.sh
```diff
@@ -0,0 +1,26 @@
...
+
+cd $(git rev-parse --show-toplevel)
+
+if [[ ! -d /opt/git-mirrors/built-docs.git ]]; then
+  echo "/opt/git-mirrors/built-docs.git must exist and contain a reference clone of the built-docs repo"
+  exit 1
+fi
+
+# Get an up to date copy of the repo
+rm -rf air_gapped/work
```
nkammah, line 17: is this needed if the agent is clean?

.buildkite/scripts/publish_air_gapped_docs.sh
```diff
@@ -0,0 +1,16 @@
...
+source ./air_gapped/build_bk.sh
+docker push "$AIR_GAPPED"
```
nkammah, line 16: I know this was as in in the Jenkins world, but it's weird to push to $AIR_GAPPED where we don't see where it was sourced from. Maybe we should add AIR_GAPPED as either passed in or falling back on a default path? and same in the build script. If we actually do that, we can override the registry path in the top level job directly (in the pipeline.yml file) - that'll altogether be more readable and cleaner.

TrushynAliaksandr: i can move it to pipeline yaml it is make sense, but i don't think that we need to override registry path using ui. Does it make sense? just put export to pipeline.yaml file

air_gapped/build_bk.sh
TrushynAliaksandr: agree, I forgot that the agent is being cleaned

.buildkite/scripts/publish_air_gapped_docs.sh
nkammah: I was thinking of doing something like: 

```
AIR_GAPPED_REGISTRY_PATH="${AIR_GAPPED_REGISTRY_PATH:=docs-private/air_gapped}"
AIR_GAPPED=docker.elastic.co/$AIR_GAPPED_REGISTRY_PATH:latest 
```
in both the build.sh and publish.sh script, and simply adding the below in the pipeline.yaml

```
env:
   AIR_GAPPED_REGISTRY_PATH=docs-private/bk_air_gapped
```

TrushynAliaksandr: hmmm, i don't know, looks too much for me, maybe do something like:

.buildkite/preview_cleaner_pipeline.yml
```diff
@@ -1,3 +1,6 @@
 steps:
-  - label: "Example Test"
-    command: echo "Hello World!"
+  - label: ":white_check_mark: Preview cleaning"
+    command: bash .buildkite/scripts//clean_preview_branches.sh
```
nkammah, line 3:
```suggestion
    command: bash .buildkite/scripts/clean_preview_branches.sh
```

.buildkite/scripts/publish_air_gapped_docs.sh
nkammah: The intent is to allow folks to run the script locally without having it throw an error - but I guess we wouldn't want them to push to the production path anyway.

.buildkite/preview_cleaner_pipeline.yml
TrushynAliaksandr: done

.buildkite/air_gapped_pipeline.yml
```diff
@@ -1,3 +1,8 @@
 steps:
-  - label: "Example Test"
-    command: echo "Hello World!"
+  - label: ":white_check_mark: Build air-gapped"
+    command: |
```
nkammah, line 3: nit pick - but let's do the below

```
        command: ".buildkite/scripts/publish_air_gapped_docs.sh"
        env:
          AIR_GAPPED=docker.elastic.co/docs-private/bk_air_gapped:latest
```

.buildkite/scripts/clean_preview_branches.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+set -eo pipefail
+
+set +x
+GITHUB_TOKEN=$(vault read -field=value secret/ci/elastic-docs/docs_preview_cleaner)
```
nkammah, line 6:
```suggestion
export GITHUB_TOKEN=$(vault read -field=value secret/ci/elastic-docs/docs_preview_cleaner)
```

.buildkite/scripts/clean_preview_branches.sh
```diff
@@ -0,0 +1,24 @@
...
+export GITHUB_TOKEN
```
nkammah, line 7:
```suggestion
```

nkammah (approved):
added some minor nit-pick for cleanliness / readability - but lgtm

.buildkite/scripts/publish_air_gapped_docs.sh
TrushynAliaksandr: okay, let it be

PR #2803, opened 2023-10-25
Update prettify for ES|QL

abdonpijpelink:
Updates Prettify for ES|QL syntax highlighting with: the ENRICH command, all operators, prepositions (AS, ON, WITH), ASC/DESC, and all functions.

Before:
image

After:
image


nik9000 (approved):
We'll end up with a lot of these! Which is probably ok.

PR #2808, opened 2023-10-30
Add docs about `page_header.html`

bmorelli25:


leemthompo (approved):
Thanks for documenting this one. Personally I might name this section something like "Global banners", because that's closer to what I'd <kbd>Command</kbd> + <kbd>F</kbd> for if I was trying to find this...but that might be totally idiosyncratic 😄

README.asciidoc
```diff
@@ -1999,3 +1999,8 @@ This is where the content for tab #1 goes.
 This is the content for tab #2 goes.
 // end::reg-config[]
 ----
+
+[[Page-headers]]
+== Page headers
+
+Add a banner to every page within a book by creating a `page_header.html` file. This file must live in the root of the documentation directory (i.e. the book's `path` in `conf.yaml`) and is picked up by the doc build automatically. This file can include any valid HTML. An example can be seen https://github.com/elastic/apm-server/pull/6833/files[here].
```
bmorelli25, line 2006:
```suggestion
[[global-banners]]
== Global banners

Add a banner to every page within a book by creating a `page_header.html` file. This file must live in the root of the documentation directory (i.e. the book's `path` in `conf.yaml`) and is picked up by the doc build automatically. This file can include any valid HTML. An example can be seen https://github.com/elastic/apm-server/pull/6833/files[here].
```

PR #2813, opened 2023-11-03
Change destination for Search solution to Elasticsearch

chriscressman:
For the "Search" solution, direct readers to the Elasticsearch docs rather than the Enterprise Search docs.

For the link text, use the same text we use in the Elastic Cloud console to describe Elasticsearch projects.


chriscressman:

leemthompo (approved):
LGTM (not sure how to re-run the buildkite check)

PR #2817, opened 2023-11-07
Update PR bot config to trigger for non docs repos

nkammah:
The current PR bot config was assuming that only changes in the docs folder should trigger docs builds - but that's not the case for some repositories, such as:
* `elastic/docs`
* `elastic/tech-content`

This PR fixes that.

Also reported [here](https://github.com/elastic/docs-projects/issues/139#issuecomment-1798840531)


gtback (approved):
Makes sense to me.

PR #2823, opened 2023-11-08
Add versioning for Android APM agent docs

colleenmcginnis:
Fixes https://github.com/elastic/observability-docs/issues/3240

[Preview of the APM Android Agent docs](https://docs_2823.docs-preview.app.elstc.co/guide/en/apm/agent/android/current/intro.html)

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -1240,9 +1240,9 @@ contents:
                 sections:
                   - title:      APM Android Agent
                     prefix:     android
-                    current:    main
-                    branches:   [ {main: master} ]
-                    live:       [ main ]
+                    current:    0.x
+                    branches:   [ {main: master}, 0.x ]
+                    live:       [ 0.x ]
                     index:      docs/index.asciidoc
                     tags:       APM Android Agent/Reference
                     subject:    APM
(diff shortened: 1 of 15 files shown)


bmorelli25 (approved):
🚀

colleenmcginnis:
@LikeTheSalad can you confirm [which versions](https://www.elastic.co/guide/en/apm/guide/current/agent-server-compatibility.html) are compatible with the APM Android agent? It looks like the iOS agent goes back to 7.14. Is the Android agent the same?

LikeTheSalad:
> @LikeTheSalad can you confirm [which versions](https://www.elastic.co/guide/en/apm/guide/current/agent-server-compatibility.html) are compatible with the APM Android agent? It looks like the iOS agent goes back to 7.14. Does the Android agent go back as far as that?

It depends 😅 - If the support is only based on the APM server capabilities, then yes, we can go with the same value as iOS. However, if it also takes into account Kibana being aware of Android agents and displaying an Android icon when it finds one, then that work was done [here](https://github.com/elastic/kibana/pull/136598), which was part of version `8.4.0` of Kibana.

bmorelli25:
Thanks, @LikeTheSalad! Colleen, I think it makes the most sense to use the APM UI implementation (8.4.0) as our minimum supported version.

colleenmcginnis:
Thanks! I'll push an update then we should be good to gooo!

nkammah:
run docs-build

bmorelli25 (approved):

PR #2826, opened 2023-11-08
Re-organize the Buildkite pipeline

nkammah:
This PR introduces a new pipeline for building the docs PR, and update the test pipeline settings/naming.

We were using the same pipeline for commits to branches/master and for PR. The issue is that all PR builds are handled by the Buildkite PR bot, and this was introducing a lot of un-necessary complexity into the build script to support both mode (juggling the env variables). Additionally, some builds were kicked-off on commits to a branch of a PR interefering with the PR bot builds.


nkammah:
I expect the Buildkite builds to fail since Buildkite still expect the pipeline with the old name (while they've been renamed, but Terrazzo has not updated the Buildkite job config to inform it).

.buildkite/build_pipeline.yml
```diff
@@ -18,19 +18,18 @@ steps:
         required: false
         options:
           - label: "Continue without warning"
-            value: "skip"
+            value: "skiplinkcheck"
```
gtback, line 21: Do we need to make corresponding changes to the `build.sh` script, which is looking for the shorter vales?

.buildkite/build_pr_pipeline.yml
```diff
@@ -0,0 +1,10 @@
+steps:
+  - label: ":white_check_mark: Build docs PR"
+    command: |
+      bash .buildkite/scripts/build_docs_pr.sh
+    agents:
+      provider: "gcp"
+      image: family/docs-ubuntu-2204
+notify:
+  - email: "docs-status@elastic.co"
```
gtback, line 9: I don't think we ever want to notify via email on failed PRs (especially to content repos, but probably even to the `docs` repo). But that's up to the docs team.

catalog-info.yaml
```diff
@@ -36,6 +39,46 @@ spec:
...
+    apiVersion: buildkite.elastic.dev/v1
+    kind: Pipeline
+    metadata:
+      name: docs / build-pr
+    spec:
+      repository: elastic/docs
+      pipeline_file: ".buildkite/build_pr_pipeline.yml"
+      provider_settings:
+        trigger_mode: none # triggered via webhook and the buildkite-pr-bot
+        skip_intermediate_builds: false
```
gtback, line 67: I'm getting a YAML schema error locally on this line. I think it needs to be up one level (a sibling of `repository`).

catalog-info.yaml
```diff
@@ -36,6 +39,46 @@ spec:
         SLACK_NOTIFICATIONS_ALL_BRANCHES: "false"
         SLACK_NOTIFICATIONS_ON_SUCCESS: "false"
 
+---
+# yaml-language-server: $schema=https://gist.githubusercontent.com/elasticmachine/988b80dae436cafea07d9a4a460a011d/raw/rre.schema.json
+apiVersion: backstage.io/v1alpha1
+kind: Resource
+metadata:
+  name: buildkite-pipeline-docs-pr
+  description: Build the docs pull requests
```
gtback, line 48:
```suggestion
  description: Build the docs on pull requests to the `elastic/docs` repo.
```

gtback: I think the description of the `buildkite-pipeline-docs` pipeline could also be changed from "Build the Docs" to something like "Build and Publish the docs"

catalog-info.yaml
```diff
@@ -36,6 +39,46 @@ spec:
...
+        docs-build-guild:
+          access_level: MANAGE_BUILD_AND_READ
+        ci-docs-migration-taskforce:
+          access_level: MANAGE_BUILD_AND_READ
+        docs:
+          access_level: BUILD_AND_READ
+        everyone:
+          access_level: READ_ONLY
+      env:
+        ELASTIC_SLACK_NOTIFICATIONS_ENABLED: "true"
```
gtback, line 78: I don't know if we want Slack notifications when a build on a PR fails. That's up to the docs team, though.

catalog-info.yaml
```diff
@@ -100,11 +143,11 @@ spec:
...
     kind: Pipeline
     metadata:
-      name: docs-test
+      name: docs / test
     spec:
-      branch_configuration: "main"
+      branch_configuration: master
       repository: elastic/docs
-      pipeline_file: ".buildkite/pr_main_mr_test.yml"
+      pipeline_file: ".buildkite/test_pipeline.yml"
```
gtback, line 150: It doesn't look like the `.buildkite/scripts/test.sh` script that this pipeline calls is executable. Maybe Buildkite handles this automatically, but it might help to change this to explicitly use `bash`, like L4 of `.buildkite/build_pr_pipeline.yml`

gtback (commented):
A few things I noticed, but I'm far from a Buildkite expert.

.buildkite/build_pipeline.yml
nkammah: The final looking change to those values will come from the `buildkite-webhook-rebuild-136` branch, and it's using the longer values. The values are just set automatically by the buildkite PR bot, not entered by folks in their PR.

gtback (approved):
Changes make sense to me. I'm not sure how to verify this works without merging it in and running the various pipelines.

nkammah:
@bmorelli25 now that codeowners work, can you kindly review :D

nkammah:
> Changes make sense to me. I'm not sure how to verify this works without merging it in and running the various pipelines.

I reached the same conclusion

bmorelli25 (approved):

PR #2831, opened 2023-11-13
Adds serverless to landing page

KOTungseth:
DO NOT MERGE UNTIL 15-NOV 8AM ET 

Adds serverless docs to the landing page.

### DIFF
--- a/extra/docs_landing.html
+++ b/extra/docs_landing.html
@@ -218,44 +218,6 @@
     </dl>
   </details>
 
-  <h2>Learn about serverless projects</h2>
-  <p>Learn how to create a serverless project for an autoscaled and fully-managed Elastic solution based on Elasticsearch, Observability, and Security. </p>
-    <div class="row my-4">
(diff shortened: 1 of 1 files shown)


kilfoyle (dismissed):
LGTM. I think the changes look great!

dedemorton:
We had an discussion in slack about this. Here are the comments summaried:

DeDe said: "IMO putting it in such a prominent place on the page might send users down the serverless rabbit hole before they’re ready. It also introduces some ambiguity as to what the Get started section applies to. I would recommend finding a way to call serverless out as something new and different and not really production ready. I think that will involve doing something visually different so that serverless doesn’t blend into everything else."

@joepeeples said: "The solution-specific links might hasten them down that rabbit-hole, too — they’ll jump straight to solution docs without much explanation of what serverless is. Could we link to something more general instead, like  [Welcome to Elastic serverless](https://docs.elastic.co/serverless)?"

PR #2833, opened 2023-11-15
[do not merge] validating buildkite builds in docs repo

nkammah:

PR #2804, opened 2023-10-26
Buildkite PR job for assembling docs

TrushynAliaksandr:
This PR introduces a Buildkite job to build pull requests. It replaces and combine the following Jenkins job:
* product repositories job Jenkins Job [template](https://github.com/elastic/infra/blob/master/ci/jjb/elasticsearch-ci/globals/templates/elastic-docs/pull-requests.yml)
* elastic/docs Jenkins [PR job](https://github.com/elastic/infra/blob/master/ci/jjb/elasticsearch-ci/defs/elastic-docs/pull-requests.yml#L114-L190)

### General approach

In the Jenkins world, we setup a dedicated job per repository with the job dsl coming from a template. We took a different approach in Buildkite, and instead setup a single Buildkite job that gets triggered for every PR builds on any of the docs product repo or the elastic/docs repo. This allows us to keep things tidy in Buildkite ( a single job) and limit considerably the amount of CI code to write in any product repository (see discussion on this approach [here](https://github.com/elastic/docs-projects/issues/136#issuecomment-1763735623)).

There are some downsides to this approach however:
* going to Buildkite build page looking for a specific build for a PR is not straight forward - it can be done by adding GET params to the builds url. For eg, to view the builds PR 3366 in `elastic/observability-docs`, we can visit the following URL: https://buildkite.com/elastic/docs-build-pr/builds?meta_data[repo_pr]=observability-docs_3366

* Since the same job is used for any PR in any of the product repositories, we allow concurrent builds of that job (otherwise, we could only run a single doc build at a time for any of the product repo). I have not seen/experienced/validated this yet, but we could run into scenarios where the same PR with 2 docs builds triggered in a short time frame try to push to the same preview branch and <insert some problem>. 

### Docs repo configuration 

We use the [Buildkite PR bot](https://docs.elastic.dev/ci/configuring-pull-requests-for-buildkite) in a non-conventional manner to achieve the above result:
* we use an "org-wide" configuration file stored in the `master` branch of `elastic/docs` - [here](https://github.com/elastic/docs/blob/master/.buildkite/pull-requests.org-wide.json) (That file is not read in other branches).
* for all the products repositories, we leverage the `"always_trigger_branch":true` option. This is currently set to this branch while under development.
* for the `elastic:docs` repository, we use the default `"always_trigger_branch":false` option - since the build logic is stored within the `elastic:docs` repo - this is what would allow us to test out changes to the build logic itself.

[...]

PR #2835, opened 2023-11-16
Buildkite build PR job

nkammah:
(moved over from https://github.com/elastic/docs/pull/2804 since the previous one was started by a contractor no-longer at Elastic and I took over the work from there).

This PR introduces a Buildkite job to build pull requests. It replaces and combine the following Jenkins job:
* product repositories job Jenkins Job [template](https://github.com/elastic/infra/blob/master/ci/jjb/elasticsearch-ci/globals/templates/elastic-docs/pull-requests.yml)
* elastic/docs Jenkins [PR job](https://github.com/elastic/infra/blob/master/ci/jjb/elasticsearch-ci/defs/elastic-docs/pull-requests.yml#L114-L190)

### General approach

In the Jenkins world, we setup a dedicated job per repository with the job dsl coming from a template. We took a different approach in Buildkite, and instead setup a single Buildkite job that gets triggered for every PR builds on any of the docs product repo or the elastic/docs repo. This allows us to keep things tidy in Buildkite ( a single job) and limit considerably the amount of CI code to write in any product repository (see discussion on this approach [here](https://github.com/elastic/docs-projects/issues/136#issuecomment-1763735623)).

There are some downsides to this approach however:
* going to Buildkite build page looking for a specific build for a PR is not straight forward - it can be done by adding GET params to the builds url. For eg, to view the builds PR 3366 in `elastic/observability-docs`, we can visit the following URL: https://buildkite.com/elastic/docs-build-pr/builds?meta_data[repo_pr]=observability-docs_3366

* Since the same job is used for any PR in any of the product repositories, we allow concurrent builds of that job (otherwise, we could only run a single doc build at a time for any of the product repo). I have not seen/experienced/validated this yet, but we could run into scenarios where the same PR with 2 docs builds triggered in a short time frame try to push to the same preview branch and <insert some problem>. 

### Docs repo configuration 

We use the [Buildkite PR bot](https://docs.elastic.dev/ci/configuring-pull-requests-for-buildkite) in a non-conventional manner to achieve the above result:
* we use an "org-wide" configuration file stored in the `master` branch of `elastic/docs` - [here](https://github.com/elastic/docs/blob/master/.buildkite/pull-requests.org-wide.json) (That file is not read in other branches).
* for all the products repositories, we leverage the `"always_trigger_branch":true` option. This is currently set to this branch while under development.
* for the `elastic:docs` repository, we use the default `"always_trigger_branch":false` option - since the build logic is stored within the `elastic:docs` repo - this is what would allow us to test out changes to the build logic itself.

Finally, the Buildkite PR bot allows for changes detections at certain path - this allows us to not trigger a PR build for every change in every docs repo, but only target doc changes. 

With the above said, we have 3 configuration blobs in the PR bot config:
* **product docs** repo that should build for any change in the repo should be added [here](https://github.com/elastic/docs/blob/master/.buildkite/pull-requests.org-wide.json#L21-L24)
* **product docs** repo that should build only when files within the `./docs` folder should be added [here](https://github.com/elastic/docs/blob/master/.buildkite/pull-requests.org-wide.json#L48-L50)
* config changes to `elastic/docs` is controlled [here](https://github.com/elastic/docs/blob/master/.buildkite/pull-requests.org-wide.json#L53-L70)

The below config options have been configured for all of the above repositories:
* **run docs-build**
* **run docs-build** `rebuild`
* **run docs-build** `warnlinkcheck`
* **run docs-build** `skiplinkcheck` (takes precedence over the warnlinkcheck option)

### PR job setup

When a docs PR job is opened / commented on, the Buildkite PR bot will trigger a [docs-build-pr](https://buildkite.com/elastic/docs-build-pr) job, passing a few key env variables about the PR.  The build script [checks out](https://github.com/elastic/docs/pull/2804/files#diff-5ea8cc5083302320861aeb6d2dc6d73f688d94c1ed975a3c4bd1ad351cd23a74R36-R52) that specific branch of the repository locally and references it in the build argument leveraging the `--sub-dir` option.

Similarly to Jenkins, PR on the `elastic/docs` repo always trigger a full `rebuild`.

PR are annotated with a link to the job preview and diff upon successful build - see example [here](https://github.com/elastic/observability-docs/pull/3366#issuecomment-1813051017).

Similar to other Buildkite job, docs preview changes are pushed to the <repo>_bk_<PR> branch until the migration is complete.

### pre/post command hooks

We're leveraging theses the pre-command hook to setup the build commit status to `pending` and the post-commit one to `success` and `failure`. At this time, we're re-using the GITHUB token from the `preview-cleaner` job that has sufficient permissions to do such API calls.

All of the perms (to clone repo, commit) are tied to elasticmachine - I suspect this may have to change at some point in the future, but that's the current setup.

## Jenkins/Buildkite comparison

Example of jobs that ran, and the differences between Jenkins and Buildkite, and the diff between the built-docs branches.

* https://github.com/elastic/observability-docs/pull/3366 - [diff](https://github.com/elastic/built-docs/compare/observability-docs_bk_3366..observability-docs_3366)
* https://github.com/elastic/tech-content/pull/382 - [diff](https://github.com/elastic/built-docs/compare/tech-content_bk_382..tech-content_382)
* https://github.com/elastic/docs/pull/2833 - [diff](https://github.com/elastic/built-docs/compare/docs_bk_2833..docs_2833)
* https://github.com/elastic/observability-docs/pull/3373 - [diff](https://github.com/elastic/built-docs/compare/observability-docs_bk_3373..observability-docs_3373)

Note - on the docs changes in the `docs` repo, we're seeing images conflicts and changes - see convo about it [here](https://elastic.slack.com/archives/C05SZ9CGB19/p1699984533230979?thread_ts=1699973252.260879&cid=C05SZ9CGB19).


.buildkite/scripts/build_pr_setup.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# The reminder of this hook should only be invoked for builds triggered by the Buildkite PR bot
```
gtback, line 5:
```suggestion
# The remainder of this hook should only be invoked for builds triggered by the Buildkite PR bot
```

gtback:
> * Since the same job is used for any PR in any of the product repositories, we allow concurrent builds of that job (otherwise, we could only run a single doc build at a time for any of the product repo). I have not seen/experienced/validated this yet, but we could run into scenarios where the same PR with 2 docs builds triggered in a short time frame try to push to the same preview branch and .

This makes sense to me. I can see this causing problems if someone pushes to a PR, and then relatively quickly pushes a followup commit. At a minimum, it's worth documenting, but folks are probably used to Jenkins enough to know to just run a new build if they see anything strange. I imagine we could also check if any other builds are running on the same content repo and PR, and either abort the current one or cancel the other one, but this might require additional credentials to talk to Buildkite about other running builds.

.buildkite/build_pr_pipeline.yml
```diff
@@ -1,7 +1,18 @@
...
-      bash .buildkite/scripts/build_pr.sh
+  - key: "build-pr-setup"
+    label: "setup"
+    command: ".buildkite/scripts/build_pr_setup.sh"
+  - key: "build-pr"
+    label: ":hammer: Build docs PR"
+    command: ".buildkite/scripts/build_pr.sh"
+    depends_on:
+      - step: build-pr-setup
+        allow_failure: true
```
gtback, line 10: I was reading a bit about how Buildkite [step dependencies](https://buildkite.com/docs/pipelines/dependencies) work. Do we want to continue building the PR if the setup step fails? It seems like the script's primary purpose is to update the PR status checks to show this is in progress. 

It deliberately exits if not run by the Buildkite PR bot. It might be helpful to stop if that fails for whatever reason (but if we don't put a "pending" status check, is that really the worst thing?

.buildkite/scripts/build_pr_setup.sh
```diff
@@ -0,0 +1,21 @@
...
+
+gitHubToken=$(vault read -field=value secret/ci/elastic-docs/docs_preview_cleaner)
+
+githubPublishStatus="https://api.github.com/repos/${GITHUB_PR_OWNER}/${GITHUB_PR_REPO}/statuses/${GITHUB_PR_TRIGGERED_SHA}"
+data='{"state":"pending","target_url":"'$BUILDKITE_BUILD_URL'","description":"Build started.","context":"buildkite/'$BUILDKITE_PIPELINE_SLUG'"}'
+echo "Setting buildkite/docs commit status to pending"
+curl -s -L \
+  -X POST \
+  -H "Accept: application/vnd.github+json" \
+  -H "Authorization: Bearer ${gitHubToken}" \
```
gtback, line 18: There's not a chance of this getting logged anywhere, is there?

.buildkite/scripts/build_pr_teardown.sh
```diff
@@ -0,0 +1,27 @@
...
+  status_state="failure"
+fi
+
+githubPublishStatus="https://api.github.com/repos/${GITHUB_PR_OWNER}/${GITHUB_PR_REPO}/statuses/${GITHUB_PR_TRIGGERED_SHA}"
+data='{"state":"'$status_state'","target_url":"'$BUILDKITE_BUILD_URL'","description":"Build finished.","context":"buildkite/'$BUILDKITE_PIPELINE_SLUG'"}'
+echo "Setting buildkite/docs commit status to ${status_state}"
+curl -s -L \
+  -X POST \
+  -H "Accept: application/vnd.github+json" \
+  -H "Authorization: Bearer ${gitHubToken}" \
```
gtback, line 24: Ditto

.buildkite/scripts/build_pr_teardown.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
```
gtback, line 1: Given the similarities between the setup and teardown scripts, does it make sense to put the bulk of the logic in a separate file which can be included and called from both?

.buildkite/scripts/build_pr_setup.sh
```diff
@@ -0,0 +1,21 @@
...
+# The reminder of this hook should only be invoked for builds triggered by the Buildkite PR bot
+if [ -z ${GITHUB_PR_OWNER+set} ] || [ -z ${GITHUB_PR_REPO+set} ] || [ -z ${GITHUB_PR_TRIGGERED_SHA+set} ];then
+  exit 0
+fi
+
+gitHubToken=$(vault read -field=value secret/ci/elastic-docs/docs_preview_cleaner)
+
+githubPublishStatus="https://api.github.com/repos/${GITHUB_PR_OWNER}/${GITHUB_PR_REPO}/statuses/${GITHUB_PR_TRIGGERED_SHA}"
+data='{"state":"pending","target_url":"'$BUILDKITE_BUILD_URL'","description":"Build started.","context":"buildkite/'$BUILDKITE_PIPELINE_SLUG'"}'
+echo "Setting buildkite/docs commit status to pending"
```
gtback, line 14: Should this use `$BUILDKITE_PIPELINE_SLUG` (so, typically `buildkite/docs-build-pr`)? Same in the teardown.

.buildkite/scripts/build_pr.sh
```diff
@@ -1,2 +1,97 @@
...
+
+# Configure the git author and committer information
+export GIT_AUTHOR_NAME='Buildkite CI'
+export GIT_AUTHOR_EMAIL='docs-status+buildkite@elastic.co'
+export GIT_COMMITTER_NAME=$GIT_AUTHOR_NAME
+export GIT_COMMITTER_EMAIL=$GIT_AUTHOR_EMAIL
+
+# Set some metadata for build filtering capabilities
+# https://buildkite.com/elastic/docs-build-pr/builds?meta_data[repo_pr]=tech-content_123
+buildkite-agent meta-data set "repo_pr" "${GITHUB_PR_BASE_REPO}_${GITHUB_PR_NUMBER}"
```
gtback, line 19: Would it be helpful to set separate metadata for the repo name and the PR number (either instead of, or in addition to, this field) ? I can see that as being useful for checking "all the builds running on this repo", not just "all the builds that have run on this PR in this repo".

.buildkite/scripts/build_pr.sh
```diff
@@ -1,2 +1,97 @@
...
+rebuild_opt=""
+build_args=""
+TARGET_BRANCH=""
+
+# Define build docs arguments
+if [[ ${GITHUB_PR_COMMENT_VAR_REBUILD_OPT:="unset"} == "rebuild" ]];then
+  rebuild_opt=" --rebuild"
+elif [[ ${GITHUB_PR_COMMENT_VAR_SKIP_OPT:="unset"} == "skiplinkcheck" ]];then
+  build_args+=" --skiplinkcheck"
+  if [[ ${GITHUB_PR_COMMENT_VAR_WARN_OPT:="unset"} == "warnlinkcheck" ]];then
```
gtback, line 30: I haven't fully thought through how this works, but it doesn't seem like it's possible for `--warnlinkcheck` to get set unless `--skiplinkcheck` is also set, but I thought that `--skiplinkcheck` overrides `--warnlinkcheck`. Am I missing something?

.buildkite/scripts/build_pr.sh
```diff
@@ -1,2 +1,97 @@
...
+
+buildkite-agent \
+    annotate \
+    --style "info" \
+    --context 'docs-info' \
+    "Triggered by a doc change in elastic/$GITHUB_PR_BASE_REPO PR: [#$GITHUB_PR_NUMBER](https://github.com/elastic/$GITHUB_PR_BASE_REPO/pull/$GITHUB_PR_NUMBER)"
+
+
+if [[ "${GITHUB_PR_BASE_REPO}" != 'docs' ]]; then
+  # Buildkite PR bot for repositories other than the `elastic/docs` repo are configured to
```
gtback, line 43: Very helpful comment!

.buildkite/scripts/build_pr.sh
```diff
@@ -1,2 +1,97 @@
...
+  # always checkout the master branch of the `elastic/docs` repo (where the build logic resides).
+  # We first need to checkout the product repo / branch in a sub directory, that we'll reference
+  # in the build process.
+  echo "Cloning the ${GITHUB_PR_BASE_REPO} PR locally"
+
+  git clone --reference /opt/git-mirrors/elastic-$GITHUB_PR_BASE_REPO \
+    git@github.com:elastic/$GITHUB_PR_BASE_REPO.git ./product-repo
+
+  cd ./product-repo &&
```
gtback, line 52: An alternative might be `pushd` and `popd`, but this is fine too.

.buildkite/scripts/build_pr.sh
```diff
@@ -1,2 +1,97 @@
...
+      git switch $GITHUB_PR_BRANCH &&
+      cd ..
+
+  build_args+=" --sub_dir $GITHUB_PR_BASE_REPO:$GITHUB_PR_TARGET_BRANCH:./product-repo"
+else
+  # Buildkite PR bot for the `elastic/docs` repo is configured to checkout the PR directly into the workspace
+  # We don't have to do anything else in this case.
+
+  # Per https://github.com/elastic/docs/issues/1821, always rebuild all
+   # books for PRs to the docs repo, for now.
```
gtback, line 63: Inconsistent spacing before the comment here. I can just run `shellcheck` or `shfmt` on these scripts if that would help.

gtback (approved):
I really like how readable the `build_pr.sh` script ended up being, and could be used for PRs in both the docs repo and content repos!

I handful of questions and suggestions, but nothing that should block the merge. Some of the suggestions could also be handled in follow-up PRs.

.buildkite/build_pr_pipeline.yml
nkammah: I agree that missing the pending state is not the end of the world, and worse - the person opening the PR may not see that the pending state was not added, and assuming the build was started - so we may as well start it.

A case could be made to fail the build for the passing/failing state at the end.

.buildkite/scripts/build_pr_setup.sh
nkammah: I can't think of any other ways to make that call. we could add an explicit set +x just before it?

.buildkite/scripts/build_pr.sh
nkammah: so theses flags will be passed by the Buildkite PR bot based on the comment left on the PR - as configured [here](https://github.com/elastic/docs/blob/master/.buildkite/pull-requests.org-wide.json#L14):

```
"trigger_comment_regex": "run docs-build ?(?<rebuild_opt>rebuild)? ?(?<warn_opt>warnlinkcheck)? ?(?<skip_opt>skiplinkcheck)?",
```

gtback: What flags would you pass to get `--warnlinkcheck` to run? If you pass `GITHUB_PR_COMMENT_VAR_WARN_OPT=warnlink`, then `build_args` won't be modified unless you also pass `GITHUB_PR_COMMENT_VAR_SKIP_OPT=skiplinkcheck`. 

But you can't pass both, or the docs build [should raise an error](https://github.com/elastic/docs/blob/master/build_docs.pl#L1127-L1129).

.buildkite/scripts/build_pr_commit_status.sh
```diff
@@ -0,0 +1,33 @@
...
+      description='Build finished';;
+    *)
+      echo "Invalid state $status_state"
+      exit 1;;
+esac
+
+githubPublishStatus="https://api.github.com/repos/${GITHUB_PR_OWNER}/${GITHUB_PR_REPO}/statuses/${GITHUB_PR_TRIGGERED_SHA}"
+data='{"state":"'$status_state'","target_url":"'$BUILDKITE_BUILD_URL'","description":"'$description'","context":"buildkite/'$BUILDKITE_PIPELINE_SLUG'"}'
+
+echo "Setting buildkite/docs commit status to ${status_state}"
```
gtback, line 26:
```suggestion
echo "Setting ${BUILDKITE_PIPELINE_SLUG} commit status to ${status_state}"
```

gtback (approved):
I'm still confused about how to pass `--warnlinkcheck`, and have one other small suggestion, but I'll approve this so you can merge, in case I'm misunderstanding how the logic works.

scottybollinger (approved):

PR #2845, opened 2023-11-30
Cancel existing builds on running PRs

nkammah:
As noted in https://github.com/elastic/docs/pull/2835, the docs-pr job always trigger the master branch of the Buildkite job.  A PR in repo1 and another PR in repo 2 will both call the head code of the master branch of the elastic/docs and the job itself is configured to fetch the PR content. As a result, I had to disable cancelling intermediate builds since that would mean only one build at a time can run. The downside is that 2 commits in a row on the same PR will trigger a build twice for the same PR.

To tackle this, we introduce this PR that filters the existing running builds (by inspecting the build metadata to check if it came from a specific PR) and cancel the running builds.

Note:  while the [Buildkite API doc](https://buildkite.com/docs/apis/rest-api/builds) states that we can filter builds by meta_data, it does not work :( as a result, we have to run some jq shenanigans.

### DIFF
--- a/.buildkite/build_pr_pipeline.yml
+++ b/.buildkite/build_pr_pipeline.yml
@@ -1,4 +1,6 @@
 steps:
+  - key: "cancel-existing-builds"
+    command: ".buildkite/scripts/cancel_running_pr.sh || true"
   - key: "build-pr-setup"
     label: "setup"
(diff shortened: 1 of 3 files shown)


nkammah:
run docs-build

gtback (approved):
I haven't tested this, but it LGTM!

PR #2861, opened 2023-12-11
Remove duplicates from conf.yaml

lcawl:
This PR updates the conf.yaml to remove two books that were appearing in multiple sections (and thus potentially causing build problems).  It also updates the conf.yaml such that the Beats books are all grouped together.


bmorelli25 (approved):
Woo!

kilfoyle (approved):
LGTM. 🦕  
Thanks!!!

PR #2863, opened 2023-12-12
Jenkins/Buildkite branch comparison build step

nkammah:

### DIFF
--- a/.buildkite/build_pipeline.yml
+++ b/.buildkite/build_pipeline.yml
@@ -33,6 +33,14 @@ steps:
     agents:
       provider: "gcp"
       image: family/docs-ubuntu-2204
+   - wait
+   - key: "branch-comparison"
+     label: "Compare branches"
+     command: |
+       .buildkite/scripts/compare_bk_jenkins_branches.sh staging master
+     agents:
+       provider: "gcp"
+       image: family/docs-ubuntu-2204
 notify:
   - email: "docs-status@elastic.co"
     if: build.state == "failed"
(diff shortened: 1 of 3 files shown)


bmorelli25 (approved):
Looks neat! Thanks for putting this together. Is there any way to test it?

gtback (approved):
I haven't tried to run this, but it seems like it would be really useful 😁

bmorelli25:
Should the [Buildkite vs Jenkins diff](https://github.com/elastic/built-docs/compare/docs_bk_2863..docs_2863) be this big?

```
Showing 22,628 changed files with 4,737,324 additions and 6,996 deletions.
```

nkammah:
> Should the [Buildkite vs Jenkins diff](https://github.com/elastic/built-docs/compare/docs_bk_2863..docs_2863) be this big?
> 
> ```
> Showing 22,628 changed files with 4,737,324 additions and 6,996 deletions.
> ```

that's because I used this PR to test it against a bunch of other repos/PT :D

nkammah:
I will merge this is for now- there are other optimizations we could do, but I have already spent too much time on this for a script that will go away once the validation is over, and want to start reaping off the benefits of it while (potentially) incrementally making this better, eg:

* (optimization) check if the Jenkins branch exists and then sleep if it does not
* (better reporting) report violations in slack

thbkrkr:
> there are other optimizations we could do, but I have already spent too much time on this for a script that will go away once the validation is over,

Out of curiosity, how long will validation take? This script seems to increase the duration by almost +50% (from ~15min to ~24min). We have a PR build that takes ~15 minutes, so now we're waiting for the doc PR build, rather than ours.

PR #2865, opened 2023-12-13
Add rebuild instructions

nkammah:

### DIFF
--- a/.github/actions/docs-preview/action.yml
+++ b/.github/actions/docs-preview/action.yml
@@ -43,7 +43,7 @@ runs:
           * Rebuild this PR: \`run docs-build\`
           * Rebuild this PR and all Elastic docs: \`run docs-build rebuild\`
 
-          <sub>`run docs-build` is much faster than `run docs-build rebuild`. A `rebuild` should only be needed in rare situations. </sub>
+          <sub>\`run docs-build\` is much faster than \`run docs-build rebuild\`. A \`rebuild\` should only be needed in rare situations. </sub>
 
           <sub>If your PR continues to fail for an unknown reason, the doc build pipeline may be broken. Elastic employees can check the pipeline status [here](https://buildkite.com/elastic/docs-build).</sub>
           </details>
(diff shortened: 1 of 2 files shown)


nkammah:
@jmikell821 this PR rolls-in your change as well as more instructions (see the [last comment](https://github.com/elastic/docs/pull/2865#issuecomment-1853163622)). I used it to test your change.

.github/actions/docs-preview/action.yml
```diff
@@ -28,12 +28,28 @@ runs:
...
+
+          \`\`\`
+              run docs-build
+              run docs-build rebuild
+              run docs-build warnlinkcheck
+              run docs-build skiplinkcheck (takes precedence over the warnlinkcheck option)
+              buildkite test this
+              buildkite test this rebuild
+              buildkite test this warnlinkcheck
+              buildkite test this skiplinkcheck
```
bmorelli25, line 49: I have a few thoughts here.

1. Do we need to publicize both versions of the build commands? As I understand it, `run docs-build *` and `buildkite test this *` are identical. I think we should pick one set of commands to keep things simple.
2. I'm not convinced we should publicize the `warnlinkcheck` and `skiplinkcheck` options. These really shouldn't be used by most folks as they'll break the doc build if you merge a PR with broken links. These are almost exclusively for members of the docs release team in emergency merge situations.
3. I think we should explain the difference between `run docs-build` and `run docs-build rebuild`.

bmorelli25: With that being said. My suggestion would be something like this:

```
        <details>
            <summary>Request a new doc build by commenting</summary>

            * Rebuild this PR: `run docs-build`
            * Rebuild this PR and all Elastic docs: `run docs-build rebuild`

            > [!NOTE]  
            > `run docs-build` is much faster than `run docs-build rebuild`. A `rebuild` should only be needed in 
            rare situations. If your PR continues to fail for an unknown reason, the doc build pipeline may be 
            broken. Elastic employees can check the pipeline status [here](https://buildkite.com/elastic/docs-build).
          </details>
```

Is this too much info?

PR #2867, opened 2023-12-13
Add help to code snippets widget

leemthompo:
## 🎦 [Live preview](https://docs_bk_2867.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/current/getting-started.html)

Adds an ℹ️ button to our docs code example widget, just before the gear icon:

Screenshot 2023-12-13 at 18 10 45

When clicked it raises an plaintext alert box that explains how to use our code examples: 

Screenshot 2023-12-18 at 15 05 49

## Problem

Lots of folks don't know how our code samples widget works. This attempts to provide help text, without switching context by linking to docs.

### Issues

This is a quick and dirty solution. We could use a more involved modal to enable rich text, but I was able to put this together in less than an hour.

### `TODO`

- [x] Fix tests
  - [ ] CI is green but I haven't added any tests for my changes

### DIFF
--- a/resources/web/docs_js/components/console_widget.jsx
+++ b/resources/web/docs_js/components/console_widget.jsx
@@ -76,10 +76,11 @@ export const ConsoleWidget = props => {
 
   // Simple function to display a modal/tooltip
   const showHelp = () => {
-    alert('1. Use the toggle in the lower left-hand corner to change client languages.\n' +
-          '2. - Select "Copy as curl" to copy the code in curl format.\n' +
-          '    - Select "View in Console" to load example in Kibana Console.\n' +
-          '3. Use the gear icon (⚙) in the lower right-hand corner to configure your Elasticsearch host URL, username and Kibana Console URL.This ensures the code examples target your Elasticsearch instance.\n\n')
+    alert('1. To change the client language, use the dropdown.\n' +
+          '2. To copy the code in curl format, click "Copy as curl".\n' +
+          '3. To open the example in Kibana Console:
+                a. Click the gear icon (⚙) in the lower right-hand corner to configure your Elasticsearch host URL, username, and Kibana Console  URL. This ensures the code examples target your Elasticsearch instance.
+                b. Click "View in Console".\n' +
 };
 
 
--- a/resources/web/docs_js/components/console_widget.jsx
+++ b/resources/web/docs_js/components/console_widget.jsx
@@ -76,11 +76,11 @@ export const ConsoleWidget = props => {
 
   // Simple function to display a modal/tooltip
   const showHelp = () => {
-    alert('1. When code examples are available for different programming languages, use the toggle in the lower left-hand corner.\n\n' +
+    alert('1. Use the toggle in the lower left-hand corner to change programming language.\n\n' +
           '2. To copy a code snippet: \n\n' +
           '- Select "Copy as curl" to copy the code in curl format. \n' +
           '- Select "View in Console" to load the example into your Kibana Console.\n\n' +
-          '3. Use the gear icon in the lower right-hand corner to configure your Elasticsearch host URL, username and Kibana Console URL.\n\nThis ensures the code examples target your Elasticsearch instance.\n\n')
+          '3. Use the gear icon (⚙) in the lower right-hand corner to configure your Elasticsearch host URL, username and Kibana Console URL.\n\nThis ensures the code examples target your Elasticsearch instance.\n\n')
 };
 
 
@@ -103,7 +103,7 @@ export const ConsoleWidget = props => {
           href={`${props[props.setting + "_url"]}?load_from=${props.baseUrl}${props.snippet}`}>{props.langStrings(props.view_in_text)}</a>
       }
       
-<a className="help_info" onClick={showHelp} style={{ textDecoration: 'none', marginRight : '10px'}}>
+<a className="help_info" onClick={showHelp}>
   <strong>ℹ️</strong>
 </a>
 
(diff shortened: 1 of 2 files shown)


abdonpijpelink (commented):
This renders a little less beautiful for me (Chrome on MacOS).

leemthompo:
> This renders a little less beautiful for me (Chrome on MacOS).

Eugh 😅

leemthompo:
Screenshot 2023-12-14 at 12 17 49

@abdonpijpelink while we're looking at this, shouldn't `curl host` and `curl username` read Elasticsearch host/username?

leemthompo:
Screenshot 2023-12-14 at 12 34 15

@abdonpijpelink by simply cutting the text down and removing excessive newlines, we can get the dialog to open in Chrome without need to scroll

leemthompo:
I'm probably breaking every UI best practice under the sun here, but I feel like the current widget is broken for new users who don't understand it. Plus I don't know how to estimate the effort it would take to make this incrementally cleaner or whether it would be worth it...

bmorelli25 (changes requested):
Hi Liam. I appreciate you opening this PR, but I have some concerns with these changes.

My main concern (and not your fault because I'm not sure how widely this was communicated) is that we're not supposed to push updates to the legacy doc build at this time. We have working migration scripts for the transition to Docsmobile, and every update to the legacy build system has the potential to break those scripts.

My secondary concern is that I just don't know how useful this update is. I don't see anything in the new descriptions that adds additional context to what this widget is and what it does. The updates just seem to just be describing the UI but with more words. As an example, the button name is `Copy as curl`. I don't think `Select "Copy as curl" to copy the code in curl format.` really adds anything useful.

If we need to change text, like changing "Enter the URL of the Console editor" to "Enter the URL of the Kibana Console editor" or changing "curl host" to "Elasticsearch host", I think that makes a lot more sense than trying to describe these buttons in an info pop up.

leemthompo:
Independent of this PR, the bigger picture question is "How do we help folks use our code samples?" given this widget is actually pretty confusing to new users.

resources/web/docs_js/components/console_widget.jsx
```diff
@@ -73,6 +73,16 @@ export const ConsoleForm = connect((state, props) =>
 // ConsoleWidget isn't quite the right name for this any more....
 export const ConsoleWidget = props => {
   const modalAction = () => props.openModal(ConsoleForm, {setting: props.setting, url_label: props.url_label});
+
+  // Simple function to display a modal/tooltip
+  const showHelp = () => {
+    alert('1. Use the toggle in the lower left-hand corner to change client languages.\n' +
```
KOTungseth, line 79:
```suggestion
    alert('1. To change the client language, use the dropdown.\n' +
```

resources/web/docs_js/components/console_widget.jsx
```diff
@@ -73,6 +73,16 @@ export const ConsoleForm = connect((state, props) =>
...
+          '2. - Select "Copy as curl" to copy the code in curl format.\n' +
```
KOTungseth, line 80:
```suggestion
          '2. To copy the code in curl format, click "Copy as curl".\n' +
```

resources/web/docs_js/components/console_widget.jsx
```diff
@@ -73,6 +73,16 @@ export const ConsoleForm = connect((state, props) =>
...
+          '    - Select "View in Console" to load example in Kibana Console.\n' +
```
KOTungseth, line 81:
```suggestion
          '3. To open the example in Kibana Console:
                a. Click the gear icon (⚙) in the lower right-hand corner to configure your Elasticsearch host URL, username, and Kibana Console  URL. This ensures the code examples target your Elasticsearch instance.
                b. Click "View in Console".\n' +
```

resources/web/docs_js/components/console_widget.jsx
```diff
@@ -73,6 +73,16 @@ export const ConsoleForm = connect((state, props) =>
...
+          '3. Use the gear icon (⚙) in the lower right-hand corner to configure your Elasticsearch host URL, username and Kibana Console URL.This ensures the code examples target your Elasticsearch instance.\n\n')
```
KOTungseth, line 82:
```suggestion
```

KOTungseth (dismissed):
I made some tweaks. Thank you for creating this!

leemthompo:
https://github.com/elastic/docs/pull/2867/commits/06c9eedcdead2f158eaa4aab1c5bbbc8447010b0 tightens up text and makes it clearer that whether using curl or console you need to first configure variables using the "⚙"

chriscressman (commented):
I think the proposed UI is valuable in the following scenario: I click "Copy as curl" or "View in Console" and it doesn't work in some way. I see the information icon and click it to see if it helps me find a solution.

In this case, I suggest simpler text in the help dialog, something like:

> By default, "Copy as curl" and "View in Console" use localhost. Configure your Elasticsearch and Kibana locations using the gear icon (⚙).

Then, I would improve the labels in the config form:

> Kibana Console URL for "View in Console"
> Elasticsearch host for "Copy as curl"
> Elasticsearch username for "Copy as curl"

leemthompo:
> I think the proposed UI is valuable in the following scenario: I click "Copy as curl" or "View in Console" 

We also have to account for folks who don't know what Console is, and are just confused or assume it means terminal.

> Then, I would improve the labels in the config form:

Agree but this messes with tests, so I didn't pull on that thread

leemthompo:
Closing as we're not supposed to push updates to the legacy doc build at this time

PR #2900, opened 2024-01-18
Add Solution get started links

lcawl:
This PR adds "Get started" links to the Solution cards on the landing page.

### DIFF
--- a/extra/docs_landing.html
+++ b/extra/docs_landing.html
@@ -434,7 +434,6 @@
   <div role="tabpanel" id="solution-tab-contents" aria-labelledby="solution-tab" tabindex="0">
     <div class="row">
       <div class="col-md-4 col-12 mb-2">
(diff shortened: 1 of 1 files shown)


dedemorton:
@lcawl I pushed a commit to your branch that changes the cursor behavior, but it will affect all the cards. The pointy finger thingy only appears when you highlight a link.

This is probably the easiest fix, but might not be what you're looking for.

The other option would probably be to define a new card style for cards that have multiple links, but my feeling is that would be harder to maintain. Plus I think it's better for the cursor to behave the same across cards. I think we should either use the default cursor...or just accept the fact that cursor is misleading on the cards that have more than one link.

Feel free to remove my commit if you don't like the change!

chriscressman (approved):

joepeeples (approved):

PR #2904, opened 2024-01-25
Pass the --update-head-ok to fetch for PRs out of main

nkammah:
It seems like docs PR coming from a local main branch fail to checkout on Buildkite with the error:
`fatal: Refusing to fetch into current branch refs/heads/main of non-bare repository`

This PR attempts solves the issue by using the `--update-head-ok` flag - see explanation in this [stack overflow](https://stackoverflow.com/questions/2236743/git-refusing-to-fetch-into-current-branch).

Note - this also addresses the issue where status checks were not reported when the build was failing

This was tested using one of the [problematic PR](https://github.com/elastic/apm-agent-android/pull/265):
* reproducing the failure
* with the fix (additional flag) (but then aborting the build)

### DIFF
--- a/.buildkite/build_pr_pipeline.yml
+++ b/.buildkite/build_pr_pipeline.yml
@@ -24,6 +24,7 @@ steps:
       fi
     depends_on:
       - step: "build-pr"
+        allow_failure: true
   - wait
   - key: "branch-comparison"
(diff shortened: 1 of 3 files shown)


bmorelli25 (approved):
Awesome. Thanks!

PR #2927, opened 2024-02-08
Build docs hourly - every 30 min is running into race conditions

nkammah:


nkammah:
FYI - @bmorelli25 : i'm gong to go ahead with this since the master pipeline is consistently failing (while it was not when we were on staging).

PR #2931, opened 2024-02-09
Set build type for scheduled builds based on doc repo changes

bmorelli25:
### Summary

In https://github.com/elastic/docs/pull/2929, we reverted to incremental builds run every 30 minutes. This works great, until a PR is merged to the elastic/docs repository. Almost every PR merged in this repo requires a `rebuild` in order for the changes to take effect on `elastic.co/guide/*`.

This PR updates our scheduled builds to first check to see if the build commit of `elastic/docs` has changed since the previous successful, scheduled run. If the commit **has** changed, we run a full docs rebuild to ensure the elastic/docs changes are propagated to all corners of elastic.co/guide. If the commit **hasn't** changed, there's no need for a full rebuild, and we instead run an incremental build—the faster and cheaper option.

### Testing

Testing this PR is fun and easy!

- You can run the bash script locally.
- You can kick off a BK build here: https://buildkite.com/elastic/docs-build/builds?branch=test-conditional-builds. Note, you must use the penultimate commit: [`c3e8250`](https://github.com/elastic/docs/pull/2931/commits/c3e82508a410a230ce278562ba07ae209d657266). Here's the successful run: https://buildkite.com/elastic/docs-build/builds/3589#018d8ce5-6265-4b9d-a8ff-49e0b8c4612a.

PR #2932, opened 2024-02-09
Update concurrency rules and limit build queue to one

bmorelli25:
### Summary

This PR builds on https://github.com/elastic/docs/pull/2931.

#### Goal 1: Set concurrency at the pipeline level, not the step level.

Pipeline-level concurrency isn't a feature of BK: https://forum.buildkite.community/t/pipeline-concurrency-limit/240. Instead, you have to set concurrency limits on each step in a build within a pipeline. In this implementation, I use a [concurrency gate](https://buildkite.com/blog/concurrency-gates) as it seems to be the cleanest implementation.

In the following screenshot you can see how the concurrency gate prevents new builds from starting the "Build the docs" step and instead holds them at the "Start concurrency gate" step:

Screenshot 2024-02-09 at 2 19 21 PM

#### Goal 2: Set a queue limit of one running build and one queued build.

The downside of concurrency limits paired with scheduled builds is that the queue for builds will grow if/when we have a `rebuild`. This happens because our `rebuild` build time is way longer (70 mins) than our scheduled build time (30 mins). 

To fix this problem, this PR adds a new first BK build step: check not this build and not the previous build, but the third build back to determine if it is still in a `running` state. If it is, it means we already have a BK build in the queue. We can then safely cancel the current build to prevent the queue from growing. See [this diagram](https://buildkite.com/docs/pipelines/defining-steps#build-states) for more on build states.

Because the `"Full rebuild or incremental build?"` step is concurrency gated, we will only determine the type of build after the previous build has been unblocked. This prevents this step from breaking the functionality added in https://github.com/elastic/docs/pull/2931.

### Testing

Testing this PR is fun and easy. Go here and create a new build: https://buildkite.com/elastic/docs-build/builds?branch=move-concurrency. Wait until an agent is assigned. Once an agent is assigned, create another new build. Repeat this process until you have three builds running.

#### Build 1
The first build should do the following in the **Check for queue** step:
```
2024-02-09 13:14:08 PST | Determining if there are multiple builds waiting.
2024-02-09 13:14:08 PST | The pipeline is ready for a new build.
```
The first build will then run the **✅ Build Docs** step.

#### Build 2
The second build should do the following in the **Check for queue** step:
```
2024-02-09 13:16:08 PST | Determining if there are multiple builds waiting.
2024-02-09 13:16:08 PST | The pipeline is ready for a new build.
```
The second build should say "Waiting on currency group" in the **Start of concurrency** step:
Screenshot 2024-02-09 at 1 50 00 PM

#### Build 3
The third build should do the following in the **Check for queue** step:
```
2024-02-09 22:17:07 UTC | Determining if there are multiple builds waiting.
2024-02-09 22:17:07 UTC | The pipeline is congested. Canceling this build.
```
The third build will then cancel itself. Example: https://buildkite.com/elastic/docs-build/builds/3638

Screenshot 2024-02-09 at 2 17 35 PM

#### Build 2

Eventually, Build 1 will complete. This will open the concurrency gate and allow Build 2 to enter the **✅ Build Docs** step.

### DIFF
--- a/.buildkite/build_pipeline.yml
+++ b/.buildkite/build_pipeline.yml
@@ -24,6 +24,10 @@ steps:
           - label: "Fail the build"
             value: ""
         hint: "Should we ignore checking broken links? Should we allow to run the build without failing if there's a broken link? Ignoring broken links is dangerous not just because bad links will leak into the public site but because subsequent builds and pull requests that do not fix the links fail."
+  - wait
+  - label: "Check for queue"
+    if: build.source == "schedule"
+    command: ".buildkite/scripts/check_queue.sh"
     concurrency: 1
     concurrency_group: docs-build-${BUILDKITE_BRANCH}
   - wait
--- a/.buildkite/build_pipeline.yml
+++ b/.buildkite/build_pipeline.yml
@@ -1,6 +1,4 @@
 steps:
-    concurrency_group: docs-build-${BUILDKITE_BRANCH}
-    concurrency: 1
   - input: "Build parameters"
     if: build.source == "ui"
     fields:
@@ -26,10 +24,13 @@ steps:
           - label: "Fail the build"
             value: ""
         hint: "Should we ignore checking broken links? Should we allow to run the build without failing if there's a broken link? Ignoring broken links is dangerous not just because bad links will leak into the public site but because subsequent builds and pull requests that do not fix the links fail."
+    concurrency: 1
+    concurrency_group: docs-build-${BUILDKITE_BRANCH}
   - wait
   - label: "Full rebuild or incremental build?"
     if: build.source == "schedule"
     command: ".buildkite/scripts/compare_commits.sh"
+    concurrency_group: docs-build-${BUILDKITE_BRANCH}
   - label: ":white_check_mark: Build docs"
     command: |
       export REBUILD="$(buildkite-agent meta-data get REBUILD --default '' --log-level fatal)"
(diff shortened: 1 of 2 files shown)


nkammah:
I could be wrong but this still looks to be set at a step level, not pipeline - but i'm not too sure how that's supposed to work, best thing would be to try it out.

.buildkite/build_pipeline.yml
```diff
@@ -1,33 +1,45 @@
...
+  #       required: false
+  #       options:
+  #         - label: "Continue without warning"
+  #           value: "skiplinkcheck"
+  #         - label: "Continue, but log a warning"
+  #           value: "warnlinkcheck"
+  #         - label: "Fail the build"
+  #           value: ""
+  #       hint: "Should we ignore checking broken links? Should we allow to run the build without failing if there's a broken link? Ignoring broken links is dangerous not just because bad links will leak into the public site but because subsequent builds and pull requests that do not fix the links fail."
+  # - wait
```
bmorelli25, line 27: Note for reviewer: All of this is commented out so we can test this PR. Once a reviewer has tested this PR, I'll uncomment this out and we can merge.

nkammah (approved):
It seems to work as designed - but I have not tested nor thought deeply of edge cases.

colleenmcginnis:
>Testing this PR is fun and easy. Go here and create a new build: https://buildkite.com/elastic/docs-build/builds?branch=move-concurrency. Wait until an agent is assigned. Once an agent is assigned, create another new build. Repeat this process until you have three builds running.

@bmorelli25 when I tested this, both [build 2](https://buildkite.com/elastic/docs-build/builds/5009) and [build 3](https://buildkite.com/elastic/docs-build/builds/5010) were canceled:

```
Determining if there are multiple builds waiting.
The pipeline is congested. Canceling this build.
```

[Build 1](https://buildkite.com/elastic/docs-build/builds/5007) started running as expected.

PR #2946, opened 2024-02-23
increase agent disk for pr builds

jonahbull:
Increase agent disk for PR builds from 100GB to 150GB, we were starting to run out of space.

Add `allow_failure` to `depends_on` stanza so the teardown step runs and updates the PR's status even if the previous step fails.


kilfoyle (approved):
Thanks for this @jonahbull!

PR #2947, opened 2024-02-28
Move ingest reference architectures up in the ingest section

dedemorton:
Received a request from @karenzone to move the reference architecture docs up in the ingest section because the content is fundamental to choosing which ingest method to use. I think this is OK since the section is not alphabetized anyhow.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -1465,6 +1465,22 @@ contents:
 
     -   title:     "Ingest: Add your data"
         sections:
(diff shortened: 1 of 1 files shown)


karenzone (approved):
LGTM! Thanks

glitteringkatie (approved):

dedemorton:
Had to fix a conflict. Can I get a code owner review please?

leemthompo (approved):

PR #2951, opened 2024-03-06
Increase agent disk for docs builds

joepeeples:
Increase VM disk space for docs build process to 150GB, to prevent intermittent breaks where the build runs out of space.

Follows a similar fix from https://github.com/elastic/docs/pull/2946.


kilfoyle (approved):
Thanks for fixing this @joepeeples!

PR #2954, opened 2024-03-08
Adds API reference page to top nav on landing page

KOTungseth:
Adds `API reference` to the top nav so users can easily access the single location for all Elastic API documentation.


kilfoyle (approved):
LGTM! 🌮

PR #2955, opened 2024-03-08
Add support for Elasticsearch tab delimited links file

rjernst:
The Elasticsearch reference docs links file formatted in json. That format makes extraction of links complicated, both on the docs side, and also in Elasticsearch needing to parse json. This commit adds support for a new tab delimited txt file where the Elasticsearch reference docs will move to.

relates elastic/elasticsearch#105813

### DIFF
--- a/build_docs.pl
+++ b/build_docs.pl
@@ -454,13 +454,22 @@ sub check_elasticsearch_links {
     # So we grab all quoted strings that contain `html`. This *should* be fine
     # for a while because the keys in the file are all in SHOUTING_SNAKE_CASE
     # so even if one contains "html" it'll contain "HTML" which doesn't match.
-    my $extractor = sub {
+    my $json_extractor = sub {
         my $contents = shift;
         return sub {
             while ( $contents =~ m!"([^"\#]+)(?:\#([^"]+))?"!g ) {
                 my $path = $1;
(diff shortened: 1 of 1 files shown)


lcawl:
While looking at [reference-docs-links.json](https://github.com/elastic/elasticsearch/blob/main/server/src/main/resources/org/elasticsearch/common/reference-docs-links.json) and [ReferenceDocs.java](https://github.com/elastic/elasticsearch/blob/main/server/src/main/java/org/elasticsearch/common/ReferenceDocs.java), I noticed that it will need to be updated to accommodate the new style of URLs as content gets migrated to docsmobile.  It might also need to start considering whether the URL is being used in a serverless context once we start creating more serverless-specific Elasticsearch docs.  They implemented serverless "flavours" in the Kibana [doc link service](https://github.com/elastic/kibana/blob/main/packages/kbn-doc-links/src/get_doc_links.ts) not too long ago, so that might be a good reference.

lcawl (approved):

rjernst:
Thanks @lcawl. I'll be following up in ES to move `ReferenceDocs.java` and convert `reference-docs-links.json` to `reference-docs-links.txt`. Happy to make any other changes you suggest there in the process.

PR #2934, opened 2024-02-09
🚧 [WIP] Remove attributes with case-sensitive names

joepeeples:
🚧 WORK IN PROGRESS — More details to come soon. Please hold off on reviewing or commenting.

---
Contributes to https://github.com/elastic/security-docs/issues/4754 by removing attributes with case-sensitive names/keys. AsciiDoc [doesn't actually](https://docs.asciidoctor.org/asciidoc/latest/attributes/names-and-values/) support uppercase attribute names, and these legacy attributes are causing incorrect capitalization of the term "data view" throughout the docs.

### Previews
[add pages to test]

### Related
- https://github.com/elastic/docs.elastic.co/pull/205 — Removes equivalent variables from docs.elastic.co. Those variables aren't causing trouble because Docsmobile seems to support uppercase variable names, but they're still legacy variables that we shouldn't encourage anyone to use.

PR #2963, opened 2024-03-18
Remove duplicate data view attributes

lcawl:
Relates to https://github.com/elastic/docs/pull/2778, https://github.com/elastic/kibana/pull/178759

Per https://docs.asciidoctor.org/asciidoc/latest/attributes/names-and-values/ "Although uppercase characters are permitted in an attribute name, the name is converted to lowercase before being stored. For example, URL-REPO and URL-Repo are treated as url-repo when a document is loaded or converted. A best practice is to only use lowercase letters in the name and avoid starting the name with a number".  Therefore the "legacy" attributes like `Data-Sources` are considered to be the same as `data-sources` and since the former appears later in the attributes file, its value is used in pages such as:

https://www.elastic.co/guide/en/kibana/current/introduction.html

image

https://www.elastic.co/guide/en/kibana/current/dashboard.html

image

This PR therefore removes the problematic case-sensitive attribute names.

### DIFF
--- a/shared/attributes.asciidoc
+++ b/shared/attributes.asciidoc
@@ -538,6 +538,4 @@ Legacy definitions
 :apm-overview-ref-70:  https://www.elastic.co/guide/en/apm/get-started/7.0
 :apm-overview-ref-m:   https://www.elastic.co/guide/en/apm/get-started/master
 :infra-guide:          https://www.elastic.co/guide/en/infrastructure/guide/{branch}
-// Do not use A-data-source or a-data-source
-:A-data-source:      A data view
-:a-data-source:      a data view
+:a-data-source:        a data view
--- a/shared/attributes.asciidoc
+++ b/shared/attributes.asciidoc
@@ -249,7 +249,9 @@ Kibana app and UI names
 :data-sources-caps:  Data Views
 :data-source-cap:    Data view
 :data-sources-cap:   Data views
-
+// Use data-source-cap instead of Data-source
+// Use data-sources-cap instead of Data-sources
+// Use data-sources-caps instead of Data-Sources
 
 //////////
 Enterprise Search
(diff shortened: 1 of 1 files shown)


shared/attributes.asciidoc
```diff
@@ -536,12 +538,6 @@ Legacy definitions
...
 :infra-guide:          https://www.elastic.co/guide/en/infrastructure/guide/{branch}
-// Use data-sources-caps instead of Data-Sources
-:Data-Sources:         Data Views
-// Use data-source-cap instead of Data-source
-:Data-source:          Data view
-// Use data-sources-cap instead of Data-sources
-:Data-sources:         Data views
 // Do not use A-data-source or a-data-source
 :A-data-source:      A data view
 :a-data-source:      a data view
```
joepeeples, line 543: Should we also remove these case-sensitive attributes, or cut one of them to avoid duplicates? I don't know how widely they're used, or if they're causing problems like the other ones were.

joepeeples (approved):
Thanks for opening the PR, @lcawl! I had the same thing on my back burner in https://github.com/elastic/docs/pull/2934 but never fully opened it (got distracted by EAH then 8.13 release).

I added a question about whether we should keep `:A-data-source:` and `:a-data-source:` since they're also duplicates, but will give this an approval to speed things along if no further changes are needed.

joepeeples:
Oh, one other thing: we still have the equivalent [case-sensitive variables](https://github.com/elastic/docs.elastic.co/blob/8c7fb3abfca1b9c274a7ad4a3c14ebf08be32c0c/config/variables/terms.js#L180-L190) in the serverless docs at docs.elastic.co. 

```
  "Data-Sources": "Data Views",
  "Data-source": "Data view",
  "data-source": "data view",
  "Data-sources": "Data views",
  "data-sources": "data views",
  "data-source-caps": "Data View",
  "data-sources-caps": "Data Views",
  "data-source-cap": "Data view",
  "data-sources-cap": "Data views",
  "A-data-source": "A data view",
  "a-data-source": "a data view",
```

They aren't causing trouble like the AsciiDoc attributes are, but we should probably update them so they match in both systems. Again, I have a draft PR for that, we can clean it up and open it if you want to use that: https://github.com/elastic/docs.elastic.co/pull/205

shared/attributes.asciidoc
lcawl: "a-data-source" is still used in some versions (e.g. 8.11 and earlier) so I think you're right that we can remove the "A-data-source" but keep the other.

lcawl: Fixed in b42131d62d2e8ec00e690170d1132398539dfea6

joepeeples (approved):
LGTM! I also updated https://github.com/elastic/docs.elastic.co/pull/205 to keep `a-data-source` in case it's needed on the serverless/Docsmobile side.

PR #2965, opened 2024-03-21
Remove z-index for edit button to prepare for incoming nav refactor

johnbarrierwilson:
We have a new navigation refactor coming in. Despite setting the z-index for the dropdowns in the new navigation to larger than 1000 (as seen in this PR), the browsers still didn't respect the enforcement. This is also a better long term way as the z-index isn't necessary on this element. In addition, the main bug here is that `opacity: 0.3` created a new stacking order that broke any z-index declarations.

### DIFF
--- a/resources/web/style/heading.pcss
+++ b/resources/web/style/heading.pcss
@@ -62,7 +62,6 @@
     position: absolute;
     font-size: 14px;
     top: 0;
-    z-index: 1000;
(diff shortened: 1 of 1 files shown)


glitteringkatie (approved):

bmorelli25:
Hey @johnbarrierwilson. Thanks for the PR. Can we merge this?

PR #2985, opened 2024-04-12
Update the aliases file for removed APM guide

kilfoyle:
When I attempt to build the Installation and Upgrade Guide using `docbldstk` it returns:

 `Can't find --resource /Users/kilfoyle/repos/apm-server/docs/guide`

It's just a guess, but maybe this path was removed as part of https://github.com/elastic/apm-server/pull/12341

Anyway, removing the `--resource=$GIT_HOME/apm-server/docs/guide` part of the alias for `docbldstk` has the Install and Upgrade Guide building nicely again. 🎉

### DIFF
--- a/doc_build_aliases.sh
+++ b/doc_build_aliases.sh
@@ -58,7 +58,7 @@ docbldlsvpr() {
 }
 
 # Installation and Upgrade Guide 7.10 and later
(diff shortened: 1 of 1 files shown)


bmorelli25 (approved):
Wow, no one has tried to build this book locally in a long time. It looks like this hasn't been needed since https://github.com/elastic/docs/pull/2760. Thanks!

PR #2993, opened 2024-04-29
Update doc_build_aliases.sh

kilfoyle:
For the [Integrations Developer Guide](http://localhost:8000/guide/index.html):
 - In `conf.yaml` we use `chunk: 1`
 - In  `doc_build_aliases` we use `chunk: 2`
 
This just makes the two versions consistent, which should make things a tad less confusing for external contributors and also for yours truly. :-) 

conf.yaml:
```
          - title:      Integrations Developer Guide
            prefix:     en/integrations-developer
            current:    main
            branches:   [ {main: master} ]
            live:       [ main ]
            index:      docs/en/integrations/index.asciidoc
            chunk:      1
            tags:       Integrations/Developer
            subject:    Integrations
            sources:
```


KOTungseth (approved):
LGTM!

PR #2996, opened 2024-05-02
Change ECE current docs version

lcawl:
This PR updates the "current" branch for the Elastic Cloud Enterprise docs in the conf.yaml 

The PR currently fails due to:

````

INFO:build_docs:Bad cross-document links:
--
  | INFO:build_docs:  /tmp/docsbuild/target_repo/html/en/cloud-enterprise/3.7/ece-migrate-ccs.html contains broken links to:
  | INFO:build_docs:   - en/cloud-enterprise/current/ece-remote-cluster-other-ece.html
  | INFO:build_docs:  /tmp/docsbuild/target_repo/html/en/cloud-enterprise/3.7/ece-remote-cluster-other-ece.html contains broken links to:
  | INFO:build_docs:   - en/cloud-enterprise/current/ece-remote-cluster-ece-ess.html
  | INFO:build_docs:  /tmp/docsbuild/target_repo/html/en/cloud-enterprise/3.7/ece-remote-cluster-same-ece.html contains broken links to:
  | INFO:build_docs:   - en/cloud-enterprise/current/ece-remote-cluster-same-ece.html
  | INFO:build_docs:  /tmp/docsbuild/target_repo/html/en/cloud/current/ec-enable-ccs.html contains broken links to:
  | INFO:build_docs:   - en/cloud-enterprise/current/ece-remote-cluster-ece-ess.html
  | INFO:build_docs:  /tmp/docsbuild/target_repo/html/en/cloud/current/ec-remote-cluster-ece.html contains broken links to:
  | INFO:build_docs:   - en/cloud-enterprise/current/ece-remote-cluster-ece-ess.html
  | INFO:build_docs:  /tmp/docsbuild/target_repo/html/en/cloud/latest/ec-enable-ccs.html contains broken links to:
  | INFO:build_docs:   - en/cloud-enterprise/current/ece-remote-cluster-ece-ess.html
  | INFO:build_docs:  /tmp/docsbuild/target_repo/html/en/cloud/latest/ec-remote-cluster-ece.html contains broken links to:
  | INFO:build_docs:   - en/cloud-enterprise/current/ece-remote-cluster-ece-ess.html
  | 🚨 Error: The command exited with status 255

````

I am testing whether we can fix that problem by making the `ece-ref` attribute version-aware in https://github.com/elastic/docs/pull/2997

PR #2997, opened 2024-05-03
Add ECE version attributes

lcawl:
Relates to https://github.com/elastic/docs/pull/2996

The `ece-ref` shared attribute is currently set to the `current` URL, which is problematic if we have to revert the "current" version or link to topics that don't exist in `current` yet.

Since we need to move away from the book-scoped `ece-version` variables anyway, this PR adds ECE version files similarly to the Stack version files. Unlike those stack versions, however, the branch names in ECE do not match the version numbers so it is a little trickier to resolve the file names.

I've used a `ece-link-version` in the shared `ece-ref` variable rather than the actual `ece-version` since I don't want to potentially break things in the old releases. This way all but the latest releases can still link to "current".

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -2076,6 +2076,9 @@ contents:
                 path:   docs/shared
                 exclude_branches: [ 1.0 ]
               -
+                repo:   docs
+                path:   shared/versions/ece/{version}.asciidoc
+              -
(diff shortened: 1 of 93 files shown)


kilfoyle (approved):
Looks great! Thank you!!! 🦒

PR #3005, opened 2024-05-22
Ensure that _all_ legacy books contain `noindex: 1`

colleenmcginnis:
Makes sure that all versions of all legacy books are not being indexed. (Not sure if we should be indexing Functionbeat and SIEM docs as they still have live versions...)

cc @lcawl @KOTungseth @katymcdougall

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -2189,12 +2189,12 @@ contents:
     -   title:      Legacy Documentation
         sections:
           - title:      Elastic Stack and Google Cloud's Anthos
+            noindex:    1
             prefix:     en/elastic-stack-gke
             current:    main
             index:      docs/en/gke-on-prem/index.asciidoc
             branches:   [ {main: master} ]
             private:    1
-            noindex:    1
             chunk:      1
             tags:       Elastic Stack/Google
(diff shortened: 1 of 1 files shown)


bmorelli25 (approved):
Looks good. The interesting thing is that some of these (36) books don't show up in the legacy section on elastic.co/guide (28). `Kibana 사용자 가이드` is one example. Either way, 🐦

colleenmcginnis:
>Kibana 사용자 가이드 is one example

This one shows up as on a subsection (i made up this term) landing page: https://www.elastic.co/guide/kr/index.html

colleenmcginnis:
run docs-build

lcawl (dismissed):
LGTM 2. Seems like it's reasonable to remove the "live" list from the functionbeat and siem docs too

bmorelli25 (approved):

colleenmcginnis:
>Seems like it's reasonable to remove the "live" list from the functionbeat and siem docs too

👍  I needed to push another commit to trigger the tests to run anyway, so I removed the remaining `live` properties in legacy books.

Issue #1614, opened 2019-12-13
Update docs to include a H1 on all pages


zuketo:
Hi @AnneB-SEO I've created an issue in the docs repo for tracking work to add h1 tags. From your original issue:

"All pages would benefit from having a H1 heading. Currently sub topics start with a H2 and do not have a H1.

This change would improve the ability for crawlers to identify the page topic and improve ranking. While there is some debate within the SEO community around the importance of H1s, we performed our own evaluation and found a significant ranking increase after adding H1s to a number of different content types."

AnneB-SEO:
Thanks @zuketo !!!

PR #2768, opened 2023-09-21
Update styling to match elastic.co/docs

colleenmcginnis:
This PR updates the styles in our AsciiDoc-powered docs to match the Docsmobile styling used on docs.elastic.co. This will allow AsciiDoc- and Docsmobile-powered doc sets to exist together without impacting the reader experience.

### Development and design

- [x] Add initial layout and styles from docs.elastic.co
- [x] Refine responsive design
- [x] Closely review all layout elements:
  - [x] left nav
  - [x] main content
  - [x] on this page
  - [x] breadcrumbs
  - [x] version selector
- [x] Closely review all content UI elements:
    - [x] Added, coming, and deprecated notes (65d9d5c676f2d5bd4359d58d7d9457d51da2dd7a)
    - [x] Admonition blocks
    - [x] Beta, Dev, and Preview
    - [x] Code blocks
    - [x] Collapsible block
    - [x] Example blocks
    - [x] Headings
    - [x] Images
    - [x] Inline text
    - [x] Links
    - [x] Lists
    - [x] Paragraphs
    - [x] Sidebars
    - [x] Tabbed widgets
    - [x] Tables
    - [x] Thematic breaks
- [x] Clean up code

#### Interaction testing

@colleenmcginnis: 

- [x] Version selector
- [x] Tab widget testing
- [x] Collapsible block testing
- [x] Console code block
- [x] Code blocks with callouts
- [x] Code block copy button

#### Cross-browser / screen size testing

- [x] @colleenmcginnis Chrome
    - [x] Large screen size
    - [x] Laptop screen size
    - [x] Tablet size
    - [x] Mobile size
- [x] @colleenmcginnis Firefox
    - [x] Large screen size
    - [x] Laptop screen size
    - [x] Tablet size
    - [x] Mobile size
- [x] @colleenmcginnis Safari
    - [x] Large screen size
    - [x] Laptop screen size
    - [x] Tablet size
    - [x] Mobile size
- [ ] TBD Edge
    - [ ] Large screen size
    - [ ] Laptop screen size
    - [ ] Tablet size
    - [ ] Mobile size

#### Cross-book testing

Visually inspect a random sample of docs from each book your team owns. In a comment in this PR, note any edge cases that don't look quite right:

- [x] @elastic/obs-docs 
- [x] @elastic/platform-docs 
- [x] @elastic/security-docs 

### Update tests

- [x] Make sure all tests pass

### Related work

[...]

PR #3009, opened 2024-06-04
Always use h1 for first heading on a page

colleenmcginnis:
Related to https://github.com/elastic/docs/pull/2768 (specifically https://github.com/elastic/docs/pull/2768#issuecomment-2138009956)
Fixes https://github.com/elastic/docs/issues/1614

~~I _think_ this should force the first heading on every page to be an `h1`. However, this fix seems to easy so I'm not sure if I'm misunderstanding the total impact of this change. Opening this PR to see if the PR preview builds and what it looks like before fixing tests.~~ Switched to a different approach detailed in https://github.com/elastic/docs/pull/3009#issuecomment-2159328351 below.


colleenmcginnis:
It looks like this "fix" sometimes results in pages with multiple `h1` headings on (I _think_) pages that don't use `[discrete]` on headings that should be on the same page so I'm not sure it's much of a fix after all... I'm not sure if I should dig in further or rely on just styling the headings appropriately like we're doing in #2768.

bmorelli25:
Can you provide a double H1 example? I'm having trouble finding one.

colleenmcginnis:
Here are all of them in the 8.13 version of the Observability guide:

* [`html/en/observability/8.13/apm-redis-output.html`](https://github.com/elastic/built-docs/tree/docs_bk_3009/html/en/observability/8.13/apm-redis-output.html)
* [`html/en/observability/8.13/apm-privileges-rum-source-map.html`](https://github.com/elastic/built-docs/tree/docs_bk_3009/html/en/observability/8.13/apm-privileges-rum-source-map.html)
* [`html/en/observability/8.13/apm-privileges-agent-central-config.html`](https://github.com/elastic/built-docs/tree/docs_bk_3009/html/en/observability/8.13/apm-privileges-agent-central-config.html)
* [`html/en/observability/8.13/apm-logstash-output.html`](https://github.com/elastic/built-docs/tree/docs_bk_3009/html/en/observability/8.13/apm-logstash-output.html)
* [`html/en/observability/8.13/apm-kafka-output.html`](https://github.com/elastic/built-docs/tree/docs_bk_3009/html/en/observability/8.13/apm-kafka-output.html)
* [`html/en/observability/8.13/apm-elasticsearch-output.html`](https://github.com/elastic/built-docs/tree/docs_bk_3009/html/en/observability/8.13/apm-elasticsearch-output.html)
* [`html/en/observability/8.13/apm-console-output.html`](https://github.com/elastic/built-docs/tree/docs_bk_3009/html/en/observability/8.13/apm-console-output.html)
* [`html/en/observability/8.13/apm-configure-tail-based-sampling.html`](https://github.com/elastic/built-docs/tree/docs_bk_3009/html/en/observability/8.13/apm-configure-tail-based-sampling.html)
* [`html/en/observability/8.13/apm-configure-head-based-sampling.html`](https://github.com/elastic/built-docs/tree/docs_bk_3009/html/en/observability/8.13/apm-configure-head-based-sampling.html)
* [`html/en/observability/8.13/apm-configure-cloud-id.html`](https://github.com/elastic/built-docs/tree/docs_bk_3009/html/en/observability/8.13/apm-configure-cloud-id.html)

So there are not too many in Observability docs, but in bigger/older books there are many more. For example, in v8.13 of the Elasticsearch guide 843 out of 1815 HTML files contain more than one `h1`.

bmorelli25:
Discussed in metrics mtg this morning. @KOTungseth will ask Katy on Friday if no H1 is better or worse than multiple H1s.

colleenmcginnis:
Ok I tried a different approach here: trying to use the built-in AsciiDoc document header. It looks like we set the `noheader` option to `true`. I'm not sure why or if there will be unintended consequences by turning this off, but I haven't come across a dealbreaker... yet.

It's not as simple as changing that option. This PR would also:

* **Move the breadcrumbs**. If we use the document header with its h1, the current placement of the breadcrumbs would be _below_ the title. To address this, we can create a placeholder `div` for the breadcrumbs when we create the header, and then replace the placeholder when we go through the chunk logic. (Note: This is breaking the tests. We need to change where to expect the breadcrumbs.)
* **Hide the first heading in `div#content`**. When we use the document header, the title is still also added in the body of the doc (specifically in `div#content`), but we can hide that heading with CSS.
* **Still hack incremental heading styles**. The logic in this PR would ensure two things: (1) the first heading on each page is an h1 and (2) that will be the only h1 on the page. _However_, it doesn't guarantee that the headings in the main body of the doc will be incremental, but we can style them appropriately at run time.

bmorelli25:
Clicked around a bit.
- The [docs landing page](https://docs_bk_3009.docs-preview.app.elstc.co/guide/index.html) is a bit broken with this new update.
- It looks like OTP highlighting doesn't work, BUT, I love that the OTP exists on every page now and is even expanded on a lot of pages due to every page starting with H1.

colleenmcginnis:
>Clicked around a bit.
>
>* The [docs landing page](https://docs_bk_3009.docs-preview.app.elstc.co/guide/index.html) is a bit broken with this new update.
>* It looks like OTP highlighting doesn't work, BUT, I love that the OTP exists on every page now and is even expanded on a lot of pages due to every page starting with H1.

bmorelli25:
Awesome!

integtest/spec/single_book_spec.rb
```diff
@@ -314,133 +314,7 @@
...
-        xpack_tag_context true, true, true, true, false
-      end
-    end
-    context 'when hiding xpack tags' do
-      context 'when the xpack role is on everything' do
-        xpack_tag_context true, true, true, true, true
-      end
-    end
-  end
+  # Do we still need the xpack tag tests?
```
bmorelli25, line 317: I guess not...?

bmorelli25: I vote to remove. https://github.com/elastic/docs/pull/2213

bmorelli25 (approved):
This looks great. Nice fix for that 🐛 in 34a16c6. I've clicked around as much as I can, and I'm not seeing any new 🐛s.

🐦

PR #3025, opened 2024-06-26
Update catalog-info.yaml

glitteringkatie:
Give link checking script permission to read from wordlake-docs


scottybollinger (commented):
🐱

PR #3031, opened 2024-07-10
Update banner for master doc versions

lcawl:
This PR updates the banner that appears in "master" versions of the documentation, since it's currently unclear about the state of the serverless details in that branch.

## Before

image

## After

image

### DIFF
--- a/integtest/spec/all_books_spec.rb
+++ b/integtest/spec/all_books_spec.rb
@@ -639,9 +639,8 @@ RSpec.describe 'building all books' do
       it 'includes the prelim docs header' do
         expect(body).to include <<~HTML
           <div class="page_header">
-          This documentation contains work-in-progress information for future Elastic Stack and Cloud releases. Use the version selector to view supported release docs.
-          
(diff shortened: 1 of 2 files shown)


dedemorton:
I wonder if we can change this banner to be less confusing and provide users with a call to action. Maybe something like:

"You are looking at non-versioned documentation that may not be accurate if you're using a versioned Elastic Stack product. To see the documentation for the correct version, use the version selector."

leemthompo:
Thanks for this @lcawl, I think it makes a lot of sense to tackle some of this confusion in the banner.

I agree with @dedemorton that it would be nice to have a CTA but we also need to address the serverless disambiguation.

Here's a cut, with two blocks of text:


> This documentation contains work-in-progress information for future Elastic Stack and Cloud releases. Use the version selector to view supported release docs.

> This documentation also contains the latest Elastic Cloud serverless information. For our serverless-specific documentation, check out our [serverless docs](https://www.elastic.co/docs/current/serverless).

lcawl:
> Here's a cut, with two blocks of text:

I like that suggestion @leemthompo but I pruned it down a bit since the two blocks of text took up a lot of room at the top of each page.

bmorelli25 (approved):

PR #3015, opened 2024-06-18
Add edit link to h1s

colleenmcginnis:
Related to #3009 

Add edit link to h1s. 

_I will fix any tests that need to be fixed after the preview builds._

PR #3049, opened 2024-08-27
Move edit button out of heading elements

colleenmcginnis:
Closes https://github.com/elastic/docs/issues/1966

This PR moves the edit button out of the heading elements in the body of the content (edit buttons were already added outside page titles in #3015). This is important because when the edit button is inside the heading element, technically the full text of the heading includes `edit` appended to the end of the intended heading text. This is getting picked up by Google, but may also cause unexpected results elsewhere including for users using screenreaders.

Screenshot 2024-08-27 at 9 44 20 AM

Note: I opted for updating the HTML and using Bootstrap CSS classes to avoid having to bump the name (`-v1` to `-v2`) of the CSS file. If the user gets the updated HTML with a cached version of the CSS and JavaScript files, the edit button should still be positioned correctly and work as expected.

### DIFF
--- a/resources/asciidoctor/lib/docbook_compat/convert_floating_title.rb
+++ b/resources/asciidoctor/lib/docbook_compat/convert_floating_title.rb
@@ -14,7 +14,7 @@ module DocbookCompat
         xpack_tag(node),
         '</', tag_name, '>',
         node.attr('edit_me_link', ''),
-        '</div>',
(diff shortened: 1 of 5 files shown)


bmorelli25 (approved):
Wow. You can even copy/paste headings now without the "edit" getting in the way!

PR #3061, opened 2024-09-11
Update local installation URL in code widget

leemthompo:
Flagged by @serenachou, we didn't have this [nice local dev page](https://www.elastic.co/guide/en/elasticsearch/reference/current/run-elasticsearch-locally.html) when we updated the widget.

## Notes for reviewers

This only spins up ES and Kibana, so reviewer let me know if you agree with pointing there.

ℹ️ We'll update again to point to `start-local` when the time comes


bmorelli25 (approved):

KOTungseth (approved):
LGTM

PR #3066, opened 2024-09-11
feat(apm-*): only build docs if there's a doc change

kruskall:
extend the check to all apm repositories and use a switch to make it easier to add more repo that share common file patterns

as usual nothing changes for repositories not mentioned in the switch (they always build and no git diff is run)


bmorelli25:
Fantastic! I added comments to your code (mostly for myself) but also since this script is pretty well documented and I wanted to keep it that way. Based on your prior testing, I'm pretty confident this will work, but I'm going to wait until tomorrow morning (my tomorrow morning) to merge. That way if things go wrong I can revert ASAP.

Thank you!

bmorelli25 (approved):

PR #3085, opened 2024-10-10
Draft some new admonitions

lcawl:
This PR adds some admonitions for serverless- and ESS-specific admonitions:

- beta_ess
- beta_serverless
- dev_ess
- dev_serverless
- ga_serverless
- ga_stack
- ga_ess
- deprecated_ess
- deprecated_serverless
- discontinued_ess
- discontinued_serverless
- discontinued_stack
- coming_ess
- coming_serverless

The work similar to the existing admonitions, for example:

image

NOTE: The rubocop config file change works around the following make error:

> lib/care_admonition/extension.rb:100:3: C: Metrics/MethodLength: Method has too many lines. [31/20]

TBD: Are there other combinations that we'll need (e.g. "ga_all" to have a single admonition that covers all contexts)?

### DIFF
--- a/resources/asciidoctor/lib/care_admonition/extension.rb
+++ b/resources/asciidoctor/lib/care_admonition/extension.rb
@@ -3,16 +3,43 @@
 require 'asciidoctor/extensions'
 
 ##
-# Extensions for marking something as `beta`, `dev`, or technical `preview`.
+# Extensions for marking something as `beta`, `beta_serverless`, `beta_ess`,
+# `dev`, `dev_serverless`, `dev_ess`, technical `preview`, `preview_serverless`,
+# `preview_ess`, `ga_serverless`, `ga_ess`.
 #
 # Usage
 #
 #   beta::[]
 #   dev::[]
 #   preview::[]
+#   beta_ess::[]
+#   beta_serverless::[]
+#   dev_ess::[]
+#   dev_serverless::[]
+#   ga_serverless::[]
+#   deprecated_ess::[]
+#   deprecated_serverless::[]
+#   discontinued_ess::[]
+#   discontinued_serverless::[]
+#   coming_ess::[]
+#   coming_serverless::[]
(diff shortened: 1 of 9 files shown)


leemthompo:
- I'd like a simple "Not available on serverless" tag too :)
- I guess we should `s/Elastic Serverless/Elastic Cloud Serverless/g`

leemthompo (approved):

resources/asciidoctor/lib/care_admonition/extension.rb
```diff
@@ -21,19 +48,83 @@ class CareAdmonition < Asciidoctor::Extensions::Group
...
+  TEXT
+  DEPRECATED_SERVERLESS_TEXT = <<~TEXT.strip
+    This functionality is deprecated in Elastic Cloud Serverless and will be removed in a future release.
+  TEXT
+  DISCONTINUED_ESS_TEXT = <<~TEXT.strip
+    This functionality is discontinued in Elasticsearch Service.
+  TEXT
+  DISCONTINUED_SERVERLESS_TEXT = <<~TEXT.strip
+    This functionality is discontinued in Elastic Cloud Serverless.
+  TEXT
```
colleenmcginnis, line 86: What is the difference between deprecated and discontinued? Can we make the discontinued description more specific?

lcawl: We haven't had "discontinued" before but it arose in discussions like https://github.com/elastic/kibana/issues/192292. Currently we just remove content when the pertinent feature goes from "deprecated" to discontinued/removed/no longer available. If our pages span multiple versions, however, or if we want to give folks a clear indication of when a feature will be removed, it seems like this might be something we'll want to have a way of calling out.

lcawl:
If we move forward with this, it'll be with https://github.com/elastic/docs/pull/3121 so this PR can be closed.

PR #3102, opened 2024-11-04
Update debian base docker image to latest (bookworm).

Mpdreamz:
Updated Docker base image from Debian Buster to Bookworm. Adjusted environment variables, package versions, and configuration for compatibility with Ruby 3.1 and newer libraries.

(include `.idea` in gitignore to ensure pycharm IDE files do not get checked in).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.idea
 .gradle
 bin
 .includepath
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@
 
 # Debian builds the docs about 20% faster than alpine. The image is larger
 # and takes longer to build but that is worth it.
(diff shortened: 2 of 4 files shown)


Mpdreamz:
Disabled `rubocop` for now, updated it so it will run successfully but not all style violations can be auto corrected with:

> ./build_docs --docker-run ruby_test rubocop -A

Returns:

> 105 files inspected, 113 offenses detected, 81 offenses corrected

Will follow up with a PR addressing this if I manage to get this green.

Mpdreamz:
Six RSpec tests are failing, all related to avoiding rebuilds when switching branches. 

Looks like something changed with how the Perl scripts handle tracking. 

Not sure yet what's causing it, so no fix in sight for now.

PR #3107, opened 2024-11-06
Change admonition title style/syntax

colleenmcginnis:
The new AsciiDoc serverless docs have lots of admonitions with titles. Admonition titles are currently `h3`s. I'd like to change them to:

* _Not_ use a heading tag. Using an `h3` messes with the "On this page" list of headings and can result in headings that are not necessarily incremental (admonitions can come after `h3`s, `h4`s, etc).
* Style using bold text instead of larger text. This is subjective. I just think the current text is too big. 

| Before | After |
|---|---|
| Screenshot 2024-11-06 at 5 03 33 PM | Screenshot 2024-11-06 at 5 03 54 PM |


lcawl (approved):
I agree it looks better

PR #3108, opened 2024-11-07
Centralize icons 

colleenmcginnis:
When we migrated serverless docs to AsciiDoc, we had to add SVGs for icons that were previously handled by the `DocIcon` component. To allow us to move more quickly, I added the SVGs to each content repo, but planned to centralize those icons in this repo later (i.e. now 😄).

The approach I used in this PR would mean the new way to include icons in AsciiDoc files would be to use a shared attribute following the pattern `icon-{type}` where `type` in the string that was used in `DocIcon` (and is listed in [EUI icon docs](https://eui.elastic.co/#/display/icons)).

For example:

```asciidoc
** **Add note to timeline** ({icon-editorComment}): Add the selected text to your currently active Timeline as a note.
** **Add to existing case** ({icon-appAddData}): Add a comment to an existing case using the selected text.
** **Copy to clipboard** ({icon-copyClipboard}): Copy the text to clipboard to paste elsewhere. Also helpful for resubmitting a previous prompt.
** **Add to timeline** ({icon-timeline}): Add a filter or query to Timeline using the text. This button appears for particular queries in AI Assistant's responses.
```

For the reviewer &mdash;  here's specifically what this PR does:

* Adds EUI SVGs for icons and logos.
* Creates a class for each icon following the pattern `icon-{type}`.
* Styles and sizes each class to fit inline in regular text.
* Creates a shared attribute for each icon following the pattern `icon-{type}`.
* Bumps the CSS file version to ensure the new classes are accessible (bust the cache).

After this is merged, I can open a PR in /docs-content, /security-docs, and /observability-docs to remove the local SVGs and update the syntax to use the new shared attributes.


colleenmcginnis:
We're going to delay this until v3. cc @bmorelli25

PR #3109, opened 2024-11-07
Add serverless shared attributes

lcawl:
Relates to https://github.com/elastic/docs-content/pull/182

This PR moves the declarations from https://github.com/elastic/docs-content/blob/7602ddbdb3de8fe99b0850423c308f85e2952d6d/serverless/index.asciidoc?plain=1#L13 to the list of shared attributes so that they can be used in all books.

### DIFF
--- a/shared/attributes.asciidoc
+++ b/shared/attributes.asciidoc
@@ -181,11 +181,13 @@ Elastic Cloud
 :ess:         Elasticsearch Service
 :ece:         Elastic Cloud Enterprise
 :eck:         Elastic Cloud on Kubernetes
-:serverless-full:  Elastic Cloud serverless
-:es-serverless: Elasticsearch on serverless
-:serverless-short: serverless
-:serverless-docs: https://docs.elastic.co/serverless
-:es3:         Elasticsearch Serverless
+:serverless-full:  Elastic Cloud Serverless
+:serverless-short: Serverless
+:es-serverless:    Elasticsearch Serverless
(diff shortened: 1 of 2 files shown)


colleenmcginnis (dismissed):
Thanks! This was in my post-migration list along with #3107 and #3108. 🙂  Feel free to delete from the file in /docs-content, too!

shared/attributes.asciidoc
```diff
@@ -181,10 +181,13 @@ Elastic Cloud
...
 :ece:         Elastic Cloud Enterprise
 :eck:         Elastic Cloud on Kubernetes
-:serverless-full:  Elastic Cloud serverless
-:es-serverless: Elasticsearch on serverless
-:serverless-short: serverless
-:serverless-docs: https://docs.elastic.co/serverless
+:serverless-full:  Elastic Cloud Serverless
+:serverless-short: Serverless
+:es-serverless:    Elasticsearch Serverless
+:es3:              Elasticsearch Serverless
```
colleenmcginnis, line 187: @lcawl does it make sense to have both of these?

colleenmcginnis (dismissed):

colleenmcginnis (approved):

shared/attributes.asciidoc
lcawl: I don't think it hurts since "es3" seems to be more common at the moment but "es-serverless" is more in alignment with the other projects (and more obvious IMO).

PR #3126, opened 2024-11-22
Accommodate sticky marketing nav

colleenmcginnis:
It looks like the marketing nav is now sticky when you scroll down the page. As a result it's covering the book title and version dropdown on the left and part of the "On this page" section on the right. 

| Before marketing nav change | Now (with sticky marketing nav) |
|---|---|
| Screenshot 2024-11-21 at 5 58 23 PM | Screenshot 2024-11-21 at 5 57 45 PM |

This PR adjusts the CSS to accommodate the new nav behavior. This fix is pretty fragile &mdash; if the height of the nav at any screen width changes, the docs content may have too much or too little space at the top. I'm not sure if there's a better solution.

### DIFF
--- a/resources/web/style/base_styles.pcss
+++ b/resources/web/style/base_styles.pcss
@@ -40,10 +40,15 @@
   .sticky-top-md {
     position: -webkit-relative;
     position: relative;
-    @media screen and (min-width: 769px) {
+    @media screen and (min-width: 1200px) {
       position: -webkit-sticky;
       position: sticky;
-      top: 0;
+      top: 72px;
+    }
+    @media screen and (min-width: 769px) and (max-width: 1199px) {
+      position: -webkit-sticky;
+      position: sticky;
+      top: 131px;
     }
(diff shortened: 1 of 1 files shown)


dedemorton:
I'm going to approve this because I think it's better than how the site currently behaves (still not great).
However it would be better to find a long-term solution that is not fragile. It would also be better if we had a chance to test changes like this before they get dumped on our users.

IMO the heavyweight nature of the site menu (with large popups that users can easily trigger by mistake) makes it really intrusive for users who are simply trying to read the docs.

Every time I accidentally trigger those popups (which is pretty much every time I try to click a breadcrumb, I feel instantly pissed off and wonder if users are experiencing the same feelings. 

People who read the docs go there for a very different reason than casual site visitors who are trying to learn about Elastic. Users of the documentation should not have to be distracted by popups that they don't care about.

dedemorton (approved):

PR #3132, opened 2024-11-28
Update shared cloud signup instructions to account for serverless GA being the new default

florent-leborgne:
[Merge only when the signup will switch to serverless by default]

@KOTungseth As I'm heading out for a few weeks, I'll give you ownership of this PR, as you may have a better sense of when the eng teams plan on making the switch (signups defaulting to serverless instead of deployments as it is right now)

This PR updates the shared cloud signup instructions. It's kind of short now but at least it's correct, I don't think it's worth describing each step as the UX is self-sufficient.
The ess trial URL variable also looked odd so I took the liberty of updating it with a more direct URL.

Rel: https://github.com/elastic/platform-docs-team/issues/485


lcawl (approved):
LGTM

PR #3136, opened 2024-12-04
Add Copy to Clipboard CTA tracking

bmorelli25:
### Summary

Adds the ability to track when users click on "Copy to Clipboard."

cc @KOTungseth @mcewenleah


christianweinke (approved):
@bmorelli25 Looks great, that works!

PR #3143, opened 2024-12-10
Updates serverless docs attribute

KOTungseth:
Updates :serverless-docs: attribute


shainaraskas (approved):
thanks - was just running into a problem with this!

PR #3148, opened 2024-12-17
Fix collapsible animation

colleenmcginnis:
In collapsible blocks, only animate the icon before the `summary` that is a direct child of the `details` element.


shainaraskas (approved):
tested using [this](https://docs_bk_3148.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/current/put-dfanalytics.html#ml-put-dfanalytics-request-body) - lgtm!

PR #3150, opened 2024-12-19
Add ecs-logging URL attributes

colleenmcginnis:
Related to https://github.com/elastic/observability-docs/issues/1393

Adds ECS logging GitHub URL attributes.

### DIFF
--- a/shared/attributes.asciidoc
+++ b/shared/attributes.asciidoc
@@ -555,6 +555,28 @@ Issue and pull request URLs
 :apm-issue:    {apm-repo}issues/
 :apm-pull:     {apm-repo}pull/
 :kibana-blob:  {kib-repo}blob/{branch}/
+
+:ecs-logging-ruby-repo:       https://github.com/elastic/ecs-logging-ruby
+:ecs-logging-ruby-blob:       {ecs-logging-ruby-repo}/blob/main
+:ecs-logging-python-repo:     https://github.com/elastic/ecs-logging-python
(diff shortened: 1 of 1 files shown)


shared/attributes.asciidoc
```diff
@@ -555,6 +555,28 @@ Issue and pull request URLs
 :apm-issue:    {apm-repo}issues/
 :apm-pull:     {apm-repo}pull/
 :kibana-blob:  {kib-repo}blob/{branch}/
+
+:ecs-logging-ruby-repo:       https://github.com/elastic/ecs-logging-ruby
+:ecs-logging-ruby-blob:       {ecs-logging-ruby-repo}/blob/main
```
lcawl, line 560: > :ecs-logging-ruby-blob:       {ecs-logging-ruby-repo}/blob/main

Drive by comment that if you want "main" to be branch-specific you could maybe use {branch} or put these in the version-specific files.

colleenmcginnis: >maybe use {branch}

That's what I did initially but it didn't work locally.

>put these in the version-specific files

We already have attributes for ECS logging in the [version attribute files](https://github.com/elastic/docs/blob/master/shared/versions/stack/8.17.asciidoc#L53-L65), but these use `master` because they are used in the `elastic.co/guide` URL attributes, but the branch name on GitHub is `main`. 🤔

colleenmcginnis:
Stale.

PR #3155, opened 2025-01-09
Add api-ls shared attribute

lcawl:
This PR adds attributes for the Logstash API doc URL: https://www.elastic.co/docs/api/doc/logstash

At this time there's only a single branch/version of the new API docs, but I've put the attribute in the same location as the ES and Kibana URL attributes just in case that changes.


karenzone (approved):
LGTM 🎸

PR #3157, opened 2025-01-21
Add ECS 8.17 to documentation.

mjwolf:
The next ECS version, 8.17.0, is being prepared. This adds ECS 8.17 to documentation. The "current" ECS version in conf.yaml is intentionally not updated yet, as ECS 8.17 isn't released yet, so it's not the current version yet.


lcawl (approved):
LGTM

PR #3161, opened 2025-01-30
Adding 8.18 and 8.18.0 details, master not bumped as it's on 9.0.0, current still 8.17

eedugon:
This needs review as I have some concerns due to:
- We have already master pointing to 9.0.0.
- There's a comment on ECS about keeping it in 8.17.

Comments added in the code.


conf.yaml
```diff
@@ -1031,7 +1031,7 @@ contents:
             prefix:     en/ecs
             # Please do not update current to 8.18
             current:    8.17
```
eedugon, line 1033: @lcawl , @bmorelli25 : I have a question here... are we going to keep ECS current in 8.18 even after releasing 8.18? why do we have that comment to not update current to 8.18? And why we don't have current = stackcurrent?

shared/versions/stack/8.18.asciidoc
```diff
@@ -0,0 +1,77 @@
...
+:kibana_version:         8.18.0
+:apm_server_version:     8.18.0
+// branch is 8.18 instead of 8.x for long-term linking purposes
+:branch:                 8.18
+:minor-version:          8.18
+:major-version:          8.x
+:prev-major-version:     7.x
+:prev-major-last:        7.17
+:major-version-only:     8
+:ecs_version:            8.18
```
eedugon, line 17: @lcawl , @bmorelli25 : would that line on ecs_version be all right? (this is related with my previous question on conf.yaml about ECS)

shared/versions/stack/master.asciidoc
```diff
@@ -11,9 +11,9 @@ bare_version never includes -alpha or -beta
 :minor-version:          9.0
 :major-version:          9.x
 :prev-major-version:     8.x
-:prev-major-last:        8.17
+:prev-major-last:        8.18
 :major-version-only:     9
-:ecs_version:            8.17
+:ecs_version:            8.18
```
eedugon, line 16: Same here @lcawl , @bmorelli25 .

Also I have left master as 9.0 and only changed the references to 8.17 to 8.18. Let me know if this looks all right.

conf.yaml
lcawl: The ECS folks manage that book and just bumped it recently in https://github.com/elastic/docs/pull/3157 it doesn't follow the same cadence as the other stack-versioned docs.

shared/versions/stack/8.18.asciidoc
lcawl: It should be left unchanged per earlier reply about ECS doing its own thing.

shared/versions/stack/master.asciidoc
lcawl: IMO you shouldn't need to touch master.asciidoc at all.

eedugon: But `master` has this...

```
prev-major-last:        8.17
```

Shouldn't that line be bumped to 8.18?

eedugon: @lcawl , I'd need a response here. If you prefer I won't touch `master.asciidoc` at all, but the `prev-major-last` will be kept on 8.17.

eedugon:
@lcawl , @bmorelli25 , I'd need an approval on this PR soon.
Based on @karenzone comment it looks this is blocking other docs repositories to create things for 8.18, like https://github.com/elastic/logstash-docs/actions/runs/13249296075.

The only remaining issue on this PR is if the `prev-major-last:` value on `master.asciidoc` needs to be bumped to `8.18` or if we leave it in `8.17`. I've updated it to 8.18.

shared/versions/stack/8.18.asciidoc
eedugon: `8.18.asciidoc` is a new file. What do you mean with `unchanged` in this context?

eedugon: Ok, I see what you mean, sorry :) the example provided in the template left ecs in a previous version in the new stack version file, so I'm doing the same. Fixed it... thanks!

shared/versions/stack/master.asciidoc
```diff
@@ -11,7 +11,7 @@ bare_version never includes -alpha or -beta
 :minor-version:          9.0
 :major-version:          9.x
 :prev-major-version:     8.x
-:prev-major-last:        8.17
+:prev-major-last:        8.18
```
eedugon, line 14: @lcawl : this is the only change planned for `master.asciidoc`. If I have to leave it untouched and keep `prev-major-last` on 8.17 let me know and I'll do it. In such case... when should we update that line? because the previous major last will be 8.18

bmorelli25:
run docs-build rebuild

bmorelli25 (approved):
LGTM

leemthompo (approved):

PR #3182, opened 2025-03-04
Test using glob in `build_pr.sh`

colleenmcginnis:
Test out using a glob pattern in `build_pr.sh` to determine when we should run the AsciiDoc preview build. If this works, it will be an imperfect solution because it would not build if only an image was updated in a PR, but that might be ok. 🤷 

To test this we'll have to:

1. Merge this PR
2. Open a test PR in elastic/ecs-logging

Then come back here and update the other repos.

@bmorelli25 what do you think about this approach?


reakaleek (approved):

colleenmcginnis:
I think it worked:

- https://github.com/elastic/ecs-logging/pull/87
- https://github.com/elastic/ecs-logging/pull/88

PR #3201, opened 2025-04-07
updating cloud for release notes in asciidoc

georgewallace:
Updating the conf.yaml for the cloud release notes that will be staying in asciidoc.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -2087,6 +2087,7 @@ contents:
               -
                 repo:   cloud
                 path:   docs/shared
+                exclude_branches: [ master ]
           - title:      Elasticsearch Add-On for Heroku - Hosted Elasticsearch and Kibana for Heroku Users
             prefix:     en/cloud-heroku
             tags:       Cloud-Heroku/Reference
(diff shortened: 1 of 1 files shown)


georgewallace:
@KOTungseth PR for the docs repo for the release notes. I think I got this right.

lcawl (dismissed):
If the builds pass, LGTM

lcawl:
Seeing all these broken links makes me question whether rather than removing all the pages from https://www.elastic.co/guide/en/cloud/current/index.html (except release notes) is the right path forward or whether it's simpler to just create a "new" book that contains only the release notes. Since redirects are being created for all those pages, IMO that saves us from having to do all this clean up of broken links.

lcawl:
I think this can be closed since it's no longer required.

florent-leborgne:
Just seeing this now as I was out when it was done. I think creating a new guide just breaks all URLs so it's not the ideal solution. If we want to avoid breaking all links could this work instead:

- Keep the currentSaas to the last cloud milestone branch containing asciidoc (ms-122 I believe)
- Hide (instead of delete) all non-release notes pages

@lcawl @georgewallace

PR #3202, opened 2025-04-07
Add Cloud Hosted Release Notes book

lcawl:
This PR is intended as an alternative to https://github.com/elastic/docs/pull/3201, which aims to continue publishing the old Cloud Hosted release notes in the old docs system.

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -2271,7 +2271,19 @@ contents:
               -
                 repo:   terraform-provider-ec
                 path:   docs-elastic/
-
+          - title:      Elastic Cloud Hosted Release Notes
+            prefix:     en/cloud-hosted
+            tags:       Cloud/Reference
+            subject:    Elastic Cloud
+            current:    master
+            branches:   [ master ]
+            index:      docs/saas/index.asciidoc
+            single:     1
+            chunk:      1
+            sources:
+              -
+                repo:   cloud
+                path:   docs/saas
     -   title:      Legacy Documentation
         sections:
           - title:      Elastic Stack and Google Cloud's Anthos


shainaraskas:
I'm not sure if this is on the cloud side or this side, but I think something is wrong here because we're getting all of the pages rendered on one big super-page, no sidebar, and links are not working

e.g. https://docs_bk_3202.docs-preview.app.elstc.co/guide/en/cloud-hosted/current/index.html

image

lcawl:
> I'm not sure if this is on the cloud side or this side, but I think something is wrong here because we're getting all of the pages rendered on one big super-page, no sidebar, and links are not working

Ah, it worked locally so I think it's because I used the "single: 1" which makes sense in the TPEC case but not here.

shainaraskas:
couple small things: 

* variables are not rendering:

		image

* title should be changed: 

       image
       
* "Elasticsearch Service" should be replaced with "Elastic Cloud Hosted" everywhere (or maybe "Elastic Cloud" given that we also have console updates in these release notes?) 

otherwise, the structure looks good to me. we do have a lot of duplicate headings in the sidebar, but those were present in the old system.

I am a little concerned about the links though - are we cool to rely on redirects to get people to the right place? e.g.:
image
https://docs_bk_3202.docs-preview.app.elstc.co/guide/en/cloud-hosted/current/ec-release-notes-2019-10-23.html

lcawl:
> couple small things:
> 
> * variables are not rendering:

I've got this working by adding references to the version and shared attribute files in https://github.com/elastic/cloud/pull/139928

> * title should be changed:

I changed it to "Elastic Cloud Release Notes" in https://github.com/elastic/cloud/pull/139928 but if there's a better name that's fine by me.

> * "Elasticsearch Service" should be replaced with "Elastic Cloud Hosted" everywhere (or maybe "Elastic Cloud" given that we also have console updates in these release notes?)

I updated the `n` attribute in https://github.com/elastic/cloud/pull/139928

> I am a little concerned about the links though - are we cool to rely on redirects to get people to the right place?

I think since these are old release notes it's okay to have them point to the old docs. They're now working (i.e. pointing to "current") in https://github.com/elastic/cloud/pull/139928

shainaraskas (approved):
as long as this looks right after https://github.com/elastic/cloud/pull/139928, that all works for me !

lcawl:
run docs-build

PR #3206, opened 2025-04-16
Update `8.x` to `8.19` in config

colleenmcginnis:
The build is failing because repos that contain docs _and_ code are changing their `8.x` branch to `8.19`. I don't expect this to pass on the first try, but I'll keep an eye on it as more repos make this change.


colleenmcginnis:
Not all repos are creating `8.19` branches yet so I'm not sure how to handle it when a repo with `8.x` and a repo with `8.19` are used in the same book.

lcawl (approved):
Pre-approval once builds are green

PR #3207, opened 2025-04-17
Update Kibana docs to use 8.19 branch for 8.x

lcawl:
Relates to https://github.com/elastic/docs/pull/3206

This PR builds the 8.x docs from the new 8.19 branch until all the 8.19 branches exist.
I've also added a hack in the `build_docs.pl` file since I think otherwise we'd fail to resolve the branch-specific URLs in the Kibana doc link service.
We might also need to update the version assignment in https://github.com/elastic/kibana/blob/main/src/platform/packages/shared/kbn-doc-links/src/get_doc_meta.ts or else accept broken links in Kibana until 8.19 docs exist.


build_docs.pl
```diff
@@ -409,7 +409,18 @@ sub check_kibana_links {
         #
         # TODO: remove as part of
         # https://github.com/elastic/docs/issues/2264
-        $branch = $version eq "master" ? "main" : $version;
+        if ($version eq "master") {
+            $branch = "main"
+        }
+        else {
+            if ($version eq "8.19") {
+                $branch = "8.x"
```
colleenmcginnis, line 417: Should this be the reverse? The branch is `8.19` but the version we show in the dropdown etc is `8.x`?

colleenmcginnis:
```suggestion
            if ($version eq "8.x") {
                $branch = "8.19"
```

PR #3211, opened 2025-04-23
Update `v3-mapping.json` with the latest v3 mapping 

colleenmcginnis:
Updates `v3-mapping.json` with the latest v3 mapping.

It also deletes all mappings that point back to the current version of the AsciiDoc docs because it doesn't make sense to link to those URLs because they just lead to the same page the viewer is already on. The link on those pages should point to elastic.co/docs instead.


bmorelli25 (approved):
❇️

PR #3213, opened 2025-04-25
Remove `(current)` from the version dropdown

colleenmcginnis:
Removes `(current)` from the version dropdown to avoid confusion about which version is the latest now that the latest docs have been migrated to docs-builder and elastic.co/docs.

Screenshot 2025-04-25 at 8 45 04 AM

Note: This will remove `(current)` from version dropdowns in _all_ books &mdash; even books that were _not_ migrated to elastic.co/docs. Let me know if there are any concerns about this.

### DIFF
--- a/.github/workflows/patch-release-version-bump.yml
+++ b/.github/workflows/patch-release-version-bump.yml
@@ -4,11 +4,11 @@ on:
    inputs:
       MAJOR_VERSION:
         type: choice
-        description: 'Major version (current)'
+        description: 'Major version'
         options: ['7', '8']
       MINOR_VERSION:
         type: choice
-        description: 'Minor version (current)'
+        description: 'Minor version'
         options: ['12', '13', '14', '15', '16', '17', '18', '19', '20', '21', '22', '23', '24', '25']
       PATCH_VERSION:
         type: choice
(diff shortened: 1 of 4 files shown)


lcawl:
I wonder if we also want to add "noindex" to the current branches so they begin to retire from search sites? Currently it looks like we only have that option in conf.yaml at the book level, however.

colleenmcginnis:
>I wonder if we also want to add "noindex" to the current branches so they begin to retire from search sites? Currently it looks like we only have that option in conf.yaml at the book level, however.

I think we should consult with the SEO team before doing this. @KOTungseth has this come up in any of your conversations?

kilfoyle (approved):
LGTM! ⛵

KOTungseth:
> > I wonder if we also want to add "noindex" to the current branches so they begin to retire from search sites? Currently it looks like we only have that option in conf.yaml at the book level, however.
> 
> I think we should consult with the SEO team before doing this. @KOTungseth has this come up in any of your conversations?

@colleenmcginnis it's going to take a few weeks for Google to reindex. SEO is going to share insights when we have real data. Since 8.18 is not current, we should remove confusion by removing `current` from the /guide version dropdown.

PR #3216, opened 2025-04-30
Update clients docs to use 8.19 branch

leemthompo:
Attempts to fix broken build

@pquentin pointed to this list of repos that renamed 8.x to 8.19: https://github.com/elastic/devtools-team/blob/main/.github/workflows/scripts/clients.config

### DIFF
--- a/conf.yaml
+++ b/conf.yaml
@@ -1073,6 +1073,8 @@ contents:
                   -
                     repo:   elasticsearch
                     path:   docs/java-rest/
+                    map_branches:
+                      8.x:  8.19
                   -
                     repo:   elasticsearch
                     path:   client
@@ -1214,8 +1216,6 @@ contents:
                   -
                     repo:   elasticsearch
                     path:   docs/java-rest
-                    map_branches:
-                      8.x:  8.19
                   -
                     repo:   elasticsearch
                     path:   docs/Versions.asciidoc
--- a/conf.yaml
+++ b/conf.yaml
@@ -1079,7 +1079,7 @@ contents:
               - title:      JavaScript Client
                 prefix:     javascript-api
                 current:    *stackcurrent
-                branches:   [ 8.x, 8.18, 8.17, 8.16, 8.15, 8.14, 8.13, 8.12, 8.11, 8.10, 8.9, 8.8, 8.7, 8.6, 8.5, 8.4, 8.3, 8.2, 8.1, 8.0, 7.17, 7.16, 6.x, 5.x ]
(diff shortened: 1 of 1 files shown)


szabosteve (approved):
LGTM

conf.yaml
```diff
@@ -386,7 +376,7 @@ contents:
               - title:      Enterprise Search Node.js client
                 prefix:     enterprise-search-node
                 current:    *stackcurrent
-                branches:   [ 8.x, 8.18, 8.17, 8.16, 8.15, 8.14, 8.13, 8.12, 8.11, 8.10, 8.9, 8.8, 8.7, 8.6 ]
+                branches:   [ 8.19, 8.18, 8.17, 8.16, 8.15, 8.14, 8.13, 8.12, 8.11, 8.10, 8.9, 8.8, 8.7, 8.6 ]
```
pquentin, line 379: So while I did rename the branch here for consistency, I don't think we expect to release an 8.19 release for any enterprise search clients. Should we exclude 8.19 for all Enterprise Search clients, or is it useful to keep in 8.19 the mention that it was deprecated in 8.18 and won't see any other release?

leemthompo: Good question is there no 8.19 release for enterprise search?

leemthompo:
@pquentin I think we need to update at least some of the java client entries, but tbh I'm not super sure about all of them:

https://github.com/elastic/docs/blob/5cb7ead41cbb6868e7f2760c98f4ba94323bc705/conf.yaml#L1064-L1082

pquentin:
@leemthompo Not sure either, I don't think we care anymore about [elasticsearch/client](https://github.com/elastic/elasticsearch/tree/main/client) or [elasticsearch/docs/java-rest](https://github.com/elastic/elasticsearch/tree/main/docs/java-rest/). @swallez @l-trotta any idea?

leemthompo:
The build error FYI:

```
INFO:build_docs:Error executing: git archive --format=tar -o /tmp/docsbuild/Z4QZ2pb99E/elasticsearch-java/.temp_git_archive.tar 8.x docs in GIT_DIR /docs_build/.repos/elasticsearch-java.git
INFO:build_docs:fatal: Not a valid object name
```

conf.yaml
colleenmcginnis, line 1: It looks like both the Python and Rust clients need to be updated in this file. The Perl client is still using `8.x` and the rest seem to have never made it to v8.

colleenmcginnis (approved):
The build is lookin' good so far!

PR #3221, opened 2025-05-07
Add shared attribute for classic docs

karenzone:
Logstash plugin docs need this attribute in order to resolve correctly for 8.x and earlier versions.


colleenmcginnis (approved):
✨

PR #3233, opened 2025-06-19
feature/build only branches with legacy asciidoc

reakaleek:
## Changes

Only build 8.x target branches.

## Context

According to https://github.com/elastic/buildkite-pr-bot?tab=readme-ov-file#configuration-1 you can actually set the `target_branch` in `pull-requests.org-wide.json`.

However, the `buildkite/docs-build-pr` status check is usually a required check.

In this PR we exit early if the target branch doesn't match the pattern, so that the status check will still be green.

### DIFF
--- a/.buildkite/pull-requests.org-wide.json
+++ b/.buildkite/pull-requests.org-wide.json
@@ -12,7 +12,6 @@
       "allowed_list": ["github-actions[bot]", "renovate[bot]", "mergify[bot]", "dependabot[bot]", "elastic-renovate-prod[bot]", "elastic-observability-automation[bot]"],
       "build_on_commit": true,
       "build_on_comment": true,
-      "target_branch": ["8.x", "8.19", "8.18", "8.17", "8.16", "8.15", "8.14", "8.13", "8.12", "8.11", "8.10", "8.9", "8.8", "8.7", "8.6", "8.5", "8.4", "8.3", "8.2", "8.1", "8.0"],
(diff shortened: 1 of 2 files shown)


shainaraskas:
We have other docs sets that are not on stack versioning that wouldn't be caught in this fix. This would help a lot with certain repos but I wonder if it needs to be refactored to cover stuff like ECE 4+ and ECK 3+

bmorelli25:
I'm trying to think through Shaina's comment. We could potentially pair Jan's change here with the repo-specific instructions that start on L68. We could write a script to parse conf.yaml and spit out a mapping of each repo and the versions of documentation that are contained within that repo. We could then check those versions.

For example:

```diff
    # repositories with a docs dir and changelog
    "apm-aws-lambda" | "apm-agent-android" | "apm-agent-nodejs" | "apm-agent-python" | "apm-agent-ruby" | "apm-agent-rum-js" | "apm-agent-go" | "apm-agent-java" | "apm-agent-dotnet" | "apm-agent-php" | "apm-agent-ios")
+      if ! [[ $GITHUB_PR_TARGET_BRANCH =~ {NEWREGEXHERE} ]]; then
+       echo "Target branch '$GITHUB_PR_TARGET_BRANCH' is not a valid version branch (must be in format
+     X.Y where X <= 8)"
+      exit 0
+        fi
      git fetch origin "$GITHUB_PR_TARGET_BRANCH"
      docs_diff=$(git diff --stat "origin/$GITHUB_PR_TARGET_BRANCH"...HEAD -- "./docs/**.asciidoc" CHANGELOG.asciidoc)
      ;;
```

(this is just to illustrate)

PR #3235, opened 2025-06-25
[TEST!] Publish Logstash plugins from `main` to be consumed by `logstash-docs-md`

colleenmcginnis:
Publishes Logstash plugin pages (like [this page](https://www.elastic.co/guide/en/logstash/8.18/plugins-integrations-aws.html) in the Logstash Reference guide) from `main` so they can be consumed by `logstash-docs-md` to be published on elastic.co/docs.


colleenmcginnis:
This should no longer be necessary. We're using a different approach.

PR #3244, opened 2025-07-25
Update v3-mapping.json for data stream migration files

georgewallace:


kilfoyle (approved):
I'm not 100% sure what this is exactly, but in comparison to the other file contents, LGTM! 🚀

PR #3286, opened 2025-11-19
Make `install_packages` resilient to 500 Server Error

gsoldevila:
Addresses eventual CI failures such as [this one](https://buildkite.com/elastic/docs-build-pr/builds/485039/steps/canvas?jid=019a9742-a415-48d2-a28d-211dd6456130).
image

```
INFO:docker build:#19 [node_deps 4/7] RUN install_packages yarn=1.22.19-1
INFO:docker build:#19 7.734 W: Failed to fetch https://deb.nodesource.com/node_12.x/dists/buster/InRelease  500  Internal Server Error [IP: 172.66.150.169 443]
INFO:docker build:#19 7.734 W: Failed to fetch https://dl.yarnpkg.com/debian/dists/stable/InRelease  500  Internal Server Error [IP: 104.16.169.120 443]
INFO:docker build:#19 7.734 W: Some index files failed to download. They have been ignored, or old ones used instead.
```

The official `install_packages` script fails if the package repository fails with `500 Server Error` when trying to fetch the package lists (which happens from time to time).
This PR adds a wrapper script with a retry mechanism, so that we don't give up that quickly.

PR #3287, opened 2025-11-20
Revise CODEOWNERS for documentation ownership

bmorelli25:
## Summary

Updated CODEOWNERS to better reflect ownership of this repository. Open to suggestions from all.

We could update the writer ownership parts to the elastic/docs team but previous feedback was that it was too noisy for some folks.

Inspired by https://github.com/elastic/docs/pull/3286 which should be owned by docs eng and not writers.

## What changed:

* Docs Eng will own the codebase.
* Writers and Docs Eng will share ownership of config files. This allows writers to move without approval from docs eng in release situations.
* Writers must approve readme changes (my least passionate change in this PR).

cc @elastic/docs-repo-owners and @elastic/docs-engineering.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -5,6 +5,7 @@
 *.asciidoc @elastic/docs-repo-owners
 
 # Writers and Docs Eng share ownership of config files
+/extra/ @elastic/docs-engineering @elastic/docs-repo-owners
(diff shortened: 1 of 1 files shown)


colleenmcginnis (dismissed):
Looks good. One non-blocking question: should writers also co-own [`extra/`](https://github.com/elastic/docs/tree/master/extra)? Not sure if any of these files are even still relevant other than [`extra/docs_landing.html`](https://github.com/elastic/docs/blob/master/extra/docs_landing.html) with the current redirect situation.

PR #3333, opened 2026-05-04
Fix image path check: anchor to book output dir not chapter dir

Mpdreamz:
Anchors the image destination check to `outdir` (the book-level output directory, e.g. `html/en/kibana/8.19/`) rather than `to_dir` (the chapter subdirectory). This allows legitimate `../sibling/` image references within a book while still preventing images from escaping the book output root. Also adds `outdir` to `build_single` for consistency with `build_chunked`.


shainaraskas (approved):
I approve this 😎

